PR #29, opened 2017-10-05
Ignore Firestore types in object traversal

schmidt-sebastian:
Fixes https://stackoverflow.com/questions/46565208/setting-a-documentreference-in-document-on-firestore-nodejs?noredirect=1#comment80108029_46565208

The underlying problem is that the Firebase Admin SDK introduces a cyclic reference that causes us to reach the nesting limit of 20 during the validateDocumentData() call (in the Admin SDK, a 'firestore' property of DocumentReference contains a cyclic reference to itself). 

This PR special cases internal Firestore types to no longer do this traversal and changes all logic that traverses nested maps to ignore Firestore special types.

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -43,6 +43,13 @@ const FieldValue = require('./field-value');
  */
 let DocumentReference;
 
+/*!
+ * Injected.
+ *
+ * @see {Firestore}
+ */
+let Firestore;
+
 /*! Injected. */
 let validate;
 
@@ -646,7 +653,7 @@ class DocumentSnapshot {
       };
     }
 
-    if (is.object(val)) {
+    if (isPlainObject(val)) {
       return {
         valueType: 'mapValue',
         mapValue: {
@@ -690,7 +697,7 @@ class DocumentSnapshot {
           target[key] = {};
           merge(target[key], value, path, pos + 1);
         }
-      } else if (is.object(target[key])) {
+      } else if (isPlainObject(target[key])) {
         if (isLast) {
           // The existing object has deeper nesting that the value we are trying
           // to merge.
@@ -877,7 +884,7 @@ class DocumentMask {
             ? currentPath.append(childSegment)
             : childSegment;
           const value = currentData[key];
-          if (is.object(value)) {
+          if (isPlainObject(value)) {
             extractFieldPaths(value, childPath);
           } else if (value !== FieldValue.SERVER_TIMESTAMP_SENTINEL) {
             fieldPaths.push(childPath.formattedName);
@@ -972,7 +979,7 @@ class DocumentTransform {
           // We need to verify that no array value contains a document transform
           encode_(val[i], path.concat(i), false);
         }
-      } else if (is.object(val)) {
+      } else if (isPlainObject(val)) {
         for (let prop in val) {
           if (val.hasOwnProperty(prop)) {
             transforms = transforms.concat(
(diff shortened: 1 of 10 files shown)


test/document.js
```diff
@@ -565,7 +565,28 @@ describe('serialize document', function() {
 
     assert.throws(() => {
       firestore.doc('collectionId/documentId').update(obj);
-    }, new RegExp('Argument "dataOrField" is not a valid Document. Input ' + 'object is deeper than 20 levels or contains a cycle.'));
+    }, new RegExp('Argument "dataOrField" is not a valid Document. Input object is deeper than 20 levels or contains a cycle.'));
+  });
+
+  it("doesn't traverse cyclic references", function() {
```
hiranya911: Perhaps something like "able to set doc references with cycles" is more  suitable here.

src/document.js
```diff
@@ -1163,7 +1168,30 @@ function validateSetOptions(options) {
...
+ *
+ * @param {*} The argument to verify.
+ * @returns {boolean} 'true' if the input can be a treated as a plain object.
+ */
+function isPlainObject(obj) {
+  return (
+    is.object(obj) &&
+    !is.array(obj) &&
+    !is.date(obj) &&
+    !is.instance(obj, Firestore) &&
```
mikelehen: If you keep this list of exclusions, I don't think Firestore belongs since this isn't something you can write in a document...  I'm also not sure if ResourcePath is valid?

test/document.js
```diff
@@ -565,7 +565,31 @@ describe('serialize document', function() {
...
+          })
+        )
+      );
+
+      callback(null, defaultWriteResult);
+    };
+
+    // The Firestore Admin SDK adds a cyclic reference to the 'Firestore' type.
+    // We emulate this behavior in this test to verify that we can properly
+    // serialize types that contain references to a cyclic Firestore type.
```
mikelehen: "types that contain references to a cyclic Firestore type" is broader than we actually should support.  I think it should just be "DocumentReference instances, even if they have cyclic references (we shouldn't try to validate them beyond the instanceof check)."

src/document.js
```diff
@@ -690,7 +697,7 @@ class DocumentSnapshot {
           target[key] = {};
           merge(target[key], value, path, pos + 1);
         }
-      } else if (is.object(target[key])) {
+      } else if (isPlainObject(target[key])) {
```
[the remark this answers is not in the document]
schmidt-sebastian: This can also be the an object that is passed in by the user if we are still at the root level.

src/document.js
```diff
@@ -1163,7 +1168,30 @@ function validateSetOptions(options) {
...
 
+/*!
+ * Verifies that 'obj' is a plain JavaScript object that does not have a custom
+ * encoding format. Objects that pass this check are encoded as 'Maps' in
+ * Firestore.
+ *
+ * @param {*} The argument to verify.
+ * @returns {boolean} 'true' if the input can be a treated as a plain object.
+ */
+function isPlainObject(obj) {
```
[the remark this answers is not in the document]
schmidt-sebastian: I'm totally in favor.

src/document.js
schmidt-sebastian: I am supporting ResourcePath. Firestore was added as a pre-caution.

But this is no longer relevant as per above.

test/document.js
schmidt-sebastian: Done

src/document.js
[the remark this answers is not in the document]
mikelehen: At the root level won't target be the {} object we create on line 712?  

I think the only way this will be a custom object is if you call expandMap({'foo': new Date(), 'foo.bar': 42}), which I think is invalid.  Oh!  But maybe that is what this code is actually trying to detect?  Gah. If so, I missed that. I thought 'data' was already validated.

schmidt-sebastian:
> I think the only way this will be a custom object is if you call expandMap({'foo': new Date(), 'foo.bar': 42}), which I think is invalid. Oh! But maybe that is what this code is actually trying to detect? Gah. If so, I missed that. I thought 'data' was already validated.

I don't think that was the intend, but thinking about it this does seem to be the only case that this logic catches. It still seems better to have it then now.

PR #32, opened 2017-10-06
Rejecting 'null' and 'undefined' as arguments for CollectionRef.doc()

schmidt-sebastian:
This is for Google-internal bug b/67419579:

[Node] .doc(undefined).set() does an add
We should catch "undefined" being passed and say "hey don't do that" not do an add

### DIFF
--- a/src/reference.js
+++ b/src/reference.js
@@ -1783,10 +1783,10 @@ class CollectionReference extends Query {
    * console.log(`Reference with auto-id: ${documentRefWithAutoId.path}`);
    */
   doc(documentPath) {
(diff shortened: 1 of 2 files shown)


schmidt-sebastian:
@lukesneeringer @stephenplusplus Added you for final Approval. Thanks.

stephenplusplus:
LGTM.

Side-question; can you paste the contents of the internal issue, or at least provide more context for what the bug was?

schmidt-sebastian:
> Side-question; can you paste the contents of the internal issue, or at least provide more context for what the bug was?

Updated the description.

PR #38, opened 2017-10-12
update optional way to getting started with node.js with cloud firestore

noomerzx:
After exprimenting on cloud firestore by following the guide by creating project in cloud platform I've some problem and cannot access firestore db.

So I've creating new project on firebase console instead then go to db section and select the firestore, then generate new config and use the same experimental code [from previous project above] to access firestore and it's work properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,22 @@ Google APIs Client Libraries, in [Client Libraries Explained][explained].
 
     [Go to the projects page][projects]
 
+
+1.  Enable the Google Cloud Firestore API.
+
+    [Enable the API][enable_api]
+
+1.  [Set up authentication with a service account][auth] so you can access the
+    API from your local workstation.
+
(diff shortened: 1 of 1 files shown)


noomerzx:
I sign it

jmdobry:
This README is auto-generated, and therefore your changes would be wiped out the next time we re-generate it.

However, you should be able to use Firestore without a Firebase project, I'm investigating.

lukesneeringer:
Unfortunately, we can not accept this PR (our tooling would plow over it).
Did we figure out the Firebase project part?

jmdobry:
@noomerzx You may have run into the problem that Firestore cannot be used in any Google Cloud Platform project that has every used Datastore. They can't both be used in the same project. You'd have to create a new Datastore-less project to use Firestore.

stephenplusplus:
Closing due to age of the issue.

PR #44, opened 2017-10-27
Rejecting empty document IDs

schmidt-sebastian:
Without this check, it was possible to get a CollectionReference by calling DocumentReference.doc() (and vice-versa).

This addresses Google-internal issue https://b.corp.google.com/issues/64934977


test/collection.js
```diff
@@ -52,15 +52,19 @@ describe('Collection interface', function() {
 
     assert.throws(() => {
       collectionRef.doc(false);
-    }, /Argument "documentPath" is not a valid ResourcePath. Path is not a string./);
+    }, /Argument "documentPath" is not a valid ResourcePath. Path must be a non-empty strin./);
```
garrettjonesgoogle: spelling (strin)

schmidt-sebastian: Fixed (and now we know that the '.' really does match everything).

PR #69, opened 2017-11-11
Adding NodeJS Conformance Test Suite

schmidt-sebastian:
This ports the conformance test from Go and adds some validation/fixes some bugs:
- Rejects nested deletes
- Rejects empty updates
- Rejects empty string arguments to FieldPath
- Allows deletes for set with {merge:true}

There are some tests that are ignored for now:
- Node doesn't skip empty updates as they might occur when the user just provides a server timestamp
- Node doesn't support FieldPaths for merges since it doesn't support "POJO" mapping.
- I believe the test data for 'Delete cannot be nested' is off as it doesn't allow deletes for top-level fields in a JSON object.

@jba: Since I have to use the binary format as well, I don't have any context on the actual test run besides the description. Would it be possible to prepend them with "set/update/..."? Thanks!


jba:
Thanks for doing this!

I added the method name to the descriptions in GoogleCloudPlatform/google-cloud-common#235. I can't add you as a reviewer, maybe because you're not in the org?

conformance/runner.js
```diff
@@ -0,0 +1,321 @@
...
+  if (result.updateTime) {
+    result.lastUpdateTime = DocumentSnapshot.toISOTime(result.updateTime);
+    delete result.updateTime;
+  }
+  return result;
+}
+
+/** List of test cases that are ignored. */
+const ignored = [
+  // Firestore Node does not omit empty writes
```
jba: I was thinking that it would be simpler in the long run if all the clients generated the same, optimal requests.

conformance/runner.js
```diff
@@ -0,0 +1,321 @@
...
+  'ServerTimestamp alone',
+  'nested ServerTimestamp field',
+  'multiple ServerTimestamp fields',
+  'ServerTimestamp with dotted field',
+
+  // Test data doesn't allow update with {a:delete}
```
jba: The actual data in the test is `{a: {b: delete}}`. I thought the way update worked is that the top-level keys of the map or object turned into the update mask, and the values of the map/object were actual firestore values (possibly with serverTimestamp). If you wanted to delete `a.b`, you would write `{a.b: delete}`. But in `{a: {b: delete}}`, the value of `a` is supposed to be an ordinary firestore value, so it can't contain the delete sentinel.

conformance/runner.js
```diff
@@ -0,0 +1,321 @@
...
+  'Delete cannot be nested',
+
+  // Node doesn't support field masks for set().
+  'Merge with a field',
```
jba: I don't understand why you're ignoring these. Some of them are ordinary uses of merge. In fact, "Merge field is not a leaf" is example 5 in go/set-update.

conformance/runner.js
schmidt-sebastian: Yes, I think that's a good optimization. It might get a little tricky to only return one WriteResult with potentially one or two requests per write, but I do agree that it is the right way forward.

conformance/runner.js
schmidt-sebastian: There are two tests that share this name (which is another reason I am a big fan of adding the method name). I believe that the request in `update-paths-11.textproto` should succeed:

```
description: "Delete cannot be nested"
update_paths: <
  doc_ref_path: "projects/projectID/databases/(default)/documents/C/d"
  field_paths: <
    field: "a"
  >
  json_values: "{\"b\": \"Delete\"}"
  is_error: true
>
```

Or does this error in Go because the field path is not included in the data?

conformance/runner.js
schmidt-sebastian: Node only supports `{ merge:true }`. I need to ignore any tests that have field paths set. Once I pull in the changes from https://github.com/GoogleCloudPlatform/google-cloud-common/pull/235, I can likely prune this list.

conformance/runner.js
jba: I just always return the first WriteResult. Is that bad if there is only a transform?

conformance/runner.js
jba: To me, 
```
update("a", {b: delete})
```
is the same thing as
```
update({a: {b: delete}})
```
The varargs and map/object forms should behave the same.

conformance/runner.js
jba: Oh, I re-read the spec and now I get it.

conformance/runner.js
schmidt-sebastian: I'm not doing this correctly right now in Node, but I think we should return the result from the write transform (whether there is one or two). If an update doesn't change the document, but the transform does, then the WriteResult for the update will be stale and the updateTime will only be correctly reflected in the document transform.

Are you returning WriteResults from WriteBatches? It gets a little trickier to filter out these extra WriteResults in WriteBatches.

conformance/runner.js
schmidt-sebastian: Sorry, I've been misreading this test case all along. I somehow thought it was applying {b:delete()} to the document and not to the field path `a`. You are correct that this should be rejected.

schmidt-sebastian:
@jba: I updated the PR to use the new test names and fixed the bug with `update( 'a', {b : delete} )`.

Can you sanity check this? I would like to submit as-is and fix the transform issues in a follow-up effort.

PR #70, opened 2017-11-14
Retrying on any error from the Watch stream

schmidt-sebastian:
This ports the exponential backoff logic from the Firestore Mobile Web client.

Fixes https://github.com/firebase/firebase-admin-node/issues/119

### DIFF
--- a/src/backoff.js
+++ b/src/backoff.js
@@ -50,6 +50,11 @@ const BACKOFF_MAX_DELAY_MS = 60 * 1000;
 const BACKOFF_FACTOR = 1.5;
 
 /*!
+ * The jitter to distribute the backoff attempts by.
+ */
+const JITTER_FACTOR = 1.0;
+
+/*!
  * List of GRPC Error Codes.
  *
  * This corresponds to
@@ -230,7 +235,19 @@ function isResourceExhaustedError(error) {
  * @private
  */
 class ExponentialBackoff {
-  constructor() {
+  /**
+   * @param {number=} options.initialDelayMs Optional override for the initial
+   * retry delay.
+   * @param {number=} options.backoffFactor Optional override for the
+   * exponential backoff factor.
+   * @param {number=} options.maxDelayMs Optional override for the maximum
+   * retry delay.
+   * @param {number=} options.jitterFactor Optional override to control the
+   * jitter. Supported ranges are [0,1].
+   */
+  constructor(options) {
+    options = options || {};
+
     /**
      * The initial delay (used as the base delay on the first retry attempt).
      * Note that jitter will still be applied, so the actual delay could be as
@@ -239,7 +256,10 @@ class ExponentialBackoff {
      * @type {number}
      * @private
      */
-    this._initialDelayMs = BACKOFF_INITIAL_DELAY_MS;
+    this._initialDelayMs =
+      options.initialDelayMs !== undefined
+        ? options.initialDelayMs
+        : BACKOFF_INITIAL_DELAY_MS;
 
     /**
      * The multiplier to use to determine the extended base delay after each
@@ -247,7 +267,10 @@ class ExponentialBackoff {
      * @type {number}
      * @private
      */
-    this._backoffFactor = BACKOFF_FACTOR;
+    this._backoffFactor =
+      options.backoffFactor !== undefined
+        ? options.backoffFactor
+        : BACKOFF_FACTOR;
 
     /**
      * The maximum base delay after which no further backoff is performed.
@@ -257,7 +280,20 @@ class ExponentialBackoff {
      * @type {number}
      * @private
      */
-    this._maxDelayMs = BACKOFF_MAX_DELAY_MS;
+    this._maxDelayMs =
+      options.maxDelayMs !== undefined
+        ? options.maxDelayMs
+        : BACKOFF_MAX_DELAY_MS;
+
+    /**
+     * The jitter factor that controls the random distribution of the backoff
+     * points.
+     *
+     * @type {number}
+     * @private
+     */
+    this._jitterFactor =
+      options.jitterFactor !== undefined ? options.jitterFactor : JITTER_FACTOR;
 
     /**
      * The backoff delay of the current attempt.
@@ -326,13 +362,17 @@ class ExponentialBackoff {
   }
 
   /**
-   * Returns a random value in the range [-currentBaseMs/2, currentBaseMs/2]
+   * Returns a random value for the current base as determined by the jitter
+   * factor.
    *
    * @private
    * @returns The jitter to apply based on the current delay.
    */
   _jitterDelayMs() {
-    return (Math.random() - 0.5) * this._currentBaseMs;
+    return (
+      (Math.random() * this._jitterFactor - this._jitterFactor / 2) *
+      this._currentBaseMs
+    );
   }
 }
 
--- a/src/backoff.js
+++ b/src/backoff.js
@@ -1,0 +1,347 @@
+/*!
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/*!
+ * Injected.
+ *
+ * @see Firestore
+ */
+let Firestore;
+
+/*
+ * @module firestore/backoff
+ * @private
+ *
+ * Contains backoff logic to facilitate RPC error handling. This class derives
+ * its implementation from the Firestore Mobile Web Client.
+ *
+ * @see https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/backoff.ts
+ */
+
+/*!
+ * Initial backoff time in milliseconds after an error.
+ * Set to 1s according to https://cloud.google.com/apis/design/errors.
+ */
+const BACKOFF_INITIAL_DELAY_MS = 1000;
+
+/*!
+ * Maximum backoff time in milliseconds.
+ */
+const BACKOFF_MAX_DELAY_MS = 60 * 1000;
+
+/*!
+ * The factor to increase the backup by after each failed attempt.
+ */
+const BACKOFF_FACTOR = 1.5;
+
+/*!
+ * List of GRPC Error Codes.
+ *
+ * This corresponds to
+ * {@link https://github.com/grpc/grpc/blob/master/doc/statuscodes.md}.
+ */
+const GRPC_STATUS_CODE = {
+  // Not an error; returned on success.
+  OK: 0,
+
+  // The operation was cancelled (typically by the caller).
+  CANCELLED: 1,
+
+  // Unknown error. An example of where this error may be returned is if a
+  // Status value received from another address space belongs to an error-space
+  // that is not known in this address space. Also errors raised by APIs that
+  // do not return enough error information may be converted to this error.
+  UNKNOWN: 2,
+
+  // Client specified an invalid argument. Note that this differs from
+  // FAILED_PRECONDITION. INVALID_ARGUMENT indicates arguments that are
+  // problematic regardless of the state of the system (e.g., a malformed file
+  // name).
+  INVALID_ARGUMENT: 3,
+
+  // Deadline expired before operation could complete. For operations that
+  // change the state of the system, this error may be returned even if the
+  // operation has completed successfully. For example, a successful response
+  // from a server could have been delayed long enough for the deadline to
+  // expire.
+  DEADLINE_EXCEEDED: 4,
+
+  // Some requested entity (e.g., file or directory) was not found.
+  NOT_FOUND: 5,
+
+  // Some entity that we attempted to create (e.g., file or directory) already
+  // exists.
+  ALREADY_EXISTS: 6,
+
+  // The caller does not have permission to execute the specified operation.
+  // PERMISSION_DENIED must not be used for rejections caused by exhausting
+  // some resource (use RESOURCE_EXHAUSTED instead for those errors).
+  // PERMISSION_DENIED must not be used if the caller can not be identified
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -809,7 +819,8 @@ class Firestore extends commonGrpc.Service {
    * @param {Stream} resultStream - The Node stream to monitor.
    * @param {Object=} request - If specified, the request that should be written
    * to the stream after it opened.
-   * @returns {Promise.<Stream>} The given Stream once it is considered healthy.
+   * @returns {Promise.<Firestore~StreamCallback>} A Promise that resolves with
+   * a callback that returns the resulting stream.
```
mikelehen: This is pretty confusing.  We have an asynchronous method that returns a callback/factory function that you have to call to actually get the result of the asynchronous method?

I kind of understand how we ended up here, but are there other ways to structure this code perhaps?  I'm somewhat disinclined to figure out exactly what's going on myself, but offhand I'm wondering if this method should be asynchronous in the first place?  Why do we want to handle "error before the first byte is read" as a failed Promise but any failure after that point as a stream error?  If this method could just synchronously return a (wrapped?) stream, that would avoid the need for the StreamCallback thing, I think.

Feel free to push back, but I think it's worth thinking a bit about other potential ways to structure this since this is getting a bit wonky.

src/index.js
```diff
@@ -947,16 +972,19 @@ class Firestore extends commonGrpc.Service {
    * A funnel for read-only streaming API requests, assigning a project ID where
    * necessary within the request options.
    *
-   * The stream is returned in paused state and needs to be resumed once all
-   * listeners are attached.
+   * The Promise returned from this method resolves with callback that needs to
+   * be invoked in order to obtain the underlying stream. The stream stream is
+   * returned in paused state and should to be resumed once all listeners are
```
mikelehen: stream stream => stream
should to be => should be

Given your callback/factory change thing, is it even necessary to pause / resume the stream anymore?

src/watch.js
```diff
@@ -20,6 +20,10 @@ const assert = require('assert');
 const rbtree = require('functional-red-black-tree');
 const through = require('through2');
 
+let isPermanentError;
```
mikelehen: Should these be under the "Injected" comment too, or are they somehow different?

src/watch.js
```diff
@@ -238,32 +240,34 @@ class Watch {
...
         request.addTarget.resumeToken = resumeToken;
         changeMap.clear();
+
+        if (err && isResourceExhaustedError(err)) {
+          self._backoff.resetToMax();
+        }
+
         resetStream();
-      } else if (err) {
+      } else {
```
mikelehen: why do we no longer check "if (err)" ?  Calling sendError() with null / undefined seems suspicious (it'll call onError() with no error, etc.)

src/watch.js
```diff
@@ -276,37 +280,48 @@ class Watch {
...
-          /* allowRetries= */ true
-        )
-        .then(backendStream => {
-          if (!isActive) {
-            Firestore.log('Watch.onSnapshot', 'Closing inactive stream');
-            backendStream.end();
-            return;
-          }
+    /**
+     * Initializes a new the stream to the backend with backoff.
```
mikelehen: new the stream => new stream

test/backoff.js
```diff
@@ -0,0 +1,150 @@
...
+
+describe('ExponentialBackoff', function() {
+  const observedDelays = [];
+
+  before(() => {
+    setTimeoutHandler((callback, timeout) => observedDelays.push(timeout));
+  });
+
+  beforeEach(() => {
+    observedDelays.length = 0;
```
mikelehen: `observedDelays = []` ?  (just because JavaScript lets you do crazy things doesn't mean you have to :P)

test/backoff.js
```diff
@@ -0,0 +1,150 @@
...
+    setTimeoutHandler(setTimeout);
+  });
+
+  function assertDelayEquals(expected) {
+    assert.equal(observedDelays.shift(), expected);
+  }
+
+  function assertDelayBetween(low, high) {
+    let actual = observedDelays.shift();
+    assert.ok(actual >= low);
```
mikelehen: FWIW (probably not much), I'm surprised we're not using a more powerful assertion framework (chai or whatever) which would let us do a proper expect(actual).toBeGreaterThanOrEqualTo(low) or whatever, with the advantage that you get better error messages on failure (rather than "expected false to be true" or whatever)

test/watch.js
```diff
@@ -591,6 +597,8 @@ describe('Query watch', function() {
   };
 
   beforeEach(function() {
+    Backoff.setTimeoutHandler(setImmediate);
```
mikelehen: So we are intentionally skipping the delays?  Can you add a comment?

test/watch.js
```diff
@@ -696,22 +708,110 @@ describe('Query watch', function() {
...
+          assert.equal(streamHelper.streamCount, 3);
         });
     });
   });
 
-  it("doesn't re-open without resume token", function() {
-    watchHelper.startWatch();
+  it("doesn't re-open inactive stream", function() {
+    // This test uses the normal timeout handler since we unsubscribe from the
+    // Watch stream while the stream is recovering from an error.
```
mikelehen: I don't really understand this comment.  Why does unsubscribing while the watch stream is recovering mean we have to use setTimeout instead of setImmediate?

test/watch.js
```diff
@@ -986,7 +1058,7 @@ describe('Query watch', function() {
               streamHelper.writeStream.destroy();
             });
         },
-        'Error: Steam Error (6)'
+        'Steam Error (6)'
```
mikelehen: Steam Error?

test/watch.js
```diff
@@ -1809,13 +1881,19 @@ describe('DocumentReference watch', function() {
   };
 
   beforeEach(function() {
+    Backoff.setTimeoutHandler(setImmediate);
```
mikelehen: Comment again?

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+ * @private
+ *
+ * Contains backoff logic to facilitate RPC error handling. This class derives
+ * its implementation from the Firestore Mobile Web Client.
+ *
+ * @see https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/backoff.ts
+ */
+
+/*!
+ * Initial backoff time in milliseconds after an error.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+ * Set to 1s according to https://cloud.google.com/apis/design/errors.
+ */
+const BACKOFF_INITIAL_DELAY_MS = 1000;
+
+/*!
+ * Maximum backoff time in milliseconds.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+ */
+const BACKOFF_MAX_DELAY_MS = 60 * 1000;
+
+/*!
+ * The factor to increase the backup by after each failed attempt.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+ */
+const BACKOFF_FACTOR = 1.5;
+
+/*!
+ * The jitter to distribute the backoff attempts by.
```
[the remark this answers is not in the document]
schmidt-sebastian: I kept it as is for now but updated some of the comments. I agree it would be better to keep it in sync, but the code can still be easily matched to the original implementation.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+ */
+const JITTER_FACTOR = 1.0;
+
+/*!
+ * List of GRPC Error Codes.
+ *
+ * This corresponds to
+ * {@link https://github.com/grpc/grpc/blob/master/doc/statuscodes.md}.
+ */
+const GRPC_STATUS_CODE = {
```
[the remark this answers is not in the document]
schmidt-sebastian: It makes a lot of sense to put this into its own file. It makes slightly less sense to put it in index.js. Slightly behind that, we have watch.js. And then, there's backoff.js.

I don't really want to add another file just for this, and putting it in index.js requires a lot of plumbing/injecting to make it available. So, for now, I have moved this code to watch.js.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+class ExponentialBackoff {
+  /**
+   * @param {number=} options.initialDelayMs Optional override for the initial
+   * retry delay.
+   * @param {number=} options.backoffFactor Optional override for the
+   * exponential backoff factor.
+   * @param {number=} options.maxDelayMs Optional override for the maximum
+   * retry delay.
+   * @param {number=} options.jitterFactor Optional override to control the
+   * jitter. Supported ranges are [0,1].
```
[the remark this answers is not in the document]
schmidt-sebastian: I added a comment that this range is "suggested".

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+function isResourceExhaustedError(error) {
+  return error.code === GRPC_STATUS_CODE.RESOURCE_EXHAUSTED;
+}
+
+/**
+ * A helper for running delayed tasks following an exponential backoff curve
+ * between attempts.
+ *
+ * Each delay is made up of a "base" delay which follows the exponential
+ * backoff curve, and a +/- 50% "jitter" that is calculated and added to the
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+   * retry delay.
+   * @param {number=} options.jitterFactor Optional override to control the
+   * jitter. Supported ranges are [0,1].
+   */
+  constructor(options) {
+    options = options || {};
+
+    /**
+     * The initial delay (used as the base delay on the first retry attempt).
+     * Note that jitter will still be applied, so the actual delay could be as
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+     * @private
+     */
+    this._backoffFactor =
+      options.backoffFactor !== undefined
+        ? options.backoffFactor
+        : BACKOFF_FACTOR;
+
+    /**
+     * The maximum base delay after which no further backoff is performed.
+     * Note that jitter will still be applied, so the actual delay could be as
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (slightly reworded)

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    return new Promise(resolve => {
+      delayExecution(resolve, delayWithJitterMs);
+    });
+  }
+
+  /**
+   * Returns a random value for the current base as determined by the jitter
+   * factor.
```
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+   *
+   * @private
+   * @returns The jitter to apply based on the current delay.
+   */
+  _jitterDelayMs() {
+    return (
+      (Math.random() * this._jitterFactor - this._jitterFactor / 2) *
```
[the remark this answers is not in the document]
schmidt-sebastian: I went back and forth between what I had and what you have there. Now we have 1.5 votes for your math, and .5 for mine, so I picked the more concise version (I do find it a little less obvious though).

src/index.js
```diff
@@ -161,6 +161,14 @@ const GRPC_UNAVAILABLE = 14;
  */
 
 /**
+ * A callback function that creates a stream.
```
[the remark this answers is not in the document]
schmidt-sebastian: I didn't want to name this factory since it should only ever be called once.

src/index.js
schmidt-sebastian: As per Josh's suggestion, I can rely on Node's implementation of microtask and macrotasks and schedule a macrotask here to make sure that the Promise chain gets resolved first.

Oh, and I really really really hated the original approach, but it was the only one I could come up with that at least somehow kept the original API contract: readStream() and readWriteStream() return a Stream that is known to be writeable.

src/index.js
schmidt-sebastian: It was not, but I didn't want to debug even more test failures...

src/watch.js
schmidt-sebastian: No longer relevant, but yes.

src/watch.js
schmidt-sebastian: I fixed this by a) always calling this method with an error and b) only invoking the onError callback if we are still active.

src/watch.js
schmidt-sebastian: Done.

test/backoff.js
schmidt-sebastian: I never verified my assumption, but I was worried that the timeout handler captured just the variable at the time and would lose its reference to it when I re-assigned it. But it turns out that that's not how closures work.

test/backoff.js
schmidt-sebastian: Now that we are in our own repo, I'm sure that we can migrate to chai. At some point that is.

test/watch.js
schmidt-sebastian: Added comment.

test/watch.js
schmidt-sebastian: I improved the comment, but this test is my pride since it executes the following line:

```
      self._backoff.backoffAndWait().then(() => {
        if (!isActive) {
          Firestore.log('Watch.onSnapshot', 'Not initializing inactive stream');
          return;
        }
```

test/watch.js
schmidt-sebastian: Done

test/watch.js
schmidt-sebastian: Also done.

src/backoff.js
```diff
@@ -0,0 +1,387 @@
...
+     * @private
+     */
+    this._maxDelayMs =
+      options.maxDelayMs !== undefined
+        ? options.maxDelayMs
+        : BACKOFF_MAX_DELAY_MS;
+
+    /**
+     * The jitter factor that controls the random distribution of the backoff
+     * points.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

schmidt-sebastian:
Addressed all comments. Thanks @mikelehen

mikelehen:
LGTM. Sorry for the delay.

schmidt-sebastian:
@stephenplusplus Can you also look at this before I merge this? This adds retry semantics for Watch.

stephenplusplus:
LGTM.

PR #78, opened 2017-11-21
Updating the spec protos

schmidt-sebastian:
Pulls in the latest file from https://github.com/GoogleCloudPlatform/google-cloud-common/blob/30710c621b2b2b1fdcb1d352ebb0b8cc63507260/testing/firestore/testdata/tests.binprotos


mikelehen:
Out of curiosity, could we use a JSON representation of the protos or something for sane diagnosability of failures, etc.?

schmidt-sebastian:
> Out of curiosity, could we use a JSON representation of the protos or something for sane diagnosability of failures, etc.?

The original tests in Go use text-based Protos which are not natively supported by ProtobufJS. I tried to make them work but eventually gave up when I ran into encoding issues. Instead, these tests print the JSON representation to the console at the beginning of each run.

PR #82, opened 2017-11-29
Retrying without an error code

schmidt-sebastian:
This hopefully addresses more causes of https://github.com/firebase/firebase-admin-node/issues/119#issuecomment-348008611

### DIFF
--- a/src/watch.js
+++ b/src/watch.js
@@ -205,14 +205,19 @@ const DOCUMENT_WATCH_COMPARATOR = (doc1, doc2) => {
 };
 
 /**
- * Determines whether a GRPC Error is considered permanent and should not be
- * retried.
+ * Determines whether an error is considered permanent and should not be
+ * retried. Errors that don't provide a GRPC error code are always considered
+ * transient in this context.
  *
  * @private
- * @param {Error} error A GRPC Error object that exposes an error code.
+ * @param {Error} error An error object.
  * @return {boolean} Whether the error is permanent.
  */
 function isPermanentError(error) {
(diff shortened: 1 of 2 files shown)


src/watch.js
```diff
@@ -205,14 +205,19 @@ const DOCUMENT_WATCH_COMPARATOR = (doc1, doc2) => {
...
+ * transient in this context.
  *
  * @private
- * @param {Error} error A GRPC Error object that exposes an error code.
+ * @param {Error} error An error object.
  * @return {boolean} Whether the error is permanent.
  */
 function isPermanentError(error) {
+  if (error.code === undefined) {
+    return false;
```
mikelehen: Hrm...  There's no way for me to know if this is correct, since I don't know what case(s) cause errors without a code.  FWIW- I ran into this no-error-code case in the client SDK when I accidentally sent invalid protos to GRPC (which *is* a permanent error)...

It's probably not terribly harmful to retry (with backoff), but it makes me uneasy...  I'd feel a little better if we at least logged something.  Is logging a thing in this SDK? :-)  Perhaps also add a comment with whatever knowledge you have about this case ("GRPC can give us an error with no code if ...  and we want to retry that.").

Even better would be to get GRPC to give us an error code for whatever case we're hoping to solve here so we don't need this.

schmidt-sebastian: Yeah, I agree with you. I mostly ran out of ideas what could be causing this, and I haven't been able to reproduce this locally.

I have added logging and will ask the customer to patch their client before we push another release.

PR #86, opened 2017-12-19
Pegging GRPC version to 1.7.1

schmidt-sebastian:
GRPC 1.8.0 seems to cause 'Stream Removed' Errors on Firebase Functions: https://github.com/grpc/grpc-node/issues/130

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "extend": "^3.0.1",
     "functional-red-black-tree": "^1.0.1",
     "google-gax": "^0.14.1",
-    "grpc": "^1.6.0",
+    "grpc": "1.7.1",
     "is": "^3.2.0",
     "safe-buffer": "^5.1.1",
     "through2": "^2.0.3"


package.json
```diff
@@ -56,7 +56,7 @@
     "extend": "^3.0.1",
     "functional-red-black-tree": "^1.0.1",
     "google-gax": "^0.14.1",
-    "grpc": "^1.6.0",
+    "grpc": "1.7.1",
```
lukesneeringer: Is there a way to specify >= 1.6.0, < 1.8?

schmidt-sebastian: Not as far as I know. >= 1.7.0 to < 1.8.0 is the closest we got.

callmehiphop:
Did using 1.7.x resolve this for you? We're seeing the same error in Spanner, however it doesn't use 1.8, it's been using 1.7 the whole time.

schmidt-sebastian:
> Did using 1.7.x resolve this for you? We're seeing the same error in Spanner, however it doesn't use 1.8, it's been using 1.7 the whole time.

Preparing to revert this PR after confirming that this issue exists in 1.7.1.

PR #88, opened 2017-12-19
Revert "Pegging GRPC version to 1.7.1"

schmidt-sebastian:
Reverts googleapis/nodejs-firestore#86


schmidt-sebastian:
We actually have no evidence that this issue occurs with 1.7.1. Leaving things as it.

PR #92, opened 2017-12-21
Only allow patch releases for our dependencies

schmidt-sebastian:
I hope this will help reduce future breakages as those with GRPC 1.8.0 (https://github.com/grpc/grpc-node/issues/130).

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,16 +50,16 @@
     "test": "repo-tools test run --cmd npm -- run cover"
   },
   "dependencies": {
-    "@google-cloud/common": "~0.15.0",
-    "@google-cloud/common-grpc": "~0.5.0",
-    "bun": "~0.0.12",
-    "extend": "~3.0.1",
-    "functional-red-black-tree": "~1.0.1",
-    "google-gax": "~0.14.2",
+    "@google-cloud/common": "^0.15.0",
+    "@google-cloud/common-grpc": "^0.5.0",
+    "bun": "^0.0.12",
+    "extend": "^3.0.1",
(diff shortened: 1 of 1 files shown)


stephenplusplus:
@ofrobots what do you think about this approach?

@schmidt-sebastian can you link the relevant issue from grpc that explains the breakage?

schmidt-sebastian:
> @schmidt-sebastian can you link the relevant issue from grpc that explains the breakage?

Sorry, I should have done that to begin with: https://github.com/grpc/grpc-node/issues/130

We saw high errors numbers on Google Cloud Functions and many user reports after we automatically pulled in this update. There are some links to these reports in the GitHub issue.

ofrobots:
If I understand correctly, grpc messed up by making a breaking behaviour change in a semver-minor update. This is terrible, but it is still a mistake made by gRPC. We should provide this feedback to the grpc folks (/cc @murgatroid99) and let them fix. Locking down *all* dependencies using `~` dependencies seems overkill. If gRPC continues break the semver contract, we should absolutely stop trusting their updates, and perhaps escalate.

The problem with a locking down dependencies is that in the npm module ecosystem fixes (bug fixes and security fixes) are typically provided on the tip of the major branch. If you are restricting to `through2` as `~2.0.3` then you're not going to get *any* fixes after `through2@2.1.0` happens. This will rapidly lead to the package being out of date and potentially insecure.

/cc @JustinBeckwith

schmidt-sebastian:
@ofrobots Sounds good. I will update this PR to allow future versions of GRPC patch releases for 1.7.x only (right now we lock to 1.7.1).

PR #94, opened 2018-01-02
Cleanup

schmidt-sebastian:
This is just a bit of cleanup from migrating this code to Java.

The "big" change in here is that _fieldsProto is now {} instead of `null` for empty documents.


src/document.js
```diff
@@ -576,8 +576,9 @@ class DocumentSnapshot {
    * @return {boolean}
    */
   get isEmpty() {
-    return !this._fieldsProto || isEmptyObject(this._fieldsProto);
+    return this._fieldsProto === undefined || isEmptyObject(this._fieldsProto);
```
stephenplusplus: Since we have `is`, how about `is.undefined(this._fieldsProto)`?

stephenplusplus: There is also `is.empty(obj)`

src/document.js
```diff
@@ -652,7 +650,6 @@ class DocumentSnapshot {
 
   /**
    * Encodes a JavaScrip value into the Firestore 'Value' representation.
```
stephenplusplus: JavaScript*

src/document.js
schmidt-sebastian: Good suggestions. I removed `isEmptyObject`.

src/document.js
schmidt-sebastian: Fixed

stephenplusplus:
Linting failed :( Probably another case of your local copy pulling from npm’s cache.

schmidt-sebastian:
> Linting failed :( Probably another case of your local copy pulling from npm’s cache.

Yes, I still don't quite know where that is coming from. I continued on some other PRs as well that have the same errors which I will send out shortly. When I finally figure out the root cause, I'll update them all in one go.

stephenplusplus:
I'll just approve so you're good to merge whenever linting stops being a downer.

schmidt-sebastian:
Whaaaa.... Updating `eslint` and `prettier` to their latest dependencies worked.

Issue #98, opened 2018-01-07
DocumentSnapshot in startAt or startAfter cause an error: Error: Cannot encode type ([object Object]) to a Firestore Value


Erbolking:
Hello,

According to the https://firebase.google.com/docs/firestore/query-data/query-cursors
we could be able to use a **document snapshot** to define the query cursor.

but in the fact, it throws an error: "Error: Cannot encode type ([object Object]) to a Firestore Value".

#### Environment details

  - OS: Windows 10
  - Node.js version: 9.2.0
  - npm version: 5.5.1
  - @google-cloud/firestore version: 0.10.0

#### Steps to reproduce

```js
const first = firestore.collection("collectionName")
  .orderBy("population")
  .limit(25)
  .get()
  .then(function (documentSnapshots) {
    // Get the last visible document
    const lastVisible = documentSnapshots.docs[documentSnapshots.docs.length-1];

    const next = firestore.collection("collectionName")
      .orderBy("population")
      .startAfter(lastVisible)
      .limit(25)
      .get()
      .then(data => {
        console.log(data.docs.length);
      });
  });
```

I tried the same with `@firebase/firestore` and it works just fine.

schmidt-sebastian:
This is not (yet) supported by the Node Server SDK. If you look under the Node.JS tab, you will see that the code snippet for Node is missing. I am planning to add this support soon.

PR #100, opened 2018-01-10
Adding DocumentSnapshot support for Cursors

schmidt-sebastian:
Adds support for DocumentSnapshots in the Query Cursor API. Related: https://github.com/googleapis/nodejs-firestore/issues/98


src/reference.js
```diff
@@ -1224,19 +1231,64 @@ class Query {
...
+      if (!hasDocumentId) {
+        // Add implicit sorting by name, using the last specified direction.
+        let lastDirection =
+          this._fieldOrders.length === 0
+            ? directionOperators.ASC
+            : this._fieldOrders[this._fieldOrders.length - 1].direction;
+
+        this._fieldOrders.push(
+          new FieldOrder(FieldPath.documentId(), lastDirection)
+        );
```
mikelehen: So I was surprised to find a state mutation buried inside a method called buildPosition_() (from the name I assumed no side-effects). And I'm not sure if this is appropriate, since we're modifying the Query object you're calling startAt() _on_ rather than the one that gets _returned_. I'm not 100% sure if that breaks anything (what happens if you then do an explicit orderBy on the document id?), but in general I don't think `query.startAt(...)` should result in mutating `query`. 

FWIW- this method is also getting pretty huge. I wonder if it'd make sense to create a startAtEndAtHelper_() that might look something like:

```
let fieldValues;
let fieldOrders = this._fieldOrders;
if (args[0] is DocumentSnapshot) {
  // Code to ensure fieldOrders contains document ID...
  fieldValues = this.getFieldValuesFromDocumentSnapshot(args[0], fieldOrders);
} else {
  fieldValues = args;
}

const position = this.buildPosition_(fieldValues);

return new Query(..., fieldOrders, position);
```
(sorry this is terrible pseudocode and feel free to choose better method names, but hopefully you get the idea)

schmidt-sebastian: Hm, so I thought about this PR some more and found more issues. For once, there is the issue you pointed out (modifying the immutable state), and there there is this rabbit-hole that Jonny once warned me about: `query.where(...).startAfter(doc)` will break in unexpected ways (at least for the casual user).

So, in order to be able to append `__name__` to where() queries, I ported the logic from the Mobile Clients. I used this to split up the method into three smaller ones:

- Extracting the field values
- Computing the implicit order by
- Generation of the Protobuf message based on the fields and the order by constraint

Unlike the Mobile SDKs, I am rejecting calls to `where()` once a user has specified a cursor to not invalidate it. This matches the behavior of the `orderBy()` API.

PR #114, opened 2018-01-23
Allow newlines in Resource Paths

schmidt-sebastian:
This fixes b/72385233 by allowing all characters for document IDs.


src/path.js
```diff
@@ -27,7 +27,7 @@ const validate = require('./validate')();
  *
  * @type {RegExp}
  */
-const RESOURCE_PATH_RE = /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents\/)?(.*)$/;
+const RESOURCE_PATH_RE = /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents\/)?([\s\S]*)$/;
```
mikelehen: While more correct, this new regex is less readable. :p Googling for \s\S does not yield any useful results. Consider adding a comment.

schmidt-sebastian: It's the JS equivalent of the /s flag (which matches newlines in most RegExp parsers). I added a comment.

PR #118, opened 2018-01-29
Adding Query Conformance Tests

schmidt-sebastian:
Porting the newly added Conformance Tests from the Query API.

This imports the tests from https://github.com/GoogleCloudPlatform/google-cloud-common/pull/245

The only changes in the client are:
- If only one query clause is specified, we no longer issue a composite query with just one clause.
- 'before:false's no longer explicitly set in the cursor Proto.


conformance/runner.js
```diff
@@ -105,14 +120,87 @@ function convertCommit(commitRequest) {
...
 }
 
-/** Converts a Protobuf Precondition to the expected user input. */
+/** Converts a Position in Proto3 JSON to Protobuf JS format. */
+function convertPosition(position) {
+  const values = [];
+  for (let value of position.values) {
+    values.push(convert.valueFromJson(value));
+  }
+  position.values = values;
```
mikelehen: Here you're mutating the provided position in-place, but below you're making a copy first. Seems like you should always make a copy first. 

FWIW- You could also ditch the for loop by using Array.map() I think.

conformance/runner.js
```diff
@@ -105,14 +120,87 @@ function convertCommit(commitRequest) {
...
+    deepCopy.startAt = convertPosition(deepCopy.startAt);
+  }
+  if (deepCopy.endAt) {
+    deepCopy.endAt = convertPosition(deepCopy.endAt);
+  }
+  return deepCopy;
+}
+
+/** Converts a `Precondition` to its API counterpart. */
 function convertPrecondition(precondition) {
```
mikelehen: We have some convert*() methods that convert between protobuf formats, and some that convert between API format and proto format. This seems confusing and potentially error-prone. Does it make sense to have an apiConverter "namespace" and protoConverter "namespace" to clarify what kind of conversions these methods do?

conformance/runner.js
schmidt-sebastian: We don't access this state afterwards, so it was fine in this case. To make the life for future Sebastian easier, I followed your advise though and made a copy.

conformance/runner.js
schmidt-sebastian: It was actually rather confusing when I went through and update the comments. I created two namespaces (one for the input and one for the Protobuf conversion) and translated what was in the comments into code.

PR #119, opened 2018-01-29
Including previous data when we encounter a transform

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-admin-node/issues/191

When we compute the merge document from an update call, we cannot just skip all leaf paths that include document transform. We need to retain any existing data that may already be available at this path.


gsoltis:
LGTM. Looks like I'm not part of the org yet though.

PR #120, opened 2018-01-29
Run circle as non-root user

alexander-fenster:
Due to an [issue](https://www.google.com/url?q=https://github.com/ForbesLindesay/tar-pack/issues/35&sa=D&usg=AFQjCNHmM3ietdPoDETP1lO1PHF9le3OVQ) with permissions in grpc tarball our CI runs extra 2 minutes per task because it recompiles grpc all the time. Running as non-root user makes it unpack tarball successfully and speeds up CI. I made the same change to most of our Node.js repos (as I'm really tired of waiting for CI to spend cycles on recompiling stuff).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,8 +69,8 @@ workflows:
             - lint
             - docs
           filters:
-            branches:
-              only: master
+#           branches:
+#             only: master
             tags:
               only: /^v[\d.]+$/
       - publish_npm:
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
Does this mean that we can remove --unsafe-perm?

alexander-fenster:
Yes, removed `--unsafe-perm`.

PR #130, opened 2018-02-09
RFC: Improving SDK interoperability

schmidt-sebastian:
This aims to address a common user error in GCF:  It is easy to read from Firestore using the Firestore triggers (via the Admin SDK), and straightforward to write to Firestore using the GCloud Firestore SDK.  Unfortunately, lots of stuff breaks in subtle ways when users mix and match SDK versions.  The real solution to this problem is of course to not use the GCloud Firestore SDK directly, but instead import it from the Admin SDK (as pointed out in our docs).

This PR aims to help to get the non-doc readers help on the right track.

As an example, the offending code can be as concise as follows:

```
const admin = require('firebase-admin');
const firestore = require('@google-cloud/firestore');

admin.initializeApp({
  credential: admin.credential.cert()...),
});

const ref = admin.firestore().doc('foo/bar');
return ref
  .set({time: firestore.FieldValue.serverTimestamp()}) <-- Wrong type
  .then(() => ref.get())
  .then(snap => console.log(snap.get('time')));
```

In the current version of the SDK, this prints... you guessed it... '{}'. firestore.FieldValue.serverTimestamp() is just seen as a plain old object (the sentinel value is {}) and the equals comparison fails.

With this PR, if users pull in the same version of the SDK via GCloud and Firebase Admin, everything just works. We simply pull the version of the SDK that was used first.

If they pull in two different versions, first we log a warning (on import):

```
Importing multiple versions @google-cloud/firestore can cause unexpected behavior. If you are using Firestore through a transitive dependency (such as the Firebase Admin SDK),   please ensure that your dependency versions match (found 0.10.0 and 0.11.2). For more information, see http://goo.gl/...
```

We then throw an error when they use the wrong type:

```
Error: Argument "data" is not a valid Document. Detected an object of type 'FieldValue' that doesn't match the expected version for this Firestore instance. For more information, please visit: http://go...
```

This PR is missing tests, comments, proper names, and more... but it would be nice to get some initial thoughts on the general idea behind it.

### DIFF
--- a/src/field-value.js
+++ b/src/field-value.js
@@ -20,13 +20,13 @@
  * Sentinel value for a field delete.
  *
  */
-const DELETE_SENTINEL = {};
+let DELETE_SENTINEL;
 
 /*!
  * Sentinel value for a server timestamp.
  *
  */
-const SERVER_TIMESTAMP_SENTINEL = {};
+let SERVER_TIMESTAMP_SENTINEL;
 
 /**
  * Sentinel values that can be used when writing documents with set() or
@@ -35,6 +35,7 @@ const SERVER_TIMESTAMP_SENTINEL = {};
  * @class
  */
 class FieldValue {
+  constructor() {}
   /**
    * Returns a sentinel used with update() to mark a field for deletion.
    *
@@ -75,6 +76,18 @@ class FieldValue {
     return SERVER_TIMESTAMP_SENTINEL;
   }
 }
+
+/*!
+ * Sentinel value for a field delete.
+ *
+ */
+ DELETE_SENTINEL = new FieldValue();
+
+/*!
+ * Sentinel value for a server timestamp.
+ *
+ */
+ SERVER_TIMESTAMP_SENTINEL = new FieldValue();
 
 module.exports = FieldValue;
 module.exports.DELETE_SENTINEL = DELETE_SENTINEL;
--- a/src/validate.js
+++ b/src/validate.js
@@ -140,5 +140,15 @@ module.exports = validators => {
     return true;
   };
 
+  exports.throwVersionMismatchErrorIfType = (obj, classType) => {
+    if (is.object(obj) && obj.constructor.name === classType.name) {
+      throw new Error(
+        `Detected an object of type '${
+          obj.constructor.name
+        }' that doesn't match the expected version for this Firestore instance. For more information, please visit: http://go...`
+      );
+    }
+  };
+
   return exports;
 };
--- a/src/document.js
+++ b/src/document.js
@@ -1332,25 +1332,39 @@ function validateFieldValue(obj, options, depth) {
     );
   }
 
-  if (obj === FieldValue.DELETE_SENTINEL) {
-    if (!options.allowNestedDeletes && (!options.allowDeletes || depth > 1)) {
-      throw new Error(
-        'Deletes must appear at the top-level and can only be used in update() or set() with {merge:true}.'
-      );
+  if (DocumentTransform.isTransformSentinel(obj)) {
+    if (obj === FieldValue.DELETE_SENTINEL) {
+      if (!options.allowNestedDeletes && (!options.allowDeletes || depth > 1)) {
+        throw new Error(
+          'Deletes must appear at the top-level and can only be used in update() or set() with {merge:true}.'
+        );
+      }
     }
-  }
-
-  if (isPlainObject(obj)) {
+  } else if (is.array(obj)) {
+    for (let prop of obj) {
+      validateFieldValue(obj[prop], options, depth + 1);
+    }
+  } else if (isPlainObject(obj)) {
     for (let prop in obj) {
       if (obj.hasOwnProperty(prop)) {
         validateFieldValue(obj[prop], options, depth + 1);
       }
     }
-  }
-  if (is.array(obj)) {
-    for (let prop of obj) {
-      validateFieldValue(obj[prop], options, depth + 1);
+  } else if (is.object(obj)) {
+    if (is.instanceof(obj, DocumentReference)) {
+      return true;
+    } else if (is.instanceof(obj, FieldValue)) {
+      return true;
     }
+
+    validate.throwVersionMismatchErrorIfType(obj, DocumentReference);
+    validate.throwVersionMismatchErrorIfType(obj, FieldValue);
+
+    throw new Error(
+      'Cannot use type (' +
+        Object.prototype.toString.call(obj) +
+        ') as a Firestore Value'
+    );
   }
 
   return true;
--- a/src/index.js
+++ b/src/index.js
@@ -1147,6 +1147,31 @@ Transaction = require('./transaction')(Firestore);
  * region_tag:firestore_quickstart
  * Full quickstart example:
  */
+
+const version = require('../package.json').version;
+
+if (!is.defined(global.__FIRESTORE_DATA)) {
+  global.__FIRESTORE_DATA = {
+    version: version,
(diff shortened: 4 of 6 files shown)


mikelehen:
What will the goo.gl link we point them to say?  Is there a sane way for them to actually keep the dependencies in sync?  Offhand I would guess that the only option would be to pin firebase-admin and @google-cloud/firestore to a very specific version in their package.json...  and then they must always change the versions together?

This means for example that if somebody updates firebase-admin, even just to a new patch release, it'll be a breaking change (their code will start hitting exceptions), which seems pretty undesirable (updating to a new patch version should be safe).

I don't have all the context, but offhand I'm not super excited about this developer experience unless I'm missing something (very possible!)...  though perhaps it's still an improvement over today (where it might be a breaking change but less obvious?).

schmidt-sebastian:
> What will the goo.gl link we point them to say? Is there a sane way for them to actually keep the dependencies in sync? Offhand I would guess that the only option would be to pin firebase-admin and to a very specific version in their package.json... and then they must always change the versions together?

I would like to link to a snippet that suggests using the admin types directly. We already suggest
```
var db = admin.firestore();
```
in https://firebase.google.com/docs/firestore/quickstart and I would like to add a section on pulling in custom types:
```
var db = admin.firestore();
const FieldPath = admin.firestore.FieldPath;
const FieldValue = admin.firestore.FieldValue;
const GeoPoint = admin.firestore.GeoPoint (which is missing in this PR right now)
```

> This means for example that if somebody updates firebase-admin, even just to a new patch release, it'll be a breaking change (their code will start hitting exceptions), which seems pretty undesirable (updating to a new patch version should be safe).

They will merely see a warning unless they use an operation that currently produces unexpected results. I think it's reasonable to accept these types from other Patch releases, but I am not sure how that would work with the current approach. Unless we figure out a way to reliably load the highest version first, the current solution would lead to a downgrade. 

We can also relax our checks even further and just look at the constructor name and the properties we are interested in. At that point, things get really hairy though and any change to any private API would lead to incompatibilities.

src/field-value.js
```diff
@@ -20,13 +20,13 @@
  * Sentinel value for a field delete.
  *
  */
-const DELETE_SENTINEL = {};
+let DELETE_SENTINEL;
```
hiranya911: Why was this necessary?

src/validate.js
```diff
@@ -140,5 +140,15 @@ module.exports = validators => {
     return true;
   };
 
+  exports.throwVersionMismatchErrorIfType = (obj, classType) => {
+    if (is.object(obj) && obj.constructor.name === classType.name) {
```
hiranya911: So if an admin object has been passed here, what would the two values be? `FieldValue` and `admin.firestore.FieldValue`?

mikelehen:
> They will merely see a warning unless they use an operation that currently produces unexpected results.

Yeah, I know, but my point is they could have code that works fine (no warnings, no crashes) that they copy/pasted from our docs somewhere, and then tomorrow they run "npm install" and now the code crashes (unless they locked down the versions in their package.json).  Perhaps this is better than today, but it's still not great.

FWIW, I think https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/ makes a strong argument that we should be using a peer dependency.

> This is where I think people tend to get a little lost, even those familiar with the peer dependency mechanism. Fortunately, the answer is relatively simple: is the dependency in question visible in any place in the package’s interface?

Basically, if the dependency is more than just an implementation detail, then it should be a peer dependency rather than a dependency.

Sorry, I think this has perhaps already been discussed to death, but what's the core opposition of making firestore a peer dependency of firebase-admin and having it throw a `@google-cloud/firestore not found. Please run: npm install --save @google-cloud/firestore` if they try to access `admin.firestore` ?

src/field-value.js
schmidt-sebastian: By making these objects constructed with `new`, they end up failing the isPlainObject() check and don't get persisted as `{}` in other clients.

src/validate.js
schmidt-sebastian: Tried this out: https://jsfiddle.net/cddxtq3a/1/. It's `FieldValue`.

schmidt-sebastian:
> Sorry, I think this has perhaps already been discussed to death, but what's the core opposition of making firestore a peer dependency of firebase-admin and having it throw a @google-cloud/firestore not found. Please run: npm install --save @google-cloud/firestore if they try to access admin.firestore ?

We have talked about it, but it's certainly worth exploring this option more. The big downside is that it bubbles up to users of Firestore functions. We would have to figure out whether this works with the dependency model used in Functions (does non-Firestore code break if the dependency is not there?). 

We already include firebase-admin as a peer dependency in firebase-functions and don't guarantee that Version X of one will work with Version Y. If a user installs the wrong version, they will see a warning - and by making Fiorestore an "optional" peer dependency we reduce the impact of these warnings.

Hiranya and I talked about splitting up the Admin SDK in general like we did for the JS SDK, but that is not on the roadmap at this point.

mikelehen:
I'm thinking that we could:
* make the firestore peer dependency optional (by not calling require() until they actually touch `admin.firestore`, and then wrapping it in a try/catch that gives the error message I suggested).
* Perhaps automatically add firestore to the package.json we create for Functions users so they get it by default. Then people can remove it if they don't actually need it (unlike today where they can't really remove it since it's an actual dependency of firebase-admin?).

src/field-value.js
hiranya911: Ok, but you can still keep them as constants right?

src/validate.js
hiranya911: When would `obj.constructor.name` and `classType.name` would not be equal?

schmidt-sebastian: Users can hit this codepath if they try to store custom objects that are initialized with new ...(). In our validator, we basically do the following:

```
if (instance FieldValue.delete) {
...
}
else if (instance FieldValue.timestamp) {
...
}
else if (!isPlainObject()) { // Check if object has a protoype
... throw 
}
```

By calling `throwVersionMismatchErrorIfType`, we can provide a better error for objects that share names with our types.

src/field-value.js
schmidt-sebastian: Once we make this a const, we can't re-assign the value. Since this value is used in the class itself, we also have to declare it before the class.

hiranya911:
> make the firestore peer dependency optional (by not calling require() until they actually touch admin.firestore, and then wrapping it in a try/catch that gives the error message I suggested).

We already do this lazy loading trick. Without it a lot of non-Firestore users were encountering issues.

mikelehen:
Ahh, nice! Though that is confusing to me since I thought currently Firestore is a non-optional dependency (https://github.com/firebase/firebase-admin-node/blob/5c4b523c181c27cf39689b010431b4133dd8e8d7/package.json#L54)...

hiranya911:
It is declared as a hard dependency. But we don't invoke (`require()`) it until a user calls `admin.firestore()`: https://github.com/firebase/firebase-admin-node/blob/master/src/firestore/firestore.ts#L112

mikelehen:
I see.  My suggestion then is to:
1. Make it a peer dependency (to avoid these version mismatches completely)
2. Wrap that require() in a try/catch to provide a slightly friendlier error message (telling them to `npm install --save @@google-cloud/firestore`).
3. Make sure that the default package.json for Functions includes @google-cloud/firestore as a dependency (so Firestore still works by default in Functions and users won't see a warning about missing a peer dependency when they `npm install`).

PR #137, opened 2018-02-16
Rejecting custom types from the API surface

schmidt-sebastian:
This cleans up our error handling for field values and adds better error messages for users that use types from other instances of the Firestore SDK.

As part of this (to get to 100% coverage), I re-added the test that transforms allPlainTypesObject into allPlainTypesProtobufJs. Along with the changes for https://github.com/googleapis/nodejs-firestore/pull/105, this meant adding support for missing (empty) array and object values.


src/path.js
```diff
@@ -484,7 +484,7 @@ class FieldPath extends Path {
   static validateFieldPath(fieldPath) {
     if (!is.instanceof(fieldPath, FieldPath)) {
       if (!is.string(fieldPath)) {
-        throw new Error(`Paths must be strings or FieldPath objects.`);
+        validate.throwCustomObjectError(fieldPath);
```
mikelehen: FYI- I think this means if you did something like `.update(3, ...)` you'd get an error saying "Cannot use custom type 'number' as a Firestore type." which is a worse error message than before.  I don't care too much though.

src/document.js
```diff
@@ -772,11 +773,7 @@ class DocumentSnapshot {
       return map;
     }
 
-    throw new Error(
-      'Cannot encode type (' +
-        Object.prototype.toString.call(val) +
-        ') to a Firestore Value'
-    );
+    validate.throwCustomObjectError(val);
```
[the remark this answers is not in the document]
schmidt-sebastian: I updated it, but I have also seen linters that do the opposite and complain when an exception isn't thrown.

src/path.js
schmidt-sebastian: This ends up being: "Update() requires either a single JavaScript object or an alternating list of field/value pairs that can be followed by an optional precondition. Argument at index 3 is not a valid FieldPath. Cannot use custom type "Object" as a Firestore type."

PR #140, opened 2018-02-17
Adding isEqual to the public API

schmidt-sebastian:
This adds isEquals to all out public API types, ignoring metadata for all *Snapshots.

The API is mostly a port from Web, with the addition of an isEqulas for WriteResult and DocumentChange.

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const assert = require('assert');
+const deepEqual = require('deep-equal');
 const is = require('is');
 
 const path = require('./path');
@@ -130,6 +131,21 @@ class GeoPoint {
     return `GeoPoint { latitude: ${this.latitude}, longitude: ${
       this.longitude
     } }`;
+  }
+
+  /**
+   * Returns true if this `GeoPoint` is equal to the provided value.
+   *
+   * @param {*} other The value to compare against.
+   * @return {boolean} true if this `GeoPoint` is equal to the provided one.
+   */
+  isEqual(other) {
+    return (
+      this === other ||
+      (is.instanceof(other, GeoPoint) &&
+        this.latitude === other.latitude &&
+        this.longitude === other.longitude)
+    );
   }
 
   /**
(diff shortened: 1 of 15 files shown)


src/write-batch.js
```diff
@@ -106,6 +106,20 @@ class WriteResult {
   get writeTime() {
     return this._writeTime;
   }
+
+  /**
+   * Returns true if this `WriteResult` is equal to the provided value.
```
mikelehen: FWIW, I can't imagine a case where this is useful so we could probably omit it... but I don't care strongly.

test/collection.js
```diff
@@ -142,4 +142,12 @@ describe('Collection interface', function() {
       assert.ok(documentRef.id.length, 20);
     });
   });
+
+  it('has equals() method', function() {
```
mikelehen: equals() => isEqual() here and elsewhere?

src/document.js
```diff
@@ -600,6 +616,23 @@ class DocumentSnapshot {
...
+   * @param {*} other The value to compare against.
+   * @return {boolean} true if this `DocumentSnapshot` is equal to the provided
+   * value.
+   */
+  isEqual(other) {
+    return (
+      this === other ||
+      (is.instance(other, DocumentSnapshot) &&
+        this._ref.isEqual(other._ref) &&
+        deepEqual(this._fieldsProto, other._fieldsProto))
```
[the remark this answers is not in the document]
schmidt-sebastian: I updated the code here and in the Query comparison to use strict equality. Note that the comparison here uses the underlying Protobuf representation and hence '3' is already different from 3. I agree though that it does make sense to not rely on this here, since I had to verify this myself first.

src/path.js
```diff
@@ -194,6 +194,20 @@ class Path {
...
+   * Returns true if this `Path` is equal to the provided value.
+   *
+   * @private
+   * @param {*} other The value to compare against.
+   * @return {boolean} true if this `Path` is equal to the provided value.
+   */
+  isEqual(other) {
+    return (
+      this === other ||
+      (is.instanceof(other, Path) && this.compareTo(other) === 0)
```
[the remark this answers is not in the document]
schmidt-sebastian: I addressed this, but I kept the override in FieldPath to generate the API documentation. The override in ResourcePath is gone.

test/collection.js
schmidt-sebastian: Done.

PR #144, opened 2018-02-26
Allowing empty merge calls

schmidt-sebastian:
This PR allows the call `set({}, merge:true)` since it can be used to ensure that a document exist on the server.

This came up as a valid use case in an internal discussion (b/73495873 - based on customer feedback in b/73132705).

There is currently a conformance test that rejects this call. I excluded the test and will work with @jba on an update.

### DIFF
--- a/src/write-batch.js
+++ b/src/write-batch.js
@@ -266,7 +266,7 @@ class WriteBatch {
 
     validate.isDocumentReference('documentRef', documentRef);
     validate.isDocument('data', data, {
-      allowEmpty: !merge,
+      allowEmpty: true,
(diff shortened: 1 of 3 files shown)


src/write-batch.js
```diff
@@ -266,7 +266,7 @@ class WriteBatch {
 
     validate.isDocumentReference('documentRef', documentRef);
     validate.isDocument('data', data, {
-      allowEmpty: !merge,
+      allowEmpty: true,
```
wilhuff: Is this allowEmpty member even required anymore?

schmidt-sebastian: It's still used for update() calls.

PR #151, opened 2018-03-08
release v0.13.0

alexander-fenster:
## Fixes

- (#146) Improving error message for multiple conflicting fields

## Updated dependencies

With this release this package started using `grpc` v1.9.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/firestore",
   "description": "Firestore Client Library for Node.js",
-  "version": "0.12.0",
+  "version": "0.13.0",
   "license": "Apache-2.0",
   "author": "Google Inc.",
   "engines": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/firestore",
   "description": "Firestore Client Library for Node.js",
-  "version": "0.12.0",
+  "version": "0.13.0",
```
stephenplusplus: According to the notes, there are only fixes in this release. Should it be 0.12.1?

alexander-fenster: That's a good question. Technically you're right, but since we had so many issues switching to `grpc` 1.8.x, and since we're jumping through 2 minor `grpc` versions at once, I made it a semver minor. I can still make it a patch release if you think it's more appropriate.

stephenplusplus: I think either way is fine. For consideration, I went with a patch release in the other libs (Bigtable, Datastore, Pub/Sub, possibly others I'm not thinking of at the moment).

Issue #143, opened 2018-02-22
Error when persisting nested objects


lfernando-silva:
#### Environment details

  - OS: Linux Mint
  - Node.js version: 8.9.4 LTS
  - npm version: 5.6.0
  - @google-cloud/firestore version: 0.11.2

#### Steps to reproduce

  1. call function ref.doc(id).set(params, options), which params is a object like: 

```javascript
{
  "notifications": {
    "web": "a random token string"
  },
  "updatedAt": 1519320011313
}
```
#### Expected behavior
- Persist the object

#### Actual behavior
- I got error `Cannot encode type ([object Object]) to a Firestore Value`. The [object Object] is the nested object ( `{web : a random token string}` ).

I realized the function isPlainObject at line 1428 returns true for {notifications...}, but returns false for the nested object {web: ....}, as it has no Object.prototype property. 

//firestore/src/document.js 1428
```javascript
function isPlainObject(input) {
  return (
    typeof input === 'object' &&
    input !== null &&
    Object.getPrototypeOf(input) === Object.prototype
  );
}
```
When I tried to force input to be an object
```javascript
function isPlainObject(input) {
  return (
    typeof input === 'object' &&
    input !== null &&
(Object.getPrototypeOf(input) === Object.prototype) || (Object.getPrototypeOf(Object.create(input)) === Object.prototype)
  );
}
```
I got another error for the other value at not nested object: 
`Argument \"data\" is not a valid Document. Object prototype may only be an Object or null: 1519320772620`.

Using lodash 's isObject method:
```javascript
function isPlainObject(input) {
    return require('lodash').isObject(input) //returns true for both objects
}
```
I got the error
`Argument \"data\" is not a valid Document. obj.hasOwnProperty is not a function`

How can I handle this problem? Thank you in advance.

colombotapps:
I had the same problem. My solution was to convert the nested objects to plain objects using lodash before sending them to Firestore. For example, instead of something like this:
```
let params = new ParamsObject();
params. notifications = new WebObject();
```
I convert the objects to plain objects using lodash before sending them to Firestore.
```
let params = _.toPlainObject(new ParamsObject());
params. notifications = _.toPlainObject(new WebObject());
```
This works, but if there is a better way to solve this problem, I would also like to know.

schmidt-sebastian:
The following code snippet works for me:

```
let ref = randomCol.doc('doc1');
    return ref
      .set({
        notifications: {
          web: 'a random token string',
[...]

PR #162, opened 2018-03-21
Improved error messages for custom types

schmidt-sebastian:
This addresses https://github.com/googleapis/nodejs-firestore/issues/143

This PR changes two of our error messages:

- When we detect an custom object during serialization, we throw: 'Firestore doesn't support serialization of JavaScript objects that were constructed using the 'new' operator.'
- When we detect an invalid type as method argument, we throw: 'Cannot use object of type "FieldPath" as a Firestore argument.'

There might be much better ways to express this, and the easiest way to figure out if the changes here make sense is to look at the test cases that show these error messages. 

Note that the code tested in `order.js` is not reachable from customer code. The tests exists to allow for 100% coverage (touches line https://github.com/googleapis/nodejs-firestore/blob/463fbfc5f70afa3d776382348f077a0a4b369c29/src/order.js#L82)

### DIFF
--- a/src/validate.js
+++ b/src/validate.js
@@ -142,21 +142,22 @@ module.exports = validators => {
 
   exports.customObjectError = val => {
     if (is.object(val) && val.constructor.name !== 'Object') {
-      switch (val.constructor.name) {
+      const typeName = val.constructor.name;
+      switch (typeName) {
         case 'DocumentReference':
         case 'FieldPath':
         case 'FieldValue':
         case 'GeoPoint':
           return new Error(
-            `Detected an object of type "${
-              val.constructor.name
-            }" that doesn't match the expected instance. Please ensure that ` +
-              'the Firestore types you are using are from the same NPM package.'
+            `Detected an object of type "${typeName}" that doesn't match the ` +
+              'expected instance. Please ensure that the Firestore types you ' +
+              'are using are from the same NPM package.'
           );
         default:
           return new Error(
-            "Firestore doesn't support serialization of JavaScript objects " +
-              "that were constructed using the 'new' operator."
+            `Couldn't serialize object of type "${typeName}". Firestore ` +
+              "doesn't support JavaScript objects with custom prototypes " +
+              "(i.e. objects that were created via the 'new' operator)."
           );
       }
     } else {
--- a/src/validate.js
+++ b/src/validate.js
@@ -141,21 +141,28 @@ module.exports = validators => {
   };
 
   exports.customObjectError = val => {
-    let typeName = is.object(val) ? val.constructor.name : typeof val;
-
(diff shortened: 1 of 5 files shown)


src/validate.js
```diff
@@ -141,21 +141,28 @@ module.exports = validators => {
...
+          return new Error(
+            `Detected an object of type "${
+              val.constructor.name
+            }" that doesn't match the expected instance. Please ensure that ` +
+              'the Firestore types you are using are from the same NPM package.'
+          );
+        default:
+          return new Error(
+            "Firestore doesn't support serialization of JavaScript objects " +
+              "that were constructed using the 'new' operator."
```
mikelehen: Interesting phrasing.  I'm not 100% sure if this will make sense to users or not.  I have two concerns:
1. "constructed using the 'new' operator" is kind of a weird differentiator for what works or not. I'm not sure if it'll resonate / makes sense for folks.
2. I wonder if anybody will get confused and think that `{ date: new Date(), location: new GeoPoint(...) }` might be disallowed since it uses new().

Is there a reason you removed the type name from the error?  I think that would address 2.  So I'm imagining an error something like:

> Couldn't serialize ${typeName} object.  Firestore doesn't support JavaScript objects with custom prototypes (i.e. that were created via the 'new' operator).

WDYT?  And can you make a similar change or open a bug to make the web client error match this?

test/document.js
```diff
@@ -290,13 +290,13 @@ describe('serialize document', function() {
...
       firestore.doc('collectionId/documentId').set({foo: undefined});
-    }, /Cannot use custom type "undefined" as a Firestore type./);
+    }, /Invalid use of type "undefined" as a Firestore argument./);
 
     assert.throws(() => {
       firestore
         .doc('collectionId/documentId')
         .set({foo: Firestore.FieldPath.documentId()});
-    }, /Cannot use "FieldPath" as a Firestore type./);
+    }, /Cannot use object of type "FieldPath" as a Firestore value./);
```
mikelehen: FWIW, I'd kinda' expect to have  a test case here that hits the new "doesn't support serialization of JavaScript objects that were constructed using the 'new' operator."  (I know it's covered the query.where() test case elsewhere, but it seems much more likely users will use custom objects with set() and so that's the error message I'm most interested in seeing tested...

src/validate.js
schmidt-sebastian: I removed the type name from the error message as the error got somewhat lengthy. That's probably not a good enough reason, so I put it back. I also mostly went with your re-wording in general.

test/document.js
schmidt-sebastian: Added this error case to this test as well. That should now give me 101% coverage, right? :)

PR #166, opened 2018-03-26
API Proposal to list all documents

schmidt-sebastian:
Proposal PR to list all documents in a collection (including missing).

Alternatives considered:

- Add overloads to CollectionReference (for Query.get()/.stream())  that take the option `includeMissing`. Any call to customize a query (where, order, limit, ...) method returns a Query instance with a Query.get() that doesn't take arguments. In not so strongly typed languages, this might create confusion... and it also means that we have to revert all QueryDocumentSnapshot changes.

- Adding a getDocuments() that takes a FieldMask. This gives users two ways to specify masks (via select() and getDocuments()).

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -1000,6 +1000,41 @@ declare namespace FirebaseFirestore {
      * @return true if this `CollectionReference` is equal to the provided one.
      */
     isEqual(other: CollectionReference): boolean;
+
+    /**
+     * Creates and returns a new CollectionReference instance that applies a
+     * field mask and only returns the specified subset of fields for calls to
+     * `get()`, `stream()`, `getAll()` and `streamAll()`. You can specify a list
+     * of field paths to return, or use an empty list to only return the
+     * references of matching documents.
+     *
+     * This function returns a new (immutable) instance of the
+     * CollectionReference (rather than modify the existing instance) to impose
+     * the field mask.
+     *
+     * @param field The field paths to return.
+     * @return The created Query.
+     */
+    select(...field: (string | FieldPath)[]): CollectionReference;
+
+    /**
+     * Retrieves all documents in this CollectionReference.  Documents with
+     * subcollections that contain no other data are returned as
+     * `DocumentSnapshots` whose `exists` property returns false.
+     *
+     * @return A Promise that will be resolved with the contents of of this
+     * collection.
+     */
+    getAll(): Promise<DocumentSnapshot>;
+
+    /*
+     * Retrieves all documents in this CollectionReference in a Node Stream.
+     * Documents with subcollections that contain no other data are returned as
+     * `DocumentSnapshots` whose `exists` property returns false.
+     *
+     * @return A stream of DocumentSnapshot.
+     */
+    streamAll(): NodeJS.ReadableStream;
   }
 
   /**
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -78,6 +78,17 @@ declare namespace FirebaseFirestore {
     getAll(...documentRef: DocumentReference[]): Promise<DocumentSnapshot[]>;
 
     /**
+     * Retrieves multiple documents from Firestore.
+     *
+     * @param options An object to configure the behavior of this `getAll` call
+     * (i.e. the set of fields to return).
+     * @param documentRef The `DocumentReferences` to receive.
+     * @return A Promise that resolves with an array of resulting document
+     * snapshots.
+     */
+    getAll(options:GetOptions, ...documentRef: DocumentReference[]): Promise<DocumentSnapshot[]>;
+
+    /**
      * Fetches the root collections that are associated with this Firestore
      * database.
      *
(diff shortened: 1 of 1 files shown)


types/firestore.d.ts
```diff
@@ -1000,6 +1000,41 @@ declare namespace FirebaseFirestore {
...
+     * references of matching documents.
+     *
+     * This function returns a new (immutable) instance of the
+     * CollectionReference (rather than modify the existing instance) to impose
+     * the field mask.
+     *
+     * @param field The field paths to return.
+     * @return The created Query.
+     */
+    select(...field: (string | FieldPath)[]): CollectionReference;
```
gsoltis: This seems odd. I would expect a `CollectionReference` to be a reference to a collection, and not also a filter on fields of documents that I might later get from that collection. If I had a vanilla `CollectionReference` and one that had been returned via `select()`, how would I tell them apart?

schmidt-sebastian:
I reworked this proposal per our offline discussion. This is the most 'drastic' change in terms of functionality, but it works around the issues in some of the other proposals and also allows us to easily add more flags to document fetches (i.e. support for ancestor queries).

schmidt-sebastian:
@mikelehen / @wilhuff: There are versions of this proposal in this PR. You can look at the first commit to see if you prefer it, but it does have the issues that Greg pointed out in his comment.

types/firestore.d.ts
```diff
@@ -526,6 +551,22 @@ declare namespace FirebaseFirestore {
...
+     * @param options An object to configure the behavior of this snapshot
+     * listener (i.e. the set of fields to return).
+     * @param onNext A callback to be called every time a new `DocumentSnapshot`
+     * is available.
+     * @param onError A callback to be called if the listen fails or is
+     * cancelled. No further callbacks will occur.
+     * @return An unsubscribe function that can be called to cancel
+     * the snapshot listener.
+     */
+    onSnapshot(options: GetOptions,
```
mikelehen: Using GetOptions with onSnapshot() seems like a naming mismatch.

types/firestore.d.ts
```diff
@@ -724,6 +765,10 @@ declare namespace FirebaseFirestore {
      * This function returns a new (immutable) instance of the Query (rather
      * than modify the existing instance) to impose the field mask.
      *
+     * @deprecated Use `get(options:GetOptions)` or `stream(options:GetOptions)`
+     * to specify a field mask. The field mask specified via `select()` does not
+     * apply to `getAll()`/`streamAll()`.
```
mikelehen: Er, yikes.  Deprecating select() in order to support this new use case seems like the tail wagging the dog a bit.

types/firestore.d.ts
```diff
@@ -1001,40 +1066,28 @@ declare namespace FirebaseFirestore {
...
-     * `DocumentSnapshots` whose `exists` property returns false.
+     * `DocumentSnapshots` whose `exists` property is set to false.
      *
+     * @param options An object to configure the behavior of this `getAll` call
+     * (i.e. the set of fields to return).
      * @return A Promise that will be resolved with the contents of of this
      * collection.
      */
-    getAll(): Promise<DocumentSnapshot>;
+    getAll(options?:GetOptions): Promise<DocumentSnapshot>;
```
mikelehen: I'm not sure I'm a fan of re-using the name getAll(). This method seems substantively different to me since the arguments are different (it enumerates all documents rather than retrieving a set of specified documents).

types/firestore.d.ts
```diff
@@ -1000,6 +1065,29 @@ declare namespace FirebaseFirestore {
...
+     * Retrieves all documents in this CollectionReference.  Documents with
+     * subcollections that contain no other data are returned as
+     * `DocumentSnapshots` whose `exists` property is set to false.
+     *
+     * @param options An object to configure the behavior of this `getAll` call
+     * (i.e. the set of fields to return).
+     * @return A Promise that will be resolved with the contents of of this
+     * collection.
+     */
+    getAll(options?:GetOptions): Promise<DocumentSnapshot>;
```
mikelehen: Should this be `Promise<DocumentSnapshot[]>` ?

types/firestore.d.ts
```diff
@@ -1000,6 +1065,29 @@ declare namespace FirebaseFirestore {
...
+    /*
+     * Retrieves all documents in this CollectionReference in a Node Stream.
+     * Documents with subcollections that contain no other data are returned as
+     * `DocumentSnapshots` whose `exists` property is set to false.
+     *
+     * @param options An object to configure the behavior of this `streamAll`
+     * call (i.e. the set of fields to return).
+     * @return A stream of DocumentSnapshots.
+     */
+    streamAll(options?:GetOptions): NodeJS.ReadableStream;
```
mikelehen: Is this streaming all the way back to the backend?  So I could read this "slowly" (e.g. 1 document per second) and it won't download the entire list?  And I can stop at any time?  (I'm trying to figure out if we think this is a suitable replacement for pagination).

PR #170, opened 2018-03-28
chore: one more workaround for repo-tools EPERM

alexander-fenster:
Sometimes it just happens, only in CircleCI and never reproduced. Here is a proof that this `chmod` fixes the problem: https://circleci.com/gh/googleapis/nodejs-speech/1376 - let's apply this to all our repos and see if it fails or not.

This PR fixes system tests and sample tests which were missed by previous PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,12 @@ jobs:
             fi
       - run:
           name: Install codecov.
-          command: npm install codecov
+          command: |-
+            npm install
+            repo_tools="node_modules/@google-cloud/nodejs-repo-tools/bin/tools"
+            if ! test -x "$repo_tools"; then
+              chmod +x "$repo_tools"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -90,7 +90,12 @@ jobs:
             fi
       - run:
           name: Install codecov.
-          command: npm install codecov
+          command: |-
+            npm install
+            repo_tools="node_modules/@google-cloud/nodejs-repo-tools/bin/tools"
+            if ! test -x "$repo_tools"; then
+              chmod +x "$repo_tools"
+            fi codecov
```
callmehiphop: @alexander-fenster looks like CI is getting caught up on this line

JustinBeckwith:
@alexander-fenster can you take another look at this one?

PR #172, opened 2018-04-05
Making Transforms smarter

schmidt-sebastian:
This PR prepares the NodeJS client to handle merges with field masks and upcoming field transforms.

All of the changes in here a purely a refactor. The next PR will add functionality on top of this refactor that takes advantage of the new class structure (such as DocumentMask.removeFields()).


src/document.js
```diff
@@ -1167,15 +1173,12 @@ class DocumentTransform {
...
+      if (val instanceof FieldTransform && val.includeInDocumentTransform) {
         if (allowTransforms) {
-          transforms.push({
-            fieldPath: path.formattedName,
-            setToServerValue: SERVER_TIMESTAMP,
-          });
+          transforms.set(path, val);
         } else {
           throw new Error(
             'Server timestamps are not supported as array values.'
```
gsoltis: Maybe this error should change now too, since this code is no longer specific to Server Timestamps?

src/field-value.js
```diff
@@ -57,7 +45,7 @@ class FieldValue {
    * });
    */
   static delete() {
-    return DELETE_SENTINEL;
+    return new DeleteTransform();
```
gsoltis: Kinda minor, so feel free to ignore, but you could have a singleton for the transform instance, right? Instead of allocating a new one each time it's used?

src/field-value.js
```diff
@@ -92,9 +80,86 @@ class FieldValue {
...
+ * An internal interface shared by all field transforms.
+ *
+ * A 'FieldTransform` subclass should implement '.includeInDocumentMask',
+ * '.includeInDocumentTransform' and 'toProto' (if '.includeInDocumentTransform'
+ * is 'true').
+ *
+ * @private
+ * @abstract
+ */
+class FieldTransform extends FieldValue {}
```
gsoltis: Should the parent type include `includeInDocumentMask` and `includeInDocumentTransform`?

You seem to be assuming that they exist for all `FieldTransform` instances.

src/field-value.js
schmidt-sebastian: I reverted this back to the way it was before and added singletons.

src/document.js
schmidt-sebastian: Made this error less specific.

src/field-value.js
schmidt-sebastian: I was going to point this out in the PR description, but forgot: With linting enabled and 100 code coverage as a requirement, it is not actually possible to add 'abstract' methods (aka: fake them by adding an unused method that throws an exception). When I move this codebase to TypeScript, this will become an interface, which will solve both of these issues.

PR #173, opened 2018-04-05
Adding support for "mergeFields"

schmidt-sebastian:
This PR adds support for "mergeFields".

We also now support 100% of all conformance tests. A fair number of changes in this PR is geared to make these tests pass - including not sending any data to the backend that is excluded via a field mask.

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -20,6 +20,7 @@ const assert = require('assert');
 const deepEqual = require('deep-equal');
 const is = require('is');
 
+const fieldValue = require('./field-value');
 const path = require('./path');
 const timestampFromJson = require('./convert').timestampFromJson;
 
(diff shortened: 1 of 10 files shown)


src/document.js
```diff
@@ -1097,6 +1114,121 @@ class DocumentMask {
...
+  applyTo(data) {
+    /*!
+     * Applies this DocumentMask to 'data' and computes the list of field paths
+     * that were specified in the mask but not present in 'data'.
+     */
+    const applyDocumentMask = data => {
+      const remainingPaths = this._sortedPaths.slice(0);
+
+      const processObject = (currentData, currentPath) => {
+        let result = null;
```
gsoltis: Does this mean that `applyTo` can return `null`? Does the final line of the function need to be `return result.filteredData || {}` ?

gsoltis: Thinking more about this, maybe either comment about why it can't happen, or add a test that causes it to happen and verify that it's handled.

schmidt-sebastian: I added a test for this. The function does return null, but since I am only using "for ... in" loops to process this data, nothing blew up. Yay.

Either way, I think we should not rely on this :) I changed this function to return an empty object instead of null. Note that while I am processing the nested objects, I still have to return null as the intermediate state.

PR #174, opened 2018-04-11
Fixing API doc

schmidt-sebastian:
Fixes some small issues I found while browsing the API docs.

### DIFF
--- a/src/document.js
+++ b/src/document.js
@@ -177,8 +177,7 @@ class GeoPoint {
  * A DocumentSnapshot is an immutable representation for a document in a
  * Firestore database. The data can be extracted with
  * [data()]{@link DocumentSnapshot#data} or
- * [get(fieldPath)]{@link DocumentSnapshot#get} to get a
- * specific field.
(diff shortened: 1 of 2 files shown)


src/document.js
```diff
@@ -177,8 +177,7 @@ class GeoPoint {
  * A DocumentSnapshot is an immutable representation for a document in a
  * Firestore database. The data can be extracted with
  * [data()]{@link DocumentSnapshot#data} or
- * [get(fieldPath)]{@link DocumentSnapshot#get} to get a
- * specific field.
+ * [get(fieldPath)]{@link DocumentSnapshot#get} (to get a specific field).
```
gsoltis: As discussed offline, consider a separate sentence or other structure instead of adding `()`s, here and throughout.

schmidt-sebastian: I've been trying to come up with a 2-sentence based approach that doesn't seem to clunky. I went back to what we had before for now, and just fixed the one actual issue in these docs.

Issue #185, opened 2018-05-03
No grpc pre-built binary


jonataswalker:
#### Environment details

  - OS: Arch Linux 4.16.6-1
  - Node.js version: 10.0.0
  - npm version: 6.0.0
  - @google-cloud/common-grpc version: 0.6.0

#### Steps to reproduce

`npm i @google-cloud/firestore`

There is a [pre-built binary](https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v1.10.0/node-v64-linux-x64-glibc.tar.gz) for `grpc@1.11.1`. And [there isn't one](https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v1.10.0/node-v64-linux-x64-glibc.tar.gz) for `grpc@1.10.0`. So to be able to install `@google-cloud/firestore` on Node 10 `@google-cloud/common-grpc` should upgrade `grpc` dependency to `1.11.0` and `@google-cloud/firestore` should upgrade `@google-cloud/common-grpc` dependency to the latest one.

Also created https://github.com/googleapis/nodejs-common-grpc/issues/44

Thanks.

PR #189, opened 2018-05-08
Updating dependencies

schmidt-sebastian:
This brings all of our dependencies up to date.

Fixes https://github.com/googleapis/nodejs-firestore/issues/185


schmidt-sebastian:
The tests are currently failing because of an issue in https://github.com/googleapis/nodejs-common/blob/master/src/util.ts#L754

We need to get a null check into this line.

schmidt-sebastian:
> We need to get a null check into this line.

PR pending: https://github.com/googleapis/nodejs-common/pull/130

schmidt-sebastian:
Yes, the system test are passing for me.

PR #197, opened 2018-05-19
Lazy-load all GRPC dependencies

schmidt-sebastian:
This PR is a fork of https://github.com/googleapis/nodejs-firestore/pull/196 and addresses a performance issue on Google Cloud Functions. Firestore usage on GCF is often read-only - that is, they provide the payload for a DocumentSnapshot and use Firestore only to deserialize this payload. We don't need to initialize the networking stack unless the user ends up writing back to Firestore.

This PR lazy-loads v1beta1 and `@google-cloud/common`.

The PR **also removes `@google-cloud/common-grpc`** (as this also pulls in the GRPC dependency). Unfortunately, this means that we no longer inherit from GrpcService. Is that a problem?

Test code:
```
const start = Date.now();
var Firestore = require('@google-cloud/firestore');
var snapshot = new Firestore().snapshot_('projects/foo/databases/bar/documents/coll/missing', {});
const end = Date.now();
console.log('Load time:', (end-start), 'ms');
```

Before:  473 ms,  After: 34ms on a fully beefed-up Macbook

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@google-cloud/firestore",
-  "version": "0.13.1",
+  "version": "0.14.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


alexander-fenster:
I'm not the one making the call here - @crwilcox is responsible for the libraries and @stephenplusplus and @callmehiphop support the common package.

schmidt-sebastian:
I confirmed that we are not using any of these features (by adding breakpoints), since all of our requests go through the Veneer layer. I have also tested a patched client that includes these changes and am able to write and read documents.

PR #220, opened 2018-06-25
Adding Timestamp class

schmidt-sebastian:
This ports the new Timestamp class to Node JS.

Also:
- Added GeoPoint range validation.
- Fixed two docs issues.

 A follow-up PR will use the new Timestamp type in the .createTime, .updateTime and .readTime.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -216,7 +216,7 @@ class Firestore {
    * `Timestamp` instead.
    * <br/>NOTE: in the future `timestampsInSnapshots: true` will become the
    * default and this option will be removed so you should change your code to
-   * use Timestamp now and opt-in to this new behavior as soon as you can.
+   * use `Timestamp` now and opt-in to this new behavior as soon as you can.
    */
   constructor(options) {
     options = extend({}, options, {
--- a/src/index.js
+++ b/src/index.js
@@ -197,6 +197,26 @@ const GRPC_UNAVAILABLE = 14;
 class Firestore {
   /**
    * @param {Object=} options - [Configuration object](#/docs).
+   * @param {string=} options.projectId The Firestore Project ID. Can be
+   * omitted in environments that support `Application Default Credentials`
+   * {@see https://cloud.google.com/docs/authentication}
+   * @param {string=} options.keyFilename Local file containing the Service
+   * Account credentials. Can be omitted in environments that support
+   * `Application Default Credentials`
+   * {@see https://cloud.google.com/docs/authentication}
+   * @param {boolean=} options.timestampsInSnapshots Enables the use of
+   * `Timestamp`s for timestamp fields in `DocumentSnapshots`.<br/>
+   * Currently, Firestore returns timestamp fields as `Date` but `Date` only
+   * supports millisecond precision, which leads to truncation and causes
+   * unexpected behavior when using a timestamp from a snapshot as a part
+   * of a subsequent query.
+   * <br/>Setting `timestampsInSnapshots` to true will cause Firestore to return
+   * `Timestamp` values instead of `Date` avoiding this kind of problem. To
+   * make this work you must also change any code that uses `Date` to use
+   * `Timestamp` instead.
+   * <br/>NOTE: in the future `timestampsInSnapshots: true` will become the
+   * default and this option will be removed so you should change your code to
+   * use Timestamp now and opt-in to this new behavior as soon as you can.
    */
   constructor(options) {
     options = extend({}, options, {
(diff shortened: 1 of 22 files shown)


src/index.js
```diff
@@ -197,6 +197,26 @@ const GRPC_UNAVAILABLE = 14;
...
+   * supports millisecond precision, which leads to truncation and causes
+   * unexpected behavior when using a timestamp from a snapshot as a part
+   * of a subsequent query.
+   * <br/>Setting `timestampsInSnapshots` to true will cause Firestore to return
+   * `Timestamp` values instead of `Date` avoiding this kind of problem. To
+   * make this work you must also change any code that uses `Date` to use
+   * `Timestamp` instead.
+   * <br/>NOTE: in the future `timestampsInSnapshots: true` will become the
+   * default and this option will be removed so you should change your code to
+   * use Timestamp now and opt-in to this new behavior as soon as you can.
```
var-const: Small nit: surround `Timestamp` with backticks?

src/timestamp.js
```diff
@@ -0,0 +1,152 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const is = require('is');
+const validate = require('./validate')();
```
var-const: Question: what is the effect of the empty braces here?

src/timestamp.js
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Returns 'true' if this `Timestamp` is equal to the provided one.
+   *
+   * @param {any} other The `Timestamp` to compare against.
+   * @return {boolean} 'true' if this `Timestamp` is equal to the provided one.
+   */
+  isEqual(other) {
+    return (
+      this === other ||
```
var-const: Question: is there a case where the first check (`this === other`) succeeds but the second check doesn't? If not, is the first part necessary?

src/validate.js
```diff
@@ -52,12 +52,24 @@ module.exports = validators => {
       integer: (value, min, max) => {
         min = is.defined(min) ? min : -Infinity;
         max = is.defined(max) ? max : Infinity;
-        return is.integer(value) && value >= min && value <= max;
+        if (!is.integer(value)) {
+          return false;
+        }
+        if (value < min || value > max) {
+          throw new Error(`Value must be within [${min}, ${max}] inclusive.`);
```
var-const: Optional: add `value` to the error message? (for example, append `, but was ${value}.`.

test/document.js
```diff
@@ -447,19 +448,33 @@ describe('serialize document', function() {
...
     assert.throws(() => {
-      new Firestore.GeoPoint('INVALID', '-4.4499982');
-    }, expectedErr);
+      new Firestore.GeoPoint('INVALID', -4.4499982);
+    }, /Argument "latitude" is not a valid number/);
 
     assert.throws(() => {
       new Firestore.GeoPoint();
-    }, expectedErr);
+    }, /Argument "latitude" is not a valid number/);
```
var-const: I'm slightly against specifying exact error text in tests, but feel free to keep as is.

types/firestore.d.ts
```diff
@@ -39,6 +39,46 @@ declare namespace FirebaseFirestore {
...
+     * of a subsequent query.
+     *
+     * Setting `timestampsInSnapshots` to true will cause Firestore to return
+     * `Timestamp` values instead of `Date` avoiding this kind of problem. To
+     * make this work you must also change any code that uses `Date` to use
+     * `Timestamp` instead.
+     *
+     * NOTE: in the future `timestampsInSnapshots: true` will become the
+     * default and this option will be removed so you should change your code to
+     * use Timestamp now and opt-in to this new behavior as soon as you can.
```
var-const: Same nitpick about backticks here.

test/timestamp.js
```diff
@@ -0,0 +1,220 @@
...
+      {timestampsInSnapshots: true},
+      DOCUMENT_WITH_TIMESTAMP
+    ).then(firestore => {
+      return firestore
+        .doc('coll/doc')
+        .get()
+        .then(res => {
+          const timestamp = res.get('moonLanding');
+          assert.equal(timestamp.seconds, -14182920);
+          assert.equal(timestamp.nanoseconds, 123000123);
```
var-const: In client SDKs, I have deliberately added code to truncate a timestamp to microseconds as soon as it is given to Firestore, so that truncation works the same offline and online. Is there a reason not to do the same in server SDKs?

var-const: Ah, I forgot that server SDKs don't have a local mode. Sorry about that.

schmidt-sebastian:
Adding @stephenplusplus / @JustinBeckwith for approval.

types/firestore.d.ts
```diff
@@ -1067,6 +1107,91 @@ declare namespace FirebaseFirestore {
...
+     * resolution. Negative second values with fractions must still have
+     * non-negative nanoseconds values that count forward in time. Must be from
+     * 0 to 999,999,999 inclusive.
+     */
+    constructor(seconds: number, nanoseconds: number);
+
+    /**
+     * The number of seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z.
+     */
+    readonly seconds: number;
```
Edalbrelord: Hi,
Sorry for commenting on a merged pull request, but it seems like the js object has a field called "_seconds" and another "_nanoseconds".
Whenever I create a Timestamp by using Timestamp.now() in TS, it is creating an object with {seconds:..., nanoseconds: ...}, not having the underscores. When I try to save this to a firestore repository it fails with the following message:
"Error: Argument "data" is not a valid Document. Detected an object of type "Timestamp" that doesn't match the expected instance. Please ensure that the Firestore types you are using are from the same NPM package."
Should the ts definition have underscores in the names of these fields as well?

schmidt-sebastian: Hi! Are you using the Admin SDK? If you are using the Admin SDK, you have to use the Admin-exported type of Timestamp (`admin.firestore.Timestamp`). That, however, is not possible until the next release of the Admin SDK, which will include the definition for this type (as per https://github.com/firebase/firebase-admin-node/pull/309).

Edalbrelord:
I am using the Admin SDK yes, what release date would you be looking at for this new SDK?

hiranya911:
@Edalbrelord we can get a patch release out sometime next week.

PR #222, opened 2018-06-25
Update dependencies

schmidt-sebastian:
Addresses https://github.com/firebase/firebase-admin-node/issues/275 by upgrading ProtobufJS.


JustinBeckwith:
We're going to need to land #221 first.  @crwilcox do you mind prioritizing this one so we unblock?

PR #232, opened 2018-07-02
Adding Typescript support

schmidt-sebastian:
Adding Typescript support to Firestore.

Note:
- I only converted `path.js` to TS for now (as a proof of concept)
- I did not (yet) run `gts fix` since it re-formats almost everything (which is why as of now all tests are failing)

The tests are also spewing errors about Google Default Credentials not working. I am not sure if this is new in this PR, but it doesn't (yet) cause problems. It might make sense to fix that in a follow up.


JustinBeckwith:
Wow, this is a big PR 🤣   Usually when we do a TypeScript transition, we're very careful to take a minimum number of steps.  I have those pretty well laid out here:
https://github.com/googleapis/nodejs-translate/issues/51

The issue with any library that uses GAPIC is that we're using generated JavaScript code.  How are you getting around that out of curiosity?

.nycrc
```diff
@@ -1,9 +1,9 @@
 {
   "report-dir": "./.coverage",
   "exclude": [
-    "src/*{/*,/**/*}.js",
-    "src/*/v*/*.js",
-    "test/**/*.js"
+    "build/src/*{/*,/**/*}.js",
```
JustinBeckwith: Do we want to include any of `builld/`?

package.json
```diff
@@ -8,14 +8,11 @@
     "node": ">=4.0.0"
   },
   "repository": "googleapis/nodejs-firestore",
-  "main": "./src/index.js",
+  "main": "./build/src/index.js",
+  "types": "./types/firestore.d.ts",
```
JustinBeckwith: Generally these end up in an `./build/src/index.d.ts`, but I'll keep reading :)

package.json
```diff
@@ -39,16 +35,23 @@
...
     "lint": "eslint src/ system-test/ test/ conformance/",
     "prettier": "prettier --write src/*.js src/**/*.js test/*.js test/**/*.js system-test/*.js system-test/**/*.js conformance/*.js conformance/**/*.js",
     "publish-module": "node ../../scripts/publish.js firestore",
-    "system-test": "mocha system-test/*.js --timeout 600000",
-    "conformance": "mocha conformance/*.js --no-timeouts",
-    "test-no-cover": "mocha test/*.js --no-timeouts",
-    "test": "npm run cover"
+    "system-test": "mocha build/system-test/*.js --timeout 600000",
+    "conformance": "mocha build/conformance/*.js --no-timeouts",
+    "test-no-cover": "mocha build/test/*.js --no-timeouts",
```
JustinBeckwith: `--no-timeouts` can cause real big misses in the tests - doesn't need to happen here, but we should file a follow up to get rid of it.

src/path.ts
```diff
@@ -229,24 +219,10 @@ class ResourcePath extends Path {
    * @param {...string|string[]} segments - Sequence of names of the parts of
    * the path.
    */
-  constructor(projectId, databaseId, segments) {
-    segments = is.array(segments)
-      ? segments
-      : Array.prototype.slice.call(arguments, 2);
-
+  constructor(
+      private readonly projectId: string, private readonly databaseId: string,
```
JustinBeckwith: I'm personally a little 👎 on the fancy constructor assignment. @ofrobots do we have anything about this in the style guide?

src/path.ts
```diff
@@ -398,8 +354,8 @@ class ResourcePath extends Path {
    * path.
    * @returns {ResourcePath} The newly created ResourcePath.
    */
-  construct(segments) {
-    return new ResourcePath(this._projectId, this._databaseId, segments);
+  construct(segments: string[]): ResourcePath {
+    return new ResourcePath(this.projectId, this.databaseId, ...segments);
```
JustinBeckwith: I'm curious why you're switching from `segments` to `...segments` here.

src/path.ts
```diff
@@ -410,27 +366,25 @@ class ResourcePath extends Path {
...
     // Ignore DocumentReference with {{projectId}} placeholders and assume that
     // the resolved IDs match the provided ResourcePath. We could alternatively
     // try to resolve the Project ID here, but this is asynchronous as it
     // requires Disk I/O.
-    if (
-      this._projectId !== '{{projectId}}' &&
-      other._projectId !== '{{projectId}}'
-    ) {
-      if (this._projectId < other._projectId) {
+    if (this.projectId !== '{{projectId}}' &&
```
JustinBeckwith: Why drop the underscore here?

src/path.ts
ofrobots: @JustinBeckwith your sensibilities are opposed to what the style guide recommends: http://go/tsstyle#parameter-properties. I'm with the style guide on this one.

src/path.ts
```diff
@@ -460,12 +419,12 @@ class FieldPath extends Path {
...
    * });
    */
-  constructor(segments) {
+  constructor(...segments: string[]) {
     validate.minNumberOfArguments('FieldPath', arguments, 1);
 
-    segments = is.array(segments)
-      ? segments
-      : Array.prototype.slice.call(arguments);
+    segments = segments.length === 1 && is.array(segments[0]) ?
```
ofrobots: This change looks semver major. Previously `FieldPath([], 42)` would behave differently than what it does now.

Not sure if this is user-exposed though. If this internal then 👍.

tsconfig.json
```diff
@@ -1,24 +1,20 @@
...
-    "noImplicitAny": true,
-    "noImplicitReturns": true,
-    "noImplicitThis": true,
-    "noFallthroughCasesInSwitch": true,
-    "strictNullChecks": true,
-    "noStrictGenericChecks": true
+    "rootDir": ".",
+    "outDir": "build",
+    "target": "es6",
+    "noImplicitAny": false
```
ofrobots: nit: likewise, a TODO comment that this should be removed eventually.

tsconfig.json
```diff
@@ -1,24 +1,20 @@
 {
+  "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
     "allowJs": true,
-    "target": "es6",
-    "moduleResolution": "node",
-    "outDir": "dist/es6",
-    "rootDir": "./",
+    "declaration": false,
```
ofrobots: nit: Just adding a note that in the final state after the refactoring, this should be removed. Most tools that read `tsconfig.json` are okay with a trailing `// comment` style comment on the same line. It would be worth added a TODO here.

.nycrc
JustinBeckwith: 👍

.prettierrc
```diff
@@ -5,4 +5,4 @@ semi: true
 singleQuote: true
 tabWidth: 2
 trailingComma: es5
-useTabs: false
+useTabs: false
```
JustinBeckwith: nit: add this return back :)

package.json
JustinBeckwith: 👍

.prettierrc
schmidt-sebastian: Done.

tsconfig.json
schmidt-sebastian: Added. I wonder though if this will ever go away since we will likely keep the generated client as is.

tsconfig.json
schmidt-sebastian: Done

PR #237, opened 2018-07-03
Adding GTS Formatting

schmidt-sebastian:


package.json
```diff
@@ -39,8 +39,6 @@
     "cover": "nyc --reporter=lcov mocha --require intelli-espower-loader build/test/*.js && nyc report",
     "docs": "jsdoc -c .jsdoc.js",
     "generate-scaffolding": "repo-tools generate all",
-    "lint": "eslint src/ system-test/ test/ conformance/",
```
JustinBeckwith: We kept these around in other repos because of samples.  Does this repo keep samples in it?

schmidt-sebastian: It does, but they have their own package.json:

https://github.com/googleapis/nodejs-firestore/blob/master/samples/package.json

Should I add it back either way?

schmidt-sebastian:
BTW, the Node 4 tests are now failing because of: 

```
/home/node/project/node_modules/gts/node_modules/meow/index.js:66
	const {pkg} = options;
```

Should we disable the Node4 runner (and the lint runner?) and abandon Node 4? It would allow me to add the spread operator back :-)

JustinBeckwith:
Yes!

schmidt-sebastian:
I removed Node4/Lint from CI, but it looks like I can't remove the checks from GitHub. @JustinBeckwith can you do this?

JustinBeckwith:
Apologies for not saying this out loud, but we should really disable nodejs 4 and 9 in a different pull request (just to keep the commits separate)

schmidt-sebastian:
I moved the commits that disabled Node 4 to #241

schmidt-sebastian:
Merged package-lock.json. This still seems to be blocked on the `lint` step.

PR #243, opened 2018-07-08
refactor: drop dependency on nodejs-common

JustinBeckwith:
This drops the dependency on nodejs-common, and uses a new tiny module cassed nodejs-projectify.  This drops the dependency on request in this module.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -213,28 +213,6 @@
         "arrify": "^1.0.1"
       }
     },
-    "@google-cloud/common": {
-      "version": "0.20.3",
-      "resolved": "https://registry.npmjs.org/@google-cloud/common/-/common-0.20.3.tgz",
-      "integrity": "sha512-jt8/R4EqDTQccv5WA9AEaS65llM5+mlxsuWu57G5Os8HTIpgPbcsOVMUeIvmTrBuPUYSoRIMW8d/pvv/95n0+g==",
-      "requires": {
-        "@types/duplexify": "^3.5.0",
-        "@types/request": "^2.47.0",
-        "arrify": "^1.0.1",
-        "axios": "^0.18.0",
-        "duplexify": "^3.6.0",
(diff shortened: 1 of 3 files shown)


jkwlui (approved):

kinwa91:
Fails on node4, do we need to drop support?

schmidt-sebastian:
Nice, this will allow me to clean up some of our `lazy-loading` code quite a bit.

schmidt-sebastian (approved):

ofrobots:
QQ, as I am not too familiar with this library. I see that we were using 'projectify' to remove the projectId token from request parameters.. who introduces the token in the first place. I though that was `@google-cloud/common` as well.

JustinBeckwith:
This module ain't my rodeo :) . @schmidt-sebastian any idea?

schmidt-sebastian:
> This module ain't my rodeo :) . @schmidt-sebastian any idea?

The project ID gets set to {{projectId} at client init time. We then build proto-requests with {{projectId}} embedded and replace it as we send the first request. Only the replacement happens in `@google-cloud/common`.

ofrobots:
IMO, such token pasting/replacement is a bad idea. `projectify` does mangling on the request object and can do data corruption for user data. This was probably copied from the other client libraries, but it was a bad idea there too.

Instead, can you make all operations that require a project ID to depend (`.then` or `await`) on a thing that acquires the promise? That should be a lot easier now that you're moving to TypeScript and can use async/await.

There are plans to make the same change in other client libraries. I expect that `projectify` will be deprecated once that is done.

This does not have to block this PR, but I would recommend moving away from using `projectify` sooner rather than later.

schmidt-sebastian:
I had the same thoughts as I was writing it, but it seems like doing this conversion would require some non-trivial effort, that I would like to tackle soon (but not now).

PR #251, opened 2018-07-11
Adding TypeScript dev dependencies

schmidt-sebastian:
Adding `ts-node`, `chai` and `chai-as-promised`. To be used in https://github.com/googleapis/nodejs-firestore/pull/242

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "google-proto-files": "^0.16.1",
     "is": "^3.2.1",
     "lodash.merge": "^4.6.1",
+    "pkg-up": "^2.0.0",
     "through2": "^2.0.3"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -68,8 +68,11 @@
     "@google-cloud/nodejs-repo-tools": "^2.3.0",
     "@types/mocha": "^5.2.3",
     "@types/node": "^10.3.5",
+    "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
     "codecov": "^3.0.2",
     "duplexify": "^3.6.0",
+    "gts": "^0.7.1",
     "ink-docstrap": "git+https://github.com/docstrap/docstrap.git",
     "intelli-espower-loader": "^1.0.1",
     "jsdoc": "^3.5.5",
@@ -78,7 +81,7 @@
     "power-assert": "^1.6.0",
     "protobufjs": "^6.8.6",
     "proxyquire": "^2.0.1",
-    "typescript": "^2.9.2",
-    "gts": "^0.7.1"
+    "ts-node": "^7.0.0",
+    "typescript": "^2.9.2"
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -19,17 +19,11 @@
 const bun = require('bun');
 const extend = require('extend');
 const is = require('is');
+const pkgUp = require('pkg-up');
 const through = require('through2');
 const util = require('util');
 
-let libVersion;
-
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -68,8 +68,11 @@
     "@google-cloud/nodejs-repo-tools": "^2.3.0",
     "@types/mocha": "^5.2.3",
     "@types/node": "^10.3.5",
+    "chai": "^4.1.2",
```
JustinBeckwith: Are you using chai and chai-as-promised? This seems a little weird.

package.json
```diff
@@ -78,7 +81,7 @@
     "power-assert": "^1.6.0",
     "protobufjs": "^6.8.6",
     "proxyquire": "^2.0.1",
-    "typescript": "^2.9.2",
-    "gts": "^0.7.1"
+    "ts-node": "^7.0.0",
```
JustinBeckwith: Where/how are you using tsnode?

package.json
schmidt-sebastian: This PR is meant to make https://github.com/googleapis/nodejs-firestore/pull/242 smaller. I want to start using `chair` in my tests. 

See https://github.com/googleapis/nodejs-firestore/pull/242/files#diff-f626ff6b37edf53ad3138d2e458d5f44R41

package.json
schmidt-sebastian: This allows me to run the TypeScript code in-place from within IntelliJ. 
It's also possible to just install this globally and not add `ts-node` to the dev-dependencies. 
I was debating committing the IntelliJ test runners as well, though, similar to what we have done in the main Firestore repo (https://github.com/firebase/firebase-js-sdk/tree/master/packages/firestore/.idea/runConfigurations). These test runner rely on `ts-node` as an initialization step.

src/index.js
```diff
@@ -22,7 +22,14 @@ const is = require('is');
 const through = require('through2');
 const util = require('util');
 
-const libVersion = require('../../package.json').version;
+let libVersion;
```
JustinBeckwith: Eugh, this is less than idea.  Is this because of the use of `ts-node`?  In most other places, we just do the import to the relative path assuming it's running from `build/src`.

JustinBeckwith: @ofrobots for thoughts here too

schmidt-sebastian: Yeah, this is pretty bad, but I don't know a good way around it. I even had to flip the order to get 100% test coverage back...

There are two problems here:
- `ts-node` skips the build step, which makes it super easy to work with (it's a huge productivity boost), but the source needs to find the package.json (unless I set it to an unknown value during development?).
- I can't copy the package.json into the `build/` folder since NPM will then recursively apply these settings. `npm pack`, for example, will then not copy the right files and try to include "build/build/src" instead.

I hope there are better ways to solve these two problems!

ofrobots: In https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs we had to deal with a similar issue where some files (compiled protos) need to be able to resolvable relative to both from compiled and uncompiled sources. To deal with this we added an extra top directory (`/ts`) so that the number of `..`'s needed to get to `package.json` is constant.

Another alternative would be to use something like [pkg-up](https://github.com/sindresorhus/pkg-up) and write the code as:

```ts
const pkgUp = require('pkg-up');
const libVersion = require(await pkgUp());  // top-level await, where art thou?
```

WDYT?

schmidt-sebastian: I do like the work-around that you described by adding another level of folders, but it would be a huge pain to merge my pending PRs. So instead I went with pkg-up (which also offers a synchronous API). The loading of the package version is now much cleaner!

PR #254, opened 2018-07-12
Test initialization refactor

schmidt-sebastian:

- [ ] Tests and linter pass


test/timestamp.js
```diff
@@ -122,7 +115,11 @@ describe('timestamps', function() {
 
   it('convert to date', function() {
     return createInstance(
-               {timestampsInSnapshots: true}, DOCUMENT_WITH_TIMESTAMP)
+               {
+                 timestampsInSnapshots: true,
+                 keyFilename: './test/fake-certificate.json',
```
var-const: Just curious: why is `keyFilename` passed here, but not in `'returned when enabled`?

test/transaction.js
```diff
@@ -224,66 +213,58 @@ function query(transaction) {
...
-    assert.deepEqual(actual, request.request);
-    return request.stream;
-  };
-
-  firestore._firestoreClient._innerApiCalls.runQuery = function(actual) {
-    request = requests.shift();
-    assert.equal(request.type, 'query');
-    assert.deepEqual(actual, request.request);
-    return request.stream;
+function runTransaction(callback, ...requests) {
```
var-const: What's the reason to support multiple requests now?

test/write-batch.js
```diff
@@ -368,29 +361,28 @@ describe('batch support', function() {
...
             seconds: 0,
           },
-        });
-      };
+        })
+      }
+    };
+
+    return createInstance(overrides).then(firestore => {
+      firestore._preferTransactions = true;
```
var-const: Why is it necessary to set these flags now?

test/util/helpers.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @return {Firestore} A new Firestore client.
+ */
+export function createInstance(
+    apiOverrides?: ApiOverride, firestoreSettings?: {}): Promise<Firestore> {
+  const initializationOptions = Object.assign(
+      {
+        projectId: PROJECT_ID,
+        sslCreds: SSL_CREDENTIALS,
+        timestampsInSnapshots: true,
+        keyFilename: './test/fake-certificate.json',
```
var-const: I think some of these parameters (for example, `keyFilename`) are set to the same values in some of the tests, is it necessary?

test/timestamp.js
schmidt-sebastian: I meant to remove `keyFilename` from these callers and only set it in `createInstance()`. I removed it here and below.

test/transaction.js
schmidt-sebastian: These are the proto requests issued by the transaction that are used in the tests:

```
  it('support get with query', function() {
    return runTransaction((transaction, docRef) => {
      let query = docRef.parent.where('foo', '==', 'bar');
      return transaction.get(query).then(results => {
        assert.equal(results.docs[0].id, 'documentId');
      });
    }, begin(), query(), commit());
  });
```
In this case, we issue a beingTransaction, a query and a commit request.

I added a comment to explain this better.

test/write-batch.js
schmidt-sebastian: `firestore` didn't exist beforehand.

test/util/helpers.ts
schmidt-sebastian: Almost all the tests go through `createInstance`. I scrubbed the codebase once more and removed some of the extra assignments. There are a few places left where we don't use createInstance to explicitly test the various options.

schmidt-sebastian:
Changed base branch to master and added @stephenplusplus for owners approval.

PR #256, opened 2018-07-12
Allowing 101 listeners

schmidt-sebastian:
The GAPIC client that handles all of Firestore's networking is limited to 100 connections. This PR lift this limit by replacing the static GAPIC client with a pool of clients. If more than 100 operations run concurrently, we create additional clients.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -133,7 +133,11 @@ const CLOUD_RESOURCE_HEADER = 'google-cloud-resource-prefix';
 const MAX_REQUEST_RETRIES = 5;
 
 /*!
- * The maximum number of concurrent requests supported by a GAPIC client.
+ * The maximum number of concurrent requests supported by a single GRPC channel,
+ * as enforced by Google's Frontend. If the SDK issues more than 100 concurrent
+ * operations, we need to use more than one GAPIC client since these client
+ * multiplex all requests over a single channel.
+ *
  * @type {number}
  */
 const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
@@ -251,7 +255,8 @@ class Firestore {
     });
 
     /**
-     * A client pool to distribute requests over multiple GAPIC clients.
+     * A client pool to distribute requests over multiple GAPIC clients in order
+     * to work around a connection limit of 100 concurrent requests per client.
      * @private
      * @type {ClientPool|null}
      */
--- a/src/index.js
+++ b/src/index.js
@@ -20,17 +20,11 @@ const assert = require('assert');
 const bun = require('bun');
 const extend = require('extend');
 const is = require('is');
+const pkgUp = require('pkg-up');
 const through = require('through2');
 const util = require('util');
 
-let libVersion;
-
-// Allow Firestore to be run both from 'src' and 'build/src'.
-try {
-  libVersion = require('../package.json').version;
-} catch (err) {
-  libVersion = require('../../package.json').version;
-}
+const libVersion = require(pkgUp.sync()).version;
 
 const path = require('./path');
 const convert = require('./convert');
@@ -135,7 +129,7 @@ const MAX_REQUEST_RETRIES = 5;
 /*!
  * The maximum number of concurrent requests supported by a single GRPC channel,
  * as enforced by Google's Frontend. If the SDK issues more than 100 concurrent
- * operations, we need to use more than one GAPIC client since these client
+ * operations, we need to use more than one GAPIC client since these clients
  * multiplex all requests over a single channel.
  *
  * @type {number}
@@ -855,7 +849,7 @@ follow these steps, YOUR APP MAY BREAK.`);
    *
    * @private
    * @param {number} attemptsRemaining - The number of available attempts.
-   * @param {string} requestTag A unique client-assigned identifier for this
+   * @param {string} requestTag - A unique client-assigned identifier for this
    * request.
    * @param {retryFunction} func - Method returning a Promise than can be
    * retried.
--- a/src/index.js
+++ b/src/index.js
@@ -16,6 +16,7 @@
 
 'use strict';
 
+const assert = require('assert');
 const bun = require('bun');
 const extend = require('extend');
 const is = require('is');
@@ -71,6 +72,11 @@ const FieldValue = require('./field-value').FieldValue;
 const Timestamp = require('./timestamp');
 
 /*!
+ * @see ClientPool
+ */
+const ClientPool = require('./pool').ClientPool;
+
+/*!
  * @see CollectionReference
  */
 let CollectionReference;
@@ -106,12 +112,12 @@ let Transaction;
 /*!
  * @see v1beta1
  */
-let v1beta1;  // Lazy-loaded in `_ensureClient()`
+let v1beta1;  // Lazy-loaded in `_runRequest()`
 
 /*!
  * @see @google-cloud/common
  */
-let common;  // Lazy-loaded in `_ensureClient()`
+let common;  // Lazy-loaded in `_runRequest()`
 
 /*!
  * HTTP header for the resource prefix to improve routing and project isolation
@@ -125,6 +131,12 @@ const CLOUD_RESOURCE_HEADER = 'google-cloud-resource-prefix';
  * @type {number}
  */
 const MAX_REQUEST_RETRIES = 5;
+
+/*!
+ * The maximum number of concurrent requests supported by a GAPIC client.
+ * @type {number}
+ */
+const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
 
 /*!
  * GRPC Error code for 'UNAVAILABLE'.
@@ -239,11 +251,11 @@ class Firestore {
     });
 
     /**
+     * A client pool to distribute requests over multiple GAPIC clients.
      * @private
-     * @type {object|null}
-     * @property {FirestoreClient} Firestore The Firestore GAPIC client.
+     * @type {ClientPool|null}
      */
-    this._firestoreClient = null;
+    this._clientPool = null;
 
     /**
      * The configuration options for the GAPIC client.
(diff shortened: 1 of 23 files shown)


src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+          'The provided client factory returned an existing instance');
+    }
+
+    this.activeClients.set(selectedClient, currentRequestCount + 1);
+
+    return selectedClient!;
+  }
+
+  /**
+   * Reduces the number of operation for the provided client, potentially
```
var-const: Nit: s/operation/operations

test/pool.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Copyright 2017 Google Inc. All Rights Reserved.
```
var-const: Shouldn't this be 2018?

test/pool.ts
```diff
@@ -0,0 +1,158 @@
...
+  it('creates new instances as needed', () => {
+    const clientPool = new ClientPool<{}>(3, () => {
+      return {};
+    });
+
+    expect(clientPool.size).to.eq(0);
+
+    const deferred = deferredPromises(4);
+
+    clientPool.run(() => deferred[0].promise);
```
var-const: Very optional: perhaps use a `for` loop here (and in similar places below)?

test/pool.ts
```diff
@@ -0,0 +1,158 @@
...
+    completionPromises.push(clientPool.run(() => operationPromises[0].promise));
+    expect(clientPool.size).to.eq(1);
+    completionPromises.push(clientPool.run(() => operationPromises[1].promise));
+    expect(clientPool.size).to.eq(1);
+    completionPromises.push(clientPool.run(() => operationPromises[2].promise));
+    expect(clientPool.size).to.eq(2);
+    completionPromises.push(clientPool.run(() => operationPromises[3].promise));
+    expect(clientPool.size).to.eq(2);
+
+    operationPromises[0].resolve();
```
var-const: Use `forEach`?

src/pool.ts
schmidt-sebastian: Fixed.

test/pool.ts
schmidt-sebastian: Yes, thanks for catching this.

test/pool.ts
schmidt-sebastian: ~~Uhm. Yes, of course.~~

I just realized that this makes the tests harder to follow since each loop iteration would have to assert a slightly different state. For that reason, I left it as is.

test/pool.ts
schmidt-sebastian: Done (here and below)

test/util/helpers.ts
```diff
@@ -19,14 +19,33 @@
...
+export class Deferred<R> {
+  promise: Promise<R>;
+  resolve: (value?: R|Promise<R>) => void = () => {};
+  reject: (reason?: Error) => void = () => {};
+
+  constructor() {
+    this.promise = new Promise(
+        (resolve: (value?: R|Promise<R>) => void,
+         reject: (reason?: Error) => void) => {
+          this.resolve = resolve;
```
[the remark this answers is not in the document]
schmidt-sebastian: This is the same deferred Promise code we also use in the Web SDK.  These deferred promises allow you to resolve a Promise outside of their assignment. Usually, you would resolve a Promise like this:

```
const promise = new Promise((resolve, reject) => {
   ... do something async
   resolve();
}):
```

If you want to resolve the Promise at a later point you can do something like this:

```
let resolveFnct;

const promise = new Promise((resolve, reject) => {
   ... do something async
   resolveFnct = resolve;
}):

resolveFnct();
```

The `Deferred` class just reduces this boilerplate.

schmidt-sebastian:
Mike, since you have more context on this client, do you mind taking a look at this as well?

src/index.js
```diff
@@ -126,6 +132,12 @@ const CLOUD_RESOURCE_HEADER = 'google-cloud-resource-prefix';
  */
 const MAX_REQUEST_RETRIES = 5;
 
+/*!
+ * The maximum number of concurrent requests supported by a GAPIC client.
```
mikelehen: Can you clarify that this is a limitation of the GAPIC client (as opposed to a maximum we've decided to impose)?  And if there's any docs / code you can link to regarding the limit, that would be useful context too.

src/index.js
```diff
@@ -239,11 +251,11 @@ class Firestore {
     });
 
     /**
+     * A client pool to distribute requests over multiple GAPIC clients.
```
mikelehen: Perhaps make it clear that we do this to work around the limit in the GAPIC client.

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * An auto-resizing pool that distributes concurrent operations over multiple
+ * clients of type `T`.
+ *
+ * ClientPool is used within Firestore to manage a pool of GAPIC clients and
+ * automatically initializes multiple clients if we issue more than 100
+ * concurrent operations.
+ */
+export class ClientPool<T> {
+  private activeClients: Map<T, number> = new Map();
```
mikelehen: /** Stores each active clients and how many operations it has outstanding. */

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+   * of concurrent operations or initializes and returns a new client.
+   */
+  private acquire(): T {
+    let selectedClient: T|null = null;
+    let currentRequestCount = 0;
+
+    this.activeClients.forEach((requestCount, client) => {
+      if (!selectedClient && requestCount < this.concurrentOperationLimit) {
+        selectedClient = client;
+        currentRequestCount = requestCount;
```
mikelehen: minor, but I'd rename currentRequestCount => selectedClientRequestCount

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+    return selectedClient!;
+  }
+
+  /**
+   * Reduces the number of operations for the provided client, potentially
+   * removing it from the pool of active clients.
+   */
+  private release(client: T): void {
+    let currentRequestCount = this.activeClients.get(client);
+    assert(currentRequestCount! > 0, 'Active client not found');
```
mikelehen: Hrm. Is this using `!` to take advantage of the fact that `undefined > 0` is legal in JS and evaluates to false?

I would rather `requestCount !== undefined && requestCount > 0` or since you use `!` again below, maybe you should just do `requestCount = this.activeClients.get(client) || 0;`

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    return selectedClient!;
+  }
+
+  /**
+   * Reduces the number of operations for the provided client, potentially
+   * removing it from the pool of active clients.
+   */
+  private release(client: T): void {
+    let currentRequestCount = this.activeClients.get(client);
```
mikelehen: current seems unnecessary. If you want an extra word in there, I think it should be `clientRequestCount`

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+   * operation limit.
+   *
+   * @param {(client: T) => Promise<V>} op - A callback function that returns a
+   * Promise. The client T will be returned to the pool when callback finishes.
+   * @return {Promise<V>} A Promise that resolves with the result of `op`.
+   */
+  run<V>(op: (client: T) => Promise<V>): Promise<V> {
+    const client = this.acquire();
+
+    return op(client)
```
mikelehen: I think node's Promise implementation has `.finally(() => this.release(client))`...

src/pool.ts
```diff
@@ -0,0 +1,134 @@
...
+        })
+        .then(res => {
+          this.release(client);
+          return res;
+        });
+  }
+
+  /**
+   * Deletes clients that are no longer executing operations. Keeps up to one
+   * idle client to reduce future initialization costs.
```
mikelehen: FWIW, I suspect we could get away without any intelligent shrinking of the pool...  clients probably aren't very expensive (especially since there are idle timeouts on the connections, right?)...  so if we wanted we could have a more naive pool implementation.  But since you've gone to the trouble and it's not too complicated, there's probably no reason not to proceed with your fancy resizing pool stuff. :-)

system-test/firestore.js
```diff
@@ -23,7 +23,13 @@ let DocumentReference = require('../src/reference').DocumentReference;
 let DocumentSnapshot =
     require('../src/document')(DocumentReference).DocumentSnapshot;
 
-let version = require('../../package.json').version;
+let version;
+try {
+  version = require('../../package.json').version;
+} catch (e) {
+  version = require('../package.json').version;
+}
```
mikelehen: uhh... this seems comment-worthy?

src/index.js
schmidt-sebastian: I have only seen this confirmed in a Google groups post and this is actually a limit in both the GFE and the clients. I expanded on the comment and further shifted the blame on everyone else but us.

src/index.js
schmidt-sebastian: Done.

src/pool.ts
schmidt-sebastian: Done.

src/pool.ts
schmidt-sebastian: Makes sense. Done.

src/pool.ts
schmidt-sebastian: I removed "current".

src/pool.ts
schmidt-sebastian: Yeah, I was trying to take advantage of all the little hacks JavaScript supports. I went with your suggestion and moved the undefined-handling to the first line of the function, which gets rid of the "!".

src/pool.ts
schmidt-sebastian: Finally!

No, unfortunately not. It's an optional feature in Node 8 that can be turned on with harmony mode. We support back to Node 6 (we just dropped Node 4!).

src/pool.ts
schmidt-sebastian: This pool will likely be used by millions of other consumer in NPM land. I am sure that one of them will need this garbage collection. :)

system-test/firestore.js
schmidt-sebastian: This will go away when I merge to master. We just replaced this hack with `pkg-up`.

PR #270, opened 2018-07-17
Adding Array-Contains

schmidt-sebastian:
This adds `array-contains` support to the Server SDK. The only "tricky" part of this change is that we need to exclude array-contains when we parse inequalities for our implicit orderBys.

### DIFF
--- a/.jsdoc.js
+++ b/.jsdoc.js
@@ -31,7 +31,7 @@ module.exports = {
   source: {
     excludePattern: '(^|\\/|\\\\)[._]',
     include: [
-      'src'
+      'build/src'
     ],
     includePattern: '\\.js$'
   },
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/firestore",
   "description": "Firestore Client Library for Node.js",
-  "version": "0.15.2",
+  "version": "0.15.4",
   "license": "Apache-2.0",
   "author": "Google Inc.",
   "engines": {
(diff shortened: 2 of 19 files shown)


laurentpayot:
This feature appeared in Firebase JavaScript SDK 5.3.0 but is still undocumented. Does it make [maps of values queries](https://firebase.google.com/docs/firestore/solutions/arrays#solution_a_map_of_values) a thing of the past? How efficient is it compared checking a map key? Is the array parsed sequentialy? Should we use this only for small arrays?
Many questions…

schmidt-sebastian:
> This feature appeared in Firebase JavaScript SDK 5.3.0 but is still undocumented. Does it make maps of values queries a thing of the past? How efficient is it compared to checking a map key? Is the array parsed sequentialy? Should we use this only for small arrays?
> Many questions…

The new Array features will be documented publicly soon. We are still waiting for a few items before publishing a broader announcement. We believe that these features will address much of the need of the "maps of values" approach, but we only offer access by value ("does this value exist?" / "remove all instances of this value" / "add an instance of this value if it doesn't already exist").

Our new Array features are backed by indices. Hence, we do not perform a sequential scan, but your arrays are capped by the maximum number of indexable elements in a Firestore document, which is 20,000 single-direction indices. In practice, as you will have other fields (or arrays) in your documents, your limits will likely be in the 1000s. See here: https://firebase.google.com/docs/firestore/query-data/index-overview#index_limitations

PR #274, opened 2018-07-18
Convert FieldValue to Typescript

schmidt-sebastian:
In preparation for ArrayUnion/ArrayRemove.

`protos/firestore_proto_api.d.ts` is copied from the Web client.

### DIFF
--- a/protos/firestore_proto_api.d.ts
+++ b/protos/firestore_proto_api.d.ts
@@ -1,0 +1,1201 @@
+/**
+ * Copyright 2018 Google Inc.
+ *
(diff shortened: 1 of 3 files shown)


JustinBeckwith:
Isn't this generated code!?

schmidt-sebastian:
> Isn't this generated code!?

- 'protos/firestore_proto_api.d.ts' is generated from inside Google3.
- 'src/field-value.js' is our own.

schmidt-sebastian:
> LGTM. Do we still need the closure type annotations?

We actually use the JSDoc types for our documentation still (well, as of one PR ago, but I am sure that you can see into the future).

Issue #277, opened 2018-07-18
Firestore 0.15.2 unnecessarily looks up Application Default Credentials 


hiranya911:
We initialize Firestore as follows:

```
var firestoreDatabase = require('@google-cloud/firestore');
const options = { 
  credentials:  { 
    private_key: pk_string,
    client_email: client_email_string
  },
  projectId
}
return new firestoreDatabase(options);
```

This used to work fine in older versions of Firestore. But it seems starting from 0.15.2, Firestore attempts to look up application default credentials even when initialized with explicit credentials.

```
Error: The file at /foo/bar does not exist, or it is not a file. Error: ENOENT: no such file or directory, lstat '/foo'
    at GoogleAuth.createError (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:516:16)
    at GoogleAuth.<anonymous> (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:366:32)
    at step (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:47:23)
    at Object.next (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:28:53)
    at /Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:22:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:18:12)
    at GoogleAuth._getApplicationCredentialsFromFilePath (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:348:16)
    at GoogleAuth.<anonymous> (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:291:48)
    at step (/Users/hkj/Projects/firebase-admin-node/temp/node_modules/google-auth-library/build/src/auth/googleauth.js:47:23
```

Here `/foo/bar` is the value of my `GOOGLE_APPLICATION_CREDENTIALS` environment variable, and it is deliberately set to a bogus value to reproduce the issue.

PR #279, opened 2018-07-18
Fixing Client initialization

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/277

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -826,7 +826,7 @@ follow these steps, YOUR APP MAY BREAK.`);
     const clientPool =
         new ClientPool(MAX_CONCURRENT_REQUESTS_PER_CLIENT, () => {
           const gapicClient =
-              module.exports.v1beta1(this._initalizationOptions);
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -826,7 +826,7 @@ follow these steps, YOUR APP MAY BREAK.`);
     const clientPool =
         new ClientPool(MAX_CONCURRENT_REQUESTS_PER_CLIENT, () => {
           const gapicClient =
-              module.exports.v1beta1(this._initalizationOptions);
+              module.exports.v1beta1(this._initalizationSettings);
```
hiranya911: What can we do to avoid something like this happening again? Can we somehow cover this in unit tests? Or how about an assertion here:

```
if (typeof this._initializationSettings === 'undefined') {
   throw new Error();
}
```

schmidt-sebastian: 1) Use TypeScript everywhere.
2) Run the system tests without application default credentials in the background. I am going to see how we can add that to CI.

PR #282, opened 2018-07-19
Generate docs from the generated files

schmidt-sebastian:
This uses the transpiled *.js files as an input to JSDoc.

AFAIK, this looks like before. I removed some of the private typings in the TypeScript docs, since JSDoc did not understand them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
   ],
   "scripts": {
     "codecov": "nyc report --reporter=json && codecov -f .coverage/*.json",
-    "docs": "npm run compile && jsdoc -c .jsdoc.js",
+    "predocs": "npm run compile",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
   ],
   "scripts": {
     "codecov": "nyc report --reporter=json && codecov -f .coverage/*.json",
-    "docs": "jsdoc -c .jsdoc.js",
+    "docs": "npm run compile && jsdoc -c .jsdoc.js",
```
JustinBeckwith: You can also do a `"predocs": "npm run compile"` step to do this :)

schmidt-sebastian: I learn something every day :)

PR #283, opened 2018-07-19
test: fix non-strict assertion

kinwa91:


kinwa91:
had to run `npm run fix` because I guess `assert.equal => assert.strictEqual` pushed some lines over 80?

PR #284, opened 2018-07-20
Using import instead of require

schmidt-sebastian:
For arrayUnion()/arrayRemove() I need to import DocumentSnapshot. I was hoping that I could get around the circular import hell with this PR. It turns out that doesn't quite help, but the PR is already written, so I propose we do the `require` to `import` migration either way.

This PR also removes the default exports from our internal modules. We only default export Firestore now.

### DIFF
--- a/conformance/runner.js
+++ b/conformance/runner.js
@@ -16,23 +16,28 @@
 
 'use strict';
 
-const assert = require('power-assert');
-const path = require('path');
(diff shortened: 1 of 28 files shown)


schmidt-sebastian:
The Web client uses the much prettier `prettier`, and this repo uses `clang-format` now. `clang-format` removes those spaces.

PR #287, opened 2018-07-23
Adding Array Transforms

schmidt-sebastian:
This PR adds arrayUnion() and arrayRemove() support to the Node client.

It also fixes the error message that is used in query evaluation. We now return: 

`Argument "value" is not a valid QueryValue. FieldValue.method() can only be used in set(), create() or update().`

since `Argument "value" is not a valid FieldValue. FieldValue.method()...` is a little bit non-sensical.

Unfortunately, this PR still uses the old-style dependency injection.

### DIFF
--- a/system-test/firestore.js
+++ b/system-test/firestore.js
@@ -255,6 +255,60 @@ describe('DocumentReference class', function() {
         });
   });
 
+  it('supports arrayUnion()', function() {
+    const baseObject = {
+      a: [],
+      b: ['foo'],
+      c: {d: ['foo']},
+    };
+    const updateObject = {
+      a: Firestore.FieldValue.arrayUnion('foo', 'bar'),
+      b: Firestore.FieldValue.arrayUnion('foo', 'bar'),
+      'c.d': Firestore.FieldValue.arrayUnion('foo', 'bar')
+    };
+    const expectedObject = {
+      a: ['foo', 'bar'],
+      b: ['foo', 'bar'],
+      c: {d: ['foo', 'bar']},
+    };
+
+    let ref = randomCol.doc('doc');
+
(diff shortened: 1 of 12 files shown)


system-test/firestore.js
```diff
@@ -255,6 +255,60 @@ describe('DocumentReference class', function() {
         });
   });
 
+  it('supports arrayUnion()', function() {
```
gsoltis: For this test and the one below, maybe add the case where the field is not already an array?

types/firestore.d.ts
```diff
@@ -1065,23 +1065,56 @@ declare namespace FirebaseFirestore {
...
+     * that already exists on the server. Each specified element that doesn't
+     * already exist in the array will be added to the end. If the field being
+     * modified is not already an array it will be overwritten with an array
+     * containing exactly the specified elements.
+     *
+     * @param elements The elements to union into the array.
+     * @return The FieldValue sentinel for use in a call to set(), create() or
+     * update().
+     */
+    static arrayUnion(...elements: any[]): FieldValue;
```
gsoltis: Does it make sense to have a public equivalent of `AnyJS`?

system-test/firestore.js
schmidt-sebastian: I am not 100% opposed, but somewhat lazy. The reason I didn't add these tests is that this is purely server behavior, and nothing in this client would ever influence the type conversion. I am going to leave this untested for now and rely on the backend to do the right thing.

types/firestore.d.ts
schmidt-sebastian: Let's talk about this in the API discussion, but I would like to keep this file as similar to the firestore.d.ts file we use for the Web client.

PR #290, opened 2018-07-24
Using the generated Proto definitions

schmidt-sebastian:
This PR updates the Node client to use a generated `.d.ts` file for its Proto definition (previously it was only semi-generated).

Major changes include:
- `valueType` is no longer set for the `Value` proto.
- Instead of passing in `REQUEST_TIME`, we use the enum value `1`.


ofrobots:
Are any of the changes in src/ directory user exposed? That is, is this semver-major?

schmidt-sebastian:
> Are any of the changes in src/ directory user exposed? That is, is this semver-major?

No, these changes are all internal.

Than being said, we are releasing this with the 0.16.x, as we added new API surface area in previous PRs.

PR #294, opened 2018-07-27
Manually editing .d.ts files to use 'string' for large numbers

schmidt-sebastian:
This changes the Proto type for large numbers to the type we use in code. `pbjs` uses `number|Long`, but we use `number|string`.

I wasn't able to find the right setting in `pbjs` to make it generate the right settings. There is `--force-long ` but no `--force-string`, and I much rather deal with strings in the API than with the Long type (we also use strings in the Web Firestore SDK)


schmidt-sebastian:
PTAL @JustinBeckwith  Thanks :)

JustinBeckwith:
I'd like it if @ofrobots and @kjin would take a look here :)  They know way more about pbjs than I do.

PR #295, opened 2018-07-27
Convert order.js to order.ts

schmidt-sebastian:
This PR converts order.js to TypeScript.


src/order.ts
```diff
@@ -16,78 +16,74 @@
 
 'use strict';
 
+import assert from 'assert';
```
JustinBeckwith: This looks like you have synthetic imports enabled.  You *really* want to turn those off and use the `import * as assert from 'assert'` style syntax.  These leak in to user end apps, and can cause compilation issues for users.

JustinBeckwith: Wait - on second thought - why are we using assert in non-test code?

schmidt-sebastian: I am using this here: https://github.com/googleapis/nodejs-firestore/pull/295/files#diff-8ea67e6b6a3211ad6928d19f123570c8R162

It's an internal assert and I essentially want to check-fail the client if this fails. I can a) not do this b) throw an Error c) use non-synthetic imports. What's your pick? :)

schmidt-sebastian: I have removed this assert for now so this PR can be committed.

PR #319, opened 2018-08-13
Adding explicit dependency on ProtobufJS

schmidt-sebastian:
The generated code from Protobuf JS has a dependency on "protobufjs", so we should have one too.

Fixes: https://github.com/firebase/firebase-admin-node/issues/338


Ghostavio:
I started having problems with this on Yarn 2

schmidt-sebastian:
@Ghostavio If you are still experiencing problems, do you mind opening an issue? Thank you!

PR #320, opened 2018-08-14
Reject FieldTransforms from FieldTransforms

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-admin-node/issues/336

This PR is a little less than ideal. The input validation only happens when during the set()/update() call, and not when Jacob initially calls `arrayUnion()`. This is because we don't have access to validate.isFieldValue() from the field transforms, since it would create a circular dependency. I debated having  two versions of FieldValue (a public one and an internal one that does have access), but am not sold on the added complexity.

### DIFF
--- a/src/field-value.ts
+++ b/src/field-value.ts
@@ -312,7 +312,7 @@ class ArrayUnionTransform extends FieldTransform {
   validate(validator: AnyDuringMigration): boolean {
     let valid = true;
     for (let i = 0; valid && i < this.elements.length; ++i) {
-      valid = validator.isArrayTransform(
+      valid = validator.isArrayElement(
           i, this.elements[i], {allowDeletes: 'none', allowTransforms: false});
     }
     return valid;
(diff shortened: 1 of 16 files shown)


src/field-value.ts
```diff
@@ -18,7 +18,7 @@
 
 import deepEqual from 'deep-equal';
 
-import {AnyJs} from './types';
+import {AnyDuringMigration, AnyJs} from './types';
```
mikelehen: Hrm...  I'm surprised to see AnyDuringMigration exists here.  That type is usually used during automated upgrades of TypeScript to keep existing code working without rewriting it...  But we shouldn't be adding new usages of it.

src/index.js
```diff
@@ -208,6 +208,8 @@ class Firestore {
     });
 
     this._validator = new Validator({
+      ArrayTransform: (name, value) =>
```
mikelehen: This results in an `isArrayTransform()` function that actually validates that a value can be used _in_ an array transform, not that it _is_ an array transform.

Should this be `ArrayTransformElement`?  Or is this generic enough (since you can pass options to allow/disallow deletes, etc.) that it should just be `FieldValue` ?

Issue #316, opened 2018-08-12
Error: Cannot find module "."


mhdevnet:
#### Environment details

  - OS: WSL or WIndows 10 PowerShell
  - Node.js version: 9..8.0
  - npm version: 6.2.0
  - `@google-cloud/firestore` version: 0.16.0

#### Steps to reproduce

  1. Create Netlify functions app using Netlify Functions: https://www.netlify.com/docs/functions/
  2. Install firebase-admin
  3. Import firebase-admin
  4. Load firestore: `admin.firestore()`

This appears to happen in the google-cloud package:

```
webpackMissingModule
$folder/node_modules/@google-cloud/firestore/build/src/index.js:24
./node_modules/@google-cloud/firestore/build/src/index.js
$folder/node_modules/@google-cloud/firestore/build/src/index.js:24
__webpack_require__
$folder/webpack/bootstrap 9583bf4009469bcec19d:678
fn
$folder/webpack/bootstrap 9583bf4009469bcec19d:88
get
$folder/node_modules/firebase-admin/lib/firebase-namespace.js:329
```

Specifically this line:

```
const libVersion = require(pkgUp.sync(__dirname)).version;
```
Does anyone have any insight?

schmidt-sebastian:
I suspect that Webpack removes the package.json here in a manner that we don't support. We have dealt with a couple issues related `pkgUp` lately, and I am inclined to fix this on our end by removing this dependency.

PR #321, opened 2018-08-14
Giving up on pkg-up

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/316

pkgUp seems to create more problems than it solves. This PR removes it and changes our source location to "dev/...", which allows us to find the package.json three levels up, whether we run from /build or from /dev.


JustinBeckwith:
This is kinda weird.  We don't have a `dev` folder in any of our other npm modules.  Why are we doing this?

ofrobots:
@JustinBeckwith 
> This is kinda weird. We don't have a dev folder in any of our other npm modules. 

Actually we do. https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs has a top-level `ts/` to ensure that pre-compile relative paths are the same as post-compile relative paths.

dev/src/v1beta1/firestore_client.js
```diff
@@ -18,9 +18,8 @@ const gapicConfig = require('./firestore_client_config');
 const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
-const pkgUp = require('pkg-up');
 
-const VERSION = require(pkgUp.sync(__dirname)).version;
+const VERSION = require('../../../package').version;
```
ofrobots: suggestion: I would recommend using the suffix `.json` even though it is not necessary.

schmidt-sebastian: Done

PR #345, opened 2018-09-11
Regenerate library with synth.py customizations

kinwa91:
firestore had a number of unique pattern that warrants a customized `synth.py`:

 - code are in `/dev`
 - init library like so: `const client = require('@google-cloud/firestore').v1()`
 - `.jsdoc.js` needs to pull source from `build/`
 - uses its own project (ok since cannot share a project with one using datastore) - running synthtool from HEAD (GoogleCloudPlatform/synthtool/pull/63)

system-test and samples-test can be verified in this branch (run-synth-system-tests):
https://circleci.com/gh/googleapis/workflows/nodejs-firestore/tree/run-synth-system-tests
(in-sync with this branch except enabling system/samples-test in .circleci configs)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,21 +149,24 @@ jobs:
           name: Decrypt credentials.
           command: |
             if ! [[ -z "${SYSTEM_TESTS_ENCRYPTION_KEY}" ]]; then
-              openssl aes-256-cbc -d -in .circleci/key.json.enc \
-                -out .circleci/key.json \
-                -k "${SYSTEM_TESTS_ENCRYPTION_KEY}"
+              for encrypted_key in .circleci/*.json.enc; do
+                openssl aes-256-cbc -d -in $encrypted_key \
+                  -out $(echo $encrypted_key | sed 's/\.enc//') \
+                  -k "${SYSTEM_TESTS_ENCRYPTION_KEY}"
+              done
             fi
       - run: *npm_install_and_link
       - run:
           name: Run system tests.
           command: npm run system-test
           environment:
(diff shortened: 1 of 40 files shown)


.circleci/config.yml
```diff
@@ -128,7 +128,7 @@ jobs:
           name: Run sample tests.
           command: npm run samples-test
           environment:
-            GCLOUD_PROJECT: long-door-651
+            GCLOUD_PROJECT: node-gcloud-ci
```
JustinBeckwith: that's weird - why is this different?

.nycrc
```diff
@@ -1,10 +1,9 @@
 {
   "report-dir": "./.coverage",
   "exclude": [
-    "build/src/*{/*,/**/*}.js",
-    "build/src/*/v*/*.js",
-    "build/protos/*.js",
-    "build/test/**/*.js"
+    "src/*{/*,/**/*}.js",
```
JustinBeckwith: Eugh.  We need to configure `nyc` to do "the right thing" here for TypeScript projects.   Please file an issue in synthtool to change the template to make this work for TypeScript and for JavaScript.

JustinBeckwith:
Also, thank you for doing this @kinwa91

.circleci/config.yml
kinwa91: Firestore uses a different project than the rest of the nodejs- libraries. Apparently it can't be used in the same project that uses Datastore?

JustinBeckwith: Oh yeah!  That's a good reason.

.nycrc
kinwa91: Done!

PR #361, opened 2018-09-25
 Adding DocumentGroup

schmidt-sebastian:
This adds the DocumentGroup API and exposes two new features:
- FieldMasks for document gets
- Snapshot Listeners for multiple documents

As part of this, I am also marking Firestore.getAll() as deprecated, but keeping Transaction.getAll(). Document Groups don't work well in Transactions, since there is no way to enforce that our users call `.get()`, which is required to obtain the pessimistic lock. I could be convinced to keep `.getAll()` everywhere, especially if we want to add it on Mobile. 

An unforeseen consequence is that QuerySnapshot.query is now not always present. We could rename it to `.source` and return a Query or a DocumentGroup or something.


dev/src/document-group.ts
```diff
@@ -0,0 +1,347 @@
...
+   *
+   * @example
+   * let docA = firestore.doc('coll/a');
+   * let docB = firestore.doc('coll/b');
+   *
+   * let documentGroup = firestore.documentGroup(docA, docB);
+   *
+   * let firestore = documentGroup.firestore;
+   */
+  get firestore() {
```
schmidt-sebastian: Note this is getter is only defined for its docstring. Otherwise I would have inlined it.

dev/src/document-group.ts
```diff
@@ -0,0 +1,348 @@
...
+
+
+    return responseStream;
+  }
+
+  /**
+   * Attaches a listener to retrieve real-time change events for the document
+   * in this document group.
+   *
+   * The results are ordered by document name and missing documents are
```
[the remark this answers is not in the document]
mikelehen: Decision: .doc() should return null if the document isn't in the QuerySnapshot (even though technically it might exist but they didn't include it in their .documents() list.

types/firestore.d.ts
```diff
@@ -122,9 +122,14 @@ declare namespace FirebaseFirestore {
      */
```
schmidt-sebastian: BTW, I am inclined to stall development on this until we have settled on the final API. To make this easier, I updated the `firestore.d.ts` file with a new version that matches my understanding of where we ended up in the standup (minus naming)

PR #362, opened 2018-09-25
fix: improve TypeScript types

JustinBeckwith:

- [ ] Tests and linter pass

### DIFF
--- a/dev/src/document-change.ts
+++ b/dev/src/document-change.ts
@@ -17,6 +17,7 @@
 'use strict';
 
 import * as is from 'is';
+import {QueryDocumentSnapshot} from './document';
 
 /**
  * A DocumentChange represents a change to the documents matching a query.
@@ -25,6 +26,10 @@ import * as is from 'is';
  * @class
  */
 export class DocumentChange {
+  _type: string;
+  _document: QueryDocumentSnapshot;
+  _oldIndex: number;
+  _newIndex: number;
   /**
    * @private
    * @hideconstructor
@@ -36,7 +41,9 @@ export class DocumentChange {
    * @param {number} newIndex - The index in the documents array after this
    * change.
    */
-  constructor(type, document, oldIndex, newIndex) {
+  constructor(
+      type: string, document: QueryDocumentSnapshot, oldIndex: number,
+      newIndex: number) {
     this._type = type;
     this._document = document;
     this._oldIndex = oldIndex;
--- a/dev/src/document-change.ts
+++ b/dev/src/document-change.ts
@@ -18,6 +18,7 @@
 
 import * as is from 'is';
 import {QueryDocumentSnapshot} from './document';
+export type DocumentChangeType = 'added' | 'removed' | 'modified';
 
 /**
  * A DocumentChange represents a change to the documents matching a query.
@@ -26,10 +27,10 @@ import {QueryDocumentSnapshot} from './document';
  * @class
  */
 export class DocumentChange {
-  _type: string;
-  _document: QueryDocumentSnapshot;
-  _oldIndex: number;
-  _newIndex: number;
+  private readonly _type: DocumentChangeType;
+  private readonly _document: QueryDocumentSnapshot;
+  private readonly _oldIndex: number;
+  private readonly _newIndex: number;
   /**
    * @private
    * @hideconstructor
@@ -42,7 +43,7 @@ export class DocumentChange {
    * change.
    */
   constructor(
-      type: string, document: QueryDocumentSnapshot, oldIndex: number,
+      type: DocumentChangeType, document: QueryDocumentSnapshot, oldIndex: number,
       newIndex: number) {
     this._type = type;
     this._document = document;
@@ -167,7 +168,7 @@ export class DocumentChange {
    * @param {*} other The value to compare against.
    * @return true if this `DocumentChange` is equal to the provided value.
    */
-  isEqual(other) {
+  isEqual(other): boolean {
     if (this === other) {
       return true;
     }
--- a/dev/src/document.ts
+++ b/dev/src/document.ts
@@ -23,8 +23,89 @@ import is from 'is';
 import {FieldPath} from './path';
 import {FieldTransform} from './field-value';
 import {Timestamp} from './timestamp';
-import {validatePkg} from './validate';
 import {isPlainObject} from './serializer';
+
+/**
+ * Returns a builder for DocumentSnapshot and QueryDocumentSnapshot instances.
+ * Invoke `.build()' to assemble the final snapshot.
+ *
+ * @private
+ * @class
+ */
+class DocumentSnapshotBuilder {
+  ref;
+  fieldsProto;
+  readTime;
+  createTime;
+  updateTime;
+  /**
+   * @private
+   * @hideconstructor
+   *
+   * @param {DocumentSnapshot=} snapshot An optional snapshot to base this
+   * builder on.
+   */
+  constructor(snapshot?) {
+    snapshot = snapshot || {};
+
+    /**
+     * The reference to the document.
+     *
+     * @type {DocumentReference}
+     */
+    this.ref = snapshot._ref;
+
+    /**
+     * The fields of the Firestore `Document` Protobuf backing this document.
+     *
+     * @type {object}
+     */
+    this.fieldsProto = snapshot._fieldsProto;
+
+    /**
+     * The time when this document was read.
+     *
+     * @type {Timestamp}
+     */
+    this.readTime = snapshot._readTime;
+
+    /**
+     * The time when this document was created.
+     *
+     * @type {Timestamp}
+     */
+    this.createTime = snapshot._createTime;
+
+    /**
+     * The time when this document was last updated.
+     *
+     * @type {Timestamp}
+     */
+    this.updateTime = snapshot._updateTime;
+  }
+
+  /**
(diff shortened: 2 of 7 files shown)


dev/src/document-change.ts
```diff
@@ -36,7 +41,9 @@ export class DocumentChange {
    * @param {number} newIndex - The index in the documents array after this
    * change.
    */
-  constructor(type, document, oldIndex, newIndex) {
+  constructor(
+      type: string, document: QueryDocumentSnapshot, oldIndex: number,
```
schmidt-sebastian: Can you just make these: `private readonly _type:string, private readonly _document ...`?

And we you want to be really pedantic, we could change the type of `type` to:
`DocumentChangeType = 'added' | 'removed' | 'modified'`

dev/src/document-change.ts
```diff
@@ -36,7 +41,9 @@ export class DocumentChange {
...
+      newIndex: number) {
```
schmidt-sebastian: Commenting on the methods below: Should we add the return type annotation?

dev/src/document.ts
```diff
@@ -23,9 +23,90 @@ import is from 'is';
...
+  createTime;
+  updateTime;
+  /**
+   * @private
+   * @hideconstructor
+   *
+   * @param {DocumentSnapshot=} snapshot An optional snapshot to base this
+   * builder on.
+   */
+  constructor(snapshot?) {
```
schmidt-sebastian: `snapshot?:DocumentSnapshot`

dev/src/document.ts
```diff
@@ -23,9 +23,90 @@ import is from 'is';
...
 
+/**
+ * Returns a builder for DocumentSnapshot and QueryDocumentSnapshot instances.
+ * Invoke `.build()' to assemble the final snapshot.
+ *
+ * @private
+ * @class
+ */
+class DocumentSnapshotBuilder {
+  ref;
```
schmidt-sebastian:
```
private ref: DocumentReference;
private fieldsProto: api.IMapValue; (?)
private readTime: Timestamp;
private createTime: Timestamp;
private updateTime: Timestamp;
```

dev/src/document.ts
```diff
@@ -23,9 +23,90 @@ import is from 'is';
...
+   * @param {DocumentSnapshot=} snapshot An optional snapshot to base this
+   * builder on.
+   */
+  constructor(snapshot?) {
+    snapshot = snapshot || {};
+
+    /**
+     * The reference to the document.
+     *
+     * @type {DocumentReference}
```
schmidt-sebastian: I would remove the JSDoc type annotation.

dev/src/document.ts
```diff
@@ -23,9 +23,90 @@ import is from 'is';
...
+
+  /**
+   * Builds the DocumentSnapshot.
+   *
+   * @private
+   * @returns {QueryDocumentSnapshot|DocumentSnapshot} Returns either a
+   * QueryDocumentSnapshot (if `fieldsProto` was provided) or a
+   * DocumentSnapshot.
+   */
+  build() {
```
schmidt-sebastian: `build(): QueryDocumentSnapshot|DocumentSnapshot`

dev/src/document.ts
```diff
@@ -23,9 +23,90 @@ import is from 'is';
...
+     * @type {Timestamp}
+     */
+    this.updateTime = snapshot._updateTime;
+  }
+
+  /**
+   * Builds the DocumentSnapshot.
+   *
+   * @private
+   * @returns {QueryDocumentSnapshot|DocumentSnapshot} Returns either a
```
schmidt-sebastian: In my conversion, I removed the JSDoc types from the comments for non public methods. It's still needed for public methods to generate the docs.

dev/src/document.ts
```diff
@@ -41,6 +122,13 @@ import {isPlainObject} from './serializer';
  * @class
  */
 export class DocumentSnapshot {
+  _ref;
+  _fieldsProto;
+  _serializer;
+  _validator;
+  _readTime;
+  _createTime;
+  _updateTime;
```
schmidt-sebastian: Same comment regarding the types and visibility as above.

dev/src/document.ts
```diff
@@ -626,7 +647,7 @@ export class DocumentMask {
    * @returns {DocumentMask}
    */
   static fromUpdateMap(data) {
-    let fieldPaths = [];
+    const fieldPaths: FieldPath[] = [];
```
schmidt-sebastian: Method signature: `fromUpdateMap(data:UpdateData) : DocumentMask`

dev/src/document.ts
```diff
@@ -663,12 +684,12 @@ export class DocumentMask {
    * @returns {DocumentMask}
    */
   static fromObject(data) {
-    let fieldPaths = [];
+    const fieldPaths: FieldPath[] = [];
```
schmidt-sebastian: Method signature: fromObject(data:DocumentData) : DocumentMask

dev/src/document.ts
```diff
@@ -645,7 +666,7 @@ export class DocumentMask {
    * @returns {DocumentMask}
    */
   static fromFieldMask(fieldMask) {
-    let fieldPaths = [];
+    const fieldPaths: FieldPath[] = [];
```
schmidt-sebastian: Method signature: fromFieldMask(fieldMask:Array<string|FieldPath>): DocumentMask

dev/src/document.ts
```diff
@@ -857,6 +878,9 @@ export class DocumentMask {
  * @class
  */
 export class DocumentTransform {
+  _ref;
+  _validator;
+  _transforms;
```
schmidt-sebastian: Same comment as above.

dev/src/document.ts
```diff
@@ -881,9 +905,9 @@ export class DocumentTransform {
    * @returns {firestore.DocumentTransform} The Document Transform.
    */
   static fromObject(ref, obj) {
-    let updateMap = new Map();
+    const updateMap = new Map();
```
schmidt-sebastian: Same type suggestions as for DocumentMask.

dev/src/timestamp.ts
```diff
@@ -41,6 +41,9 @@ const MS_TO_NANOS = 1000000;
  * @see https://github.com/google/protobuf/blob/master/src/google/protobuf/timestamp.proto
  */
 export class Timestamp {
+  _seconds: number;
```
schmidt-sebastian: Inline these into constructor and mark `private readonly`.

dev/src/timestamp.ts
```diff
@@ -41,6 +41,9 @@ const MS_TO_NANOS = 1000000;
...
+  _nanoseconds: number;
+
```
schmidt-sebastian: As in other files, I think we should add argument and return types.

dev/src/watch.ts
```diff
@@ -27,6 +27,11 @@ import {ResourcePath} from './path';
 import {requestTag} from './util';
 import {DocumentSnapshot} from './document';
 import {DocumentChange} from './document-change';
+import {PassThrough} from 'stream';
```
schmidt-sebastian: Thanks for tackling this file. I would like to also be much more pedantic here about the types, but I think that's beyond the scope of a simple TS conversion. If you'd like, I can enforce stricter types once this PR is merged.

dev/system-test/firestore.ts
```diff
@@ -44,7 +51,7 @@ describe('Firestore class', function() {
   it('has collection() method', function() {
     let ref = firestore.collection('col');
     assert.equal(ref.id, 'col');
-  });
+  });``
```
schmidt-sebastian: Remove

dev/system-test/firestore.ts
```diff
@@ -19,17 +19,24 @@
 import * as assert from 'power-assert';
 import is from 'is';
 
-import Firestore from '../src';
+const Firestore = require('../src');
 import {autoId} from '../src/util';
+import {DocumentSnapshot} from '../src/document';
```
schmidt-sebastian: Should we convert this file to use arrow functions for describe/it?

dev/src/document-change.ts
JustinBeckwith: I tend to only specify the type annotation when it can't be inferred.  In almost all of these cases, it can be.

dev/src/document.ts
JustinBeckwith: I'm not looking to get to complete typedness here (FWIW, this change is complex because `DocumentSnapshot` has a bunch of required fields and methods atm)

dev/src/document.ts
JustinBeckwith: Sadly we can't :(  Right now we rely on these for generating the reference documentation.

dev/src/document.ts
JustinBeckwith: I kind of want to keep them around for now.  We can always come back and clean this kind of stuff up.

dev/src/document.ts
JustinBeckwith: The plan for types (here and beyond) is to come back around and do that in another PR.  I like to do the bare minimum to get a file converted to TypeScript, then make it type-awesome in a followup :)

schmidt-sebastian:
I merged this into a new branch called `oldbranch` that is behind `master`. I will clean up the conflicts in the main repo.

PR #367, opened 2018-09-29
Convert test/collection.ts

schmidt-sebastian:
This PR fixes some compile errors using ts_node and converts test/collection.js to TS.


dev/system-test/firestore.ts
```diff
@@ -25,11 +26,21 @@ import {DocumentSnapshot} from '../src/document';
...
+  promise: Promise<{}>;
+
+  constructor() {
+    this.resolve = () => {
+      throw new Error('DeferredPromise.resolve has not been initialized');
+    };
+    this.reject = () => {
+      throw new Error('DeferredPromise.reject has not been initialized');
+    };
+    this.promise = Promise.resolve('DeferredPromise.promise has not been initialized');
```
schmidt-sebastian: Note that I wanted to make this Promise.reject(), but that creates those Node warnings if it is not handled (which it is not).

tslint.json
```diff
@@ -2,5 +2,8 @@
   "extends": "gts/tslint.json" ,
   "jsRules": {
     "no-bitwise": true
+  },
+  "rules": {
```
JustinBeckwith: huh, why do we need this?

PR #368, opened 2018-09-29
 Add List() API

schmidt-sebastian:
This adds support for the ListDocuments RPC, which allows us to enumerate all documents, including missing documents. We only return DocumentReference and don't fetch any data.


dev/src/reference.ts
```diff
@@ -1816,6 +1816,54 @@ export class CollectionReference extends Query {
...
+   *    let documentsWithData = 0;
+   *    let missingDocuments = 0;
+   *
+   *    for (let documentSnapshot of documentSnapshots) {
+   *       if (documentSnapshot.exists) {
+   *        documentsWithData++;
+   *       } else {
+   *        missingDocuments++;
+   *       }
+   *    }
```
mikelehen: Maybe console.log the values?

dev/test/collection.ts
```diff
@@ -135,6 +135,28 @@ describe('Collection interface', () => {
     });
   });
 
+  it('hast list() method', () => {
```
mikelehen: German?

types/firestore.d.ts
```diff
@@ -1035,6 +1035,18 @@ declare namespace FirebaseFirestore {
      */
     readonly path: string;
 
+    /**
+     * Retrieves the list of documents in this collection.
+     *
+     * The document references returned may include missing documents, which
+     * are documents that have not been explicitly created but contain
+     * subcollections. If you attempt to read a missing document, we will return
+     * a DocumentSnapshot whose `.exists` property is set to false.
```
mikelehen: To tighten the terminology slightly, I'd say:

The document references may include references to "missing documents", i.e. document locations that have no document present but which contain subcollections with documents.  Attempting to read such a document reference (e.g. via `.get()` or `.onSnapshot()`) will return a `DocumentSnapshot` whose `.exists` property is false.

dev/test/collection.ts
mikelehen: That is exactly what I thought of. 😀

mikelehen:
Still LGTM.

PR #373, opened 2018-10-02
Deprecate getCollections in favor of listCollections

schmidt-sebastian:
Deprecate getCollections in favor of listCollections

To prepare for `listDocuments()`.

### DIFF
--- a/dev/src/index.js
+++ b/dev/src/index.js
@@ -591,6 +591,27 @@ class Firestore {
    * with an array of CollectionReferences.
    *
    * @example
+   * firestore.listCollections().then(collections => {
+   *   for (let collection of collections) {
+   *     console.log(`Found collection with id: ${collection.id}`);
+   *   }
+   * });
+   */
+  listCollections() {
+    let rootDocument = new DocumentReference(this, this._referencePath);
+    return rootDocument.listCollections();
+  }
+
+  /**
+   * Fetches the root collections that are associated with this Firestore
+   * database.
+   *
+   * @deprecated Use `listCollections()`.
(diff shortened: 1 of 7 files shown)


dev/src/index.js
```diff
@@ -591,15 +591,35 @@ class Firestore {
...
+  /**
+   * Fetches the root collections that are associated with this Firestore
+   * database.
+   *
+   * @deprecated Use `listCollections()`.
+   *
+   * @returns {Promise.<Array.<CollectionReference>>} A Promise that resolves
+   * with an array of CollectionReferences.
+   *
+   * @example
```
mikelehen: FWIW, I think it'd be fair to omit example code for deprecated methods.  I don't care strongly though.

dev/src/index.js
```diff
@@ -591,15 +591,35 @@ class Firestore {
...
    *
    * @example
+   * firestore.listCollections().then(collections => {
+   *   for (let collection of collections) {
+   *     console.log(`Found collection with id: ${collection.id}`);
+   *   }
+   * });
+   */
+  listCollections() {
+    let rootDocument = new DocumentReference(this, this._referencePath);
```
JustinBeckwith: Use `const` instead of `let` if it isn't re-assigned.  The linter should be catching this 🤔

dev/src/index.js
schmidt-sebastian: Done

dev/src/index.js
schmidt-sebastian: Done

PR #375, opened 2018-10-03
Make sure we don't drop the google-cloud-resource-header

schmidt-sebastian:
The newly generated Gapic client drops the custom google-cloud-resource-prefix header. This fixes the generated code.


schmidt-sebastian:
Hm, I pushed a PR to fix the lint errors in system-test, since 4d507c8 added it to tslint.

PR #379, opened 2018-10-08
Convert Index and WriteBatch

schmidt-sebastian:
This converts the last bit of the client code to TS. There are still tests in JS.

I also enabled TSLint for the API test (test/typescript.ts), which uncovered a slew of problems.

### DIFF
--- a/dev/src/document.ts
+++ b/dev/src/document.ts
@@ -131,25 +131,27 @@ export class DocumentSnapshot {
   private _fieldsProto;
   private _serializer;
   private _validator;
-  private _readTime: Timestamp;
-  private _createTime: Timestamp;
-  private _updateTime: Timestamp;
+  private _readTime: Timestamp|undefined;
+  private _createTime: Timestamp|undefined;
+  private _updateTime: Timestamp|undefined;
+
   /**
    * @private
    * @hideconstructor
    *
-   * @param {firestore/DocumentReference} ref - The reference to the
-   * document.
-   * @param {object=} fieldsProto - The fields of the Firestore `Document`
-   * Protobuf backing this document (or undefined if the document does not
-   * exist).
-   * @param {Timestamp} readTime - The time when this snapshot was read.
-   * @param {Timestamp=} createTime - The time when the document was created
-   * (or undefined if the document does not exist).
-   * @param {Timestamp=} updateTime - The time when the document was last
-   * updated (or undefined if the document does not exist).
+   * @param ref The reference to the document.
+   * @param fieldsProto The fields of the Firestore `Document` Protobuf backing
+   * this document (or undefined if the document does not exist).
+   * @param readTime The time when this snapshot was read  (or undefined if
+   * the document exists only locally).
+   * @param createTime The time when the document was created (or undefined if
+   * the document does not exist).
+   * @param updateTime The time when the document was last updated (or undefined
+   * if the document does not exist).
    */
-  constructor(ref, fieldsProto, readTime?, createTime?, updateTime?) {
+  constructor(
+      ref: DocumentReference, fieldsProto?: api.IDocument, readTime?: Timestamp,
+      createTime?: Timestamp, updateTime?: Timestamp) {
     this._ref = ref;
     this._fieldsProto = fieldsProto;
     this._serializer = ref.firestore._serializer;
@@ -163,11 +165,11 @@ export class DocumentSnapshot {
    * Creates a DocumentSnapshot from an object.
    *
    * @private
-   * @param {firestore/DocumentReference} ref - The reference to the document.
-   * @param {Object} obj - The object to store in the DocumentSnapshot.
-   * @return {firestore.DocumentSnapshot} The created DocumentSnapshot.
+   * @param ref The reference to the document.
+   * @param obj The object to store in the DocumentSnapshot.
+   * @return The created DocumentSnapshot.
    */
-  static fromObject(ref, obj): DocumentSnapshot {
+  static fromObject(ref: DocumentReference, obj: {}): DocumentSnapshot {
     const serializer = ref.firestore._serializer;
     return new DocumentSnapshot(ref, serializer.encodeFields(obj));
   }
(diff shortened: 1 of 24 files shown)


dev/src/index.ts
```diff
@@ -1121,77 +1174,69 @@ follow these steps, YOUR APP MAY BREAK.`);
...
- * @param {string} options.allowDeletes At what level field deletes are
- * supported (acceptable values are 'none', 'root' or 'all').
- * @param {boolean} options.allowServerTimestamps Whether server timestamps
- * are supported.
- * @param {number=} depth The current depth of the traversal.
- * @param {number=} inArray Whether we are inside an array.
- * @returns {boolean} 'true' when the object is valid.
- * @throws {Error} when the object is invalid.
+ * @param val JavaScript value to validate.
+ * @param  options Validation options
```
mikelehen: nit: Maybe grep the code for "@param  " because I know there's at least a couple of these.

dev/src/index.ts
```diff
@@ -1290,15 +1330,7 @@ function validateDocumentData(obj, options) {
...
-/**
- * Sets the log function for all active Firestore instances.
- *
- * @method Firestore.setLogFunction
- * @param {Firestore~logFunction} logger - A log function that takes a single
- * string.
- */
-Firestore.setLogFunction = setLogFunction;
-
+// tslint:disable-next-line:no-default-export
```
mikelehen: It looks like you ended up adding this below...

dev/src/types.ts
```diff
@@ -32,10 +33,57 @@ export type ApiMapValue = {
...
+   * `Timestamp` instead.
+   *
+   * NOTE: in the future `timestampsInSnapshots: true` will become the
+   * default and this option will be removed so you should change your code to
+   * use `Timestamp` now and opt-in to this new behavior as soon as you can.
+   */
+  timestampsInSnapshots?: boolean;
+
+  // tslint:disable-next-line:no-any
+  [key: string]: any;  // Accept other properties, such as GRPC settings.
```
mikelehen: By the way, what version of TypeScript are we using?  TypeScript 3.0 introduces the type-safe type `unknown`, which may be more suitable than `any` for some of these things...

dev/src/write-batch.ts
```diff
@@ -76,21 +78,36 @@ export class WriteResult {
...
     return (
         this === other ||
         (is.instanceof(other, WriteResult) &&
          this._writeTime.isEqual(other._writeTime)));
   }
 }
 
+
+/** Helper type to manage the list of writes in a WriteBatch. */
+// TODO: Replace with api.IWrite
```
mikelehen: Who should do this? When? Why is it not being done now?

dev/src/write-batch.ts
```diff
@@ -565,11 +583,11 @@ export class WriteBatch {
  * Validates that the update data does not contain any ambiguous field
  * definitions (such as 'a.b' and 'a').
  *
- * @param {Map.<FieldPath, *>} data - An update map with field/value pairs.
- * @returns {boolean} 'true' if the input is a valid update map.
+ * @param data - An update map with field/value pairs.
```
mikelehen: Is the `-` supposed to be there?

tsconfig.json
```diff
@@ -7,17 +7,12 @@
     "noImplicitAny": false, // TODO: Remove when all source are migrated to TS
```
mikelehen: Are we here yet? Or no?

dev/src/document.ts
```diff
@@ -163,7 +167,7 @@ export class DocumentSnapshot {
    * @param {Object} obj - The object to store in the DocumentSnapshot.
```
mikelehen: Are these type annotations still intentional?

schmidt-sebastian: I only want to keep them for external methods, as they are used in JSDoc. Removed from here.

FYI: document.ts still has room for a bunch of TS-related improvements.

dev/src/index.ts
```diff
@@ -223,46 +281,10 @@ class Firestore {
...
-    this._clientPool = null;
-
-    /**
-     * Whether the initialization settings can still be changed by invoking
-     * `settings()`.
-     * @private
-     * @type {boolean}
-     */
-    this._settingsFrozen = false;
+    } as AnyDuringMigration);
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, that's the goal.

dev/src/index.ts
```diff
@@ -291,7 +313,7 @@ class Firestore {
    *
    * @param {object} settings The settings to use for all Firestore operations.
    */
-  settings(settings) {
+  settings(settings: Settings) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes to fixing this instance now, and turning this check on later.

dev/src/index.ts
```diff
@@ -657,16 +693,18 @@ class Firestore {
    * @returns {Promise<Array.<DocumentSnapshot>>} A Promise that contains an array with
    * the resulting documents.
    */
-  getAll_(docRefs, requestTag, transactionId) {
+  getAll_(
```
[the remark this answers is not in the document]
schmidt-sebastian: Haha. This follows the following convention:

_ + name : Method is actually private
name + _: Method is "package-protected"

I am planning on eventually getting rid of this convention :)

dev/src/index.ts
schmidt-sebastian: Fixed 8 of them.

dev/src/index.ts
schmidt-sebastian: For some reason, the JSDoc comment itself triggers this warning, as well as the actual line that defines it.  I suspect it has to do with

```
@module {Firestore} @google-cloud/firestore
```

dev/src/types.ts
schmidt-sebastian: We are using TS 3 as of last week. I filed a bug for `unknown` (b/117464450).

dev/src/write-batch.ts
schmidt-sebastian: I'll do this shortly. Wanted to keep this PR nice and concise.

dev/src/write-batch.ts
schmidt-sebastian: Sometimes. Sometimes not :/ Removed here.

tsconfig.json
schmidt-sebastian: Soon. I think we might have to migrate the tests.

PR #384, opened 2018-10-10
Removing last bits of asserts

schmidt-sebastian:
This removes all remaining asserts from the unit tests and replaces them with chai.

### DIFF
--- a/dev/src/backoff.ts
+++ b/dev/src/backoff.ts
@@ -51,7 +51,7 @@ const DEFAULT_JITTER_FACTOR = 1.0;
 /*!
  * The timeout handler used by `ExponentialBackoff`.
  */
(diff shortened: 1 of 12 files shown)


dev/src/backoff.ts
```diff
@@ -62,7 +62,8 @@ let delayExecution: typeof setTimeout = setTimeout;
  * @private
  * @param {function} handler A handler than matches the API of `setTimeout()`.
  */
-export function setTimeoutHandler(handler: typeof setTimeout): void {
+export function setTimeoutHandler(
```
schmidt-sebastian: The usage in test/watch.ts suddenly didn't compile anymore since `setImmediate` is not of type `setTimeout`.

PR #388, opened 2018-10-11
Using Strings instead of Enums in Protobuf Types

schmidt-sebastian:
I spent some more time digging through the protobufjs code to figure out how we can actually generate the .d.ts file for the Proto options that we are using in Veneer. This PR is the result of this, and it will unblock me from adding strict types to watch.ts (as we use Enum values in the Protocol).

As part of this, I also instructed `protobufjs` to not generate the serialization code, which reduces the amount of dead code considerably.


schmidt-sebastian:
PR that uses these types: https://github.com/googleapis/nodejs-firestore/pull/389

PR #391, opened 2018-10-12
Grepping dashes

schmidt-sebastian:
RegExp cleanup for dashes in comments.


dev/src/reference.ts
```diff
@@ -388,13 +386,13 @@ export class DocumentReference {
...
-   * containing the fields and values with which to update the document
-   * or the path of the first field to update.
+   * @param {UpdateData|string|FieldPath} dataOrField An object containing the
+   * fields and values with which to update the document or the path of the
+   * first field to update.
    * @param {
-   * ...(*|string|FieldPath|Precondition)} preconditionOrValues -
-   * An alternating list of field paths and values to update or a Precondition
-   * to restrict this update.
+   * ...(*|string|FieldPath|Precondition)} preconditionOrValues An alternating
```
mikelehen: Wow, your grep was even better than I imagined. 😀

dev/src/write-batch.ts
```diff
@@ -232,10 +232,10 @@ export class WriteBatch {
    * [SetOptions]{@link SetOptions}., the provided data can be merged
    * into the existing document.
    *
-   * @param {DocumentReference} documentRef - A reference to the
-   * document to be set.
-   * @param {DocumentData} data - The object to serialize as the document.
-   * @param {SetOptions=} options - An object to configure the set behavior.
+   * @param {DocumentReference} documentRef A reference to thedocument to be
```
mikelehen: missing space

dev/src/reference.ts
schmidt-sebastian: Clang format did this for me (in most places).

PR #405, opened 2018-10-19
docs: convert samples to async/await

muraliQlogic:

- [ ] Tests and linter pass

### DIFF
--- a/samples/quickstart.js
+++ b/samples/quickstart.js
@@ -18,7 +18,6 @@
 // [START firestore_quickstart]
 const Firestore = require('@google-cloud/firestore');
 async function main() {
-  try{
     const firestore = new Firestore({
       projectId: 'YOUR_PROJECT_ID',
       keyFilename: '/path/to/keyfile.json',
@@ -42,10 +41,7 @@ async function main() {
     
     // Delete the document.
     await document.delete();
-        
-  } catch (error){
-    console.error('Something went wrong:', error);
-  }
+  
 };
 
 main().catch(console.error);
--- a/samples/quickstart.js
+++ b/samples/quickstart.js
@@ -17,36 +17,34 @@
 
 // [START firestore_quickstart]
 const Firestore = require('@google-cloud/firestore');
-
-const firestore = new Firestore({
-  projectId: 'YOUR_PROJECT_ID',
-  keyFilename: '/path/to/keyfile.json',
-});
-
-const document = firestore.doc('posts/intro-to-firestore');
-
-// Enter new data into the document.
-document.set({
-  title: 'Welcome to Firestore',
-  body: 'Hello World',
-}).then(() => {
-  // Document created successfully.
-});
-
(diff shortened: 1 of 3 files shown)


samples/quickstart.js
```diff
@@ -17,36 +17,34 @@
...
-  // Document read successfully.
-});
-
-// Delete the document.
-document.delete().then(() => {
-  // Document deleted successfully.
-});
-// [END firestore_quickstart]
\ No newline at end of file
+(async () => {
```
JustinBeckwith: Instead of the closure here, can you create a `main` method?

```js
async function main() {
  ...
}

main.catch(console.error);
```

By doing it this way, you can totally drop the try/catch as well :)

muraliQlogic: Changes committed as suggested.

JustinBeckwith:
Looks like this broke the sample tests :/

muraliQlogic:
@JustinBeckwith : What do you suggest to fix the broken sample tests ? Should we modify the sample or the sample tests?

muraliQlogic:
@JustinBeckwith : Changes committed for sample tests .

samples/quickstart.js
```diff
@@ -17,36 +17,33 @@
...
 const Firestore = require('@google-cloud/firestore');
-
-const firestore = new Firestore({
-  projectId: 'YOUR_PROJECT_ID',
-  keyFilename: '/path/to/keyfile.json',
-});
-
-const document = firestore.doc('posts/intro-to-firestore');
-
-// Enter new data into the document.
```
JustinBeckwith: Let's keep the comments that were describing what was happening

samples/system-test/quickstart.test.js
```diff
@@ -16,30 +16,24 @@
...
-    '@google-cloud/firestore': FirestoreMock,
-  });
-
-  t.is(docMock.set.callCount, 1);
-  t.is(docMock.update.callCount, 1);
-  t.is(docMock.get.callCount, 1);
-  t.is(docMock.delete.callCount, 1);
-});
+
+describe('should make some API calls',function(){
```
JustinBeckwith: Lets use arrow functions instead of `function(){`

samples/system-test/quickstart.test.js
```diff
@@ -16,30 +16,24 @@
...
+ 
+  it('firestore_inspect_string', async function() {
```
JustinBeckwith: same arrow function feedback

samples/quickstart.js
muraliQlogic: Done

samples/system-test/quickstart.test.js
muraliQlogic: Done.

samples/system-test/quickstart.test.js
muraliQlogic: Done

PR #414, opened 2018-10-23
Add Firebase Version to header

schmidt-sebastian:
This PR allows the Admin SDK to pass its version to our SDK. We then add it as a header value to the existing x-goog-api-client header.

This looks like this:

x-goog-api-client: gl-node/v8.11.3 grpc/1.15.1 gax/0.20.0 gapic/0.18.0 gccl/0.18.0 fire/6.1.0


dev/src/index.ts
```diff
@@ -304,7 +299,16 @@ export class Firestore {
...
 
 
-    this.validateAndApplySettings(settings!);
+    const libraryHeader =  {
+      libName: 'gccl',
+      libVersion,
+    };
+
+    if (settings && settings.firebaseVersion) {
+      libraryHeader.libVersion += ' fire/' + settings.firebaseVersion;
```
Feiyang1: It seems hacky to add another header this way. Is it possible to make an array of headers ?

dev/src/index.ts
```diff
@@ -279,11 +279,6 @@ export class Firestore {
    * use `Timestamp` now and opt-in to this new behavior as soon as you can.
    */
   constructor(settings?: Settings) {
```
Feiyang1: can you please update the typing information for Settings?

dev/src/index.ts
schmidt-sebastian: I don't disagree with you, but I don't think this it is possible without changing the generated code. Currently, the generated code only exposes libName and libVersion.

@JustinBeckwith What are your thoughts on this?

dev/src/index.ts
schmidt-sebastian: I don't want to expose this setting in the public API. Hence, it is missing from both the `.d.ts` file and the JSDoc documentation. Note that they way the settings are defined allows you to pass any arbitrary key-value pair.

schmidt-sebastian:
Ping :)

alexander-fenster:
It looks the code will add `fire/x.x.x`, not `firebase/x.x.x` as the description states. LGTM if that's expected. If you can create an issue for [googleapis/gapic-generator](https://github.com/googleapis/gapic-generator) to have better support for multiple library-version pairs, it would be great.

schmidt-sebastian:
> It looks the code will add `fire/x.x.x`, not `firebase/x.x.x` as the description states. LGTM if that's expected. If you can create an issue for [googleapis/gapic-generator](https://github.com/googleapis/gapic-generator) to have better support for multiple library-version pairs, it would be great.

We updated the code to use `fire/`. I changed the description to match this.

Issue: https://github.com/googleapis/gapic-generator/issues/2400

PR #427, opened 2018-10-31
Improve types for document.ts and convert.ts

schmidt-sebastian:


dev/src/convert.ts
```diff
@@ -182,9 +185,9 @@ function convertValue(fieldValue) {
       };
     case 'arrayValue': {
       const arrayValue: Array<{}> = [];
-      if (is.array(fieldValue.arrayValue.values)) {
-        for (const value of fieldValue.arrayValue.values) {
-          arrayValue.push(convertValue(value));
+      if (is.array(fieldValue.arrayValue!.values)) {
```
mikelehen: Would `Array.isArray()` be better (might get rid of one `!` from the next line)?

dev/src/document.ts
```diff
@@ -23,92 +23,56 @@ import {google} from '../protos/firestore_proto_api';
...
+        (this.fieldsProto !== undefined) === (this.updateTime !== undefined),
         'Update time should be set iff document exists.');
     return this.fieldsProto ?
         new QueryDocumentSnapshot(
-            this.ref, this.fieldsProto, this.readTime, this.createTime,
-            this.updateTime) :
-        new DocumentSnapshot(this.ref, undefined, this.readTime);
+            this.ref!, this.fieldsProto!, this.readTime!, this.createTime!,
+            this.updateTime!) :
+        new DocumentSnapshot(this.ref!, undefined, this.readTime!);
```
mikelehen: Just FYI- Using `!` in the process of making things more typed seems fine, but if this was new code, I'd ask if there's a way to rework this to avoid the need for `!` (either make the type non-undefined-able or add if-guards that guarantee it's defined).  I think ultimately we should be removing`!` and `any` usages over time.

dev/src/convert.ts
schmidt-sebastian: That works, but it doesn't get rid of the `!`. I changed it either way.

dev/src/document.ts
schmidt-sebastian: But it looks so much more important with all those `!!!`. I'll see if I can get rid of some of them soonish.

PR #432, opened 2018-11-02
Update Conformance Tests

schmidt-sebastian:
This pulls in the latest version of the Conformance tests and adds support for ArrayUnion and ArrayRemove.

This PR also uses the Protobuf library to do some of the "formatting" changes for us. This is needed since the test asserts distinguish between { mapValue: { fields: [] }} and { mapValue: { }}, whereas the Protobuf runner didn't.


dev/conformance/runner.ts
```diff
@@ -177,80 +201,16 @@ const convertInput = {
...
-    if (deepCopy.startAt) {
-      deepCopy.startAt = convertProto.position(deepCopy.startAt);
-    }
-    if (deepCopy.endAt) {
-      deepCopy.endAt = convertProto.position(deepCopy.endAt);
-    }
-    return deepCopy;
-  },
-  listenRequest: listenRequest => {
+  listenResponse: listenRequest => {
```
schmidt-sebastian: Note that I kept this because in the default format, running this through the Protobuf library replaced the Enum String constants with their integer equivalent. That caused all sorts of issues.

dev/conformance/runner.ts
```diff
@@ -16,7 +16,8 @@
 
 import {expect} from 'chai';
 const duplexify = require('duplexify');
-import googleProtoFiles from 'google-proto-files';
+const googleProtoFiles = require('google-proto-files');
```
schmidt-sebastian: I needed to change this to run the tests locally. To make sure that we catch this type of issue automatically, I added the conformance tests back to the CI.

PR #442, opened 2018-11-09
Better Types

schmidt-sebastian:
Improve the types in preparation for .getAll().

I also had to remove an invocation for JSON.stringify in the Watch test, as it started complaining about circular references.


dev/src/write-batch.ts
```diff
@@ -519,44 +521,48 @@ export class WriteBatch {
...
-            ++offset;
+    return this._firestore
+        .request<api.CommitResponse>(
+            'commit', request, tag, /* allowRetries= */ false)
+        .then(resp => {
+          const writeResults: WriteResult[] = [];
+
+          if (request.writes!.length > 0) {
+            assert(
+                Array.isArray(resp.writeResults) &&
```
schmidt-sebastian: "resp.writeResults instanceof Array" to "Array.isArray(resp.writeResults)" is the only change in this block

dev/src/index.ts
```diff
@@ -226,7 +227,7 @@ export class Firestore {
    * The serializer to use for the Protobuf transformation.
    * @private
    */
-  private _serializer: Serializer|null = null;
+  _serializer: Serializer|null = null;
```
gsoltis:
```suggestion
  _serializer: Serializer | null = null;
```

dev/src/serializer.ts
```diff
@@ -43,7 +44,7 @@ export class Serializer {
   private createReference: (path: string) => AnyDuringMigration;
 
   constructor(
-      firestore: AnyDuringMigration,
+      private readonly firestore: Firestore,
```
gsoltis: This looks like it's adding a field. Is that necessary? Do you actually reference it anywhere?

dev/src/transaction.ts
```diff
@@ -333,10 +338,12 @@ export class Transaction {
...
-    return this._writeBatch.commit_({
-      transactionId: this._transactionId,
-      requestTag: this._requestTag,
-    });
+    return this._writeBatch
+        .commit_({
+          transactionId: this._transactionId,
+          requestTag: this._requestTag,
+        })
+        .then(() => {});
```
gsoltis:
```suggestion
        .then();
```

dev/src/index.ts
schmidt-sebastian: I'll try this but I bet clang format is going to change this back.

dev/src/serializer.ts
schmidt-sebastian: Ahhhh. That's probably why the JSON.stringify() failed. Thanks for catching this.

dev/src/transaction.ts
schmidt-sebastian: This doesn't compile for me and seems to forward the existing return value.

dev/src/index.ts
schmidt-sebastian: (Yeah, it's back to what it was before)

dev/src/transaction.ts
gsoltis: Oh, huh. It works with `.next()`, but maybe we have a different type setup for that vs `.then()`.

Issue #42, opened 2017-10-27
FR: allow DocumentReference.select similar to Query.select


goldensunliu:
Hey guys,
    I have been snooping around the source code and documentation. It seems like the only way to do apply a field mask to the result and returns only the specified subset of fields is through the Query.select method. 
https://cloud.google.com/nodejs/docs/reference/firestore/0.8.x/Query#select
https://github.com/googleapis/nodejs-firestore/blob/master/src/reference.js#L1080

Is there any plan or ways to enable the same feature when getting a single document via DocumentReference? The code seems plenty of references to https://cloud.google.com/firestore/docs/reference/rest/v1beta1/DocumentMask which is used to restrict a get or update operation on a document to a subset of its fields, but I didn't have any luck tracking down usage of this in DocumentReference.get

#### Relevant Code:
```javascript
// based on the select example at https://cloud.google.com/nodejs/docs/reference/firestore/0.8.x/Query#select
let collectionRef = firestore.collection('col');
let documentRef = collectionRef.doc('doc');

documentRef.set({x:10, y:5}).then(() => {
  return collectionRef.where('x', '>', 5).select('y').get();
}).then((res) => {
  console.log(`y is ${res.docs[0].get('y')}.`);
});
// would like to be able to also do
documentRef.select('y').get().then((snapShot) => {
  console.log(`snapShot only has one field ${snapShot}`);
  // Expected output: {y: 5}
});

**Comment by goldensunliu** [comment]
I have also filed this against https://github.com/firebase/firebase-js-sdk/issues/272 since I am not sure where the source of development on the firestore part is housed.

**Comment by goldensunliu** [comment]
https://github.com/firebase/firebase-js-sdk/issues/212 answers this

**Comment by IchordeDionysos** [comment]
@goldensunliu Are you sure that that the issue in the Firebase JS SDK really answers this?
As I understood the answer from @wilhuff, is that this won't appear in the JS CLIENT SDK as you are having offline capabilities there and so on.

I don't see a problem in the ADMIN SDK to add such a method to the DocumentReference. As far as I know, there is no caching in the ADMIN SDK, so it should be no problem to allow a snippet/method like this:

```js
admin.firestore().document("path/to/a/document").select("y").get().then(snap => {
    console.log(`snapShot only has one field ${snap.data()}`);
    // Expected output: {y: 5}
});
```

Or am I wrong?

wilhuff:
[...]

PR #443, opened 2018-11-09
Adding FieldMask support to GetAll()

schmidt-sebastian:
This adds a `ReadOptions` argument to getAll() to allow users to specify a field mask.

I tried to keep the API useable and declared the first argument of the varargs method so that it is known the DocumentReferences have to be passed first.

Fixes: https://github.com/googleapis/nodejs-firestore/issues/42

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -37,9 +37,9 @@ import {DocumentReference} from './reference';
 import {isPlainObject, Serializer} from './serializer';
 import {Timestamp} from './timestamp';
 import {Transaction} from './transaction';
-import {DocumentData, GapicClient, Settings, ValidationOptions} from './types';
+import {DocumentData, GapicClient, ReadOptions, Settings, ValidationOptions} from './types';
 import {AnyDuringMigration, AnyJs} from './types';
-import {requestTag} from './util';
+import {parseGetAllArguments, requestTag} from './util';
 import {customObjectError, Validator} from './validate';
 import {WriteBatch, WriteResult} from './write-batch';
 import {validateUpdateMap} from './write-batch';
@@ -161,6 +161,18 @@ const MAX_DEPTH = 20;
  * It is an error to pass a SetOptions object to a set() call that is missing a
  * value for any of the fields specified here.
  * @typedef {Object} SetOptions
+ */
+
+/**
+ * An options object that can be used to configure the behavior of
+ * [getAll()]{@link Firestore#getAll} calls. By providing a `fieldMask`, these
+ * calls can be configured to only return a subset of fields.
+ *
+ * @property {Array<(string|FieldPath)>} fieldMask Specifies the set of fields
+ * to return and reduces the amount of data transmitted by the backend.
+ * Adding a field mask does not filter results. Documents do not need to
+ * contain values for all the fields in the mask to be part of the result set.
+ * @typedef {Object} ReadOptions
  */
 
 /**
@@ -296,6 +308,7 @@ export class Firestore {
       QueryValue: validateFieldValue,
       ResourcePath: ResourcePath.validateResourcePath,
       SetOptions: validateSetOptions,
+      ReadOptions: validateReadOptions,
       UpdateMap: validateUpdateMap,
       UpdatePrecondition: precondition =>
           validatePrecondition(precondition, /* allowExists= */ false),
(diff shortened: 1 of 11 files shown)


dev/src/index.ts
```diff
@@ -1345,6 +1364,39 @@ function validateDocumentData(
...
 
+/**
+ * Validates the use of 'options' as ReadOptions and enforces that 'fieldMask'
+ * is an array of strings or field paths.
+ *
+ * @private
+ * @param options.fieldMask - The subset of fields to return from a read
+ * operation.
+ */
+export function validateReadOptions(options: ReadOptions): boolean {
```
gsoltis: Is there a reason this function returns a boolean? It looks like it either returns `true` or throws an error. I think either making it void and have it throw on error, or having it return `Error|null` and have the caller throw would be more consistent.

dev/src/transaction.ts
```diff
@@ -117,7 +118,7 @@ export class Transaction {
 
     if (refOrQuery instanceof DocumentReference) {
       return this._firestore
-          .getAll_([refOrQuery], this._requestTag, this._transactionId)
+          .getAll_([refOrQuery], null, this._requestTag, this._transactionId)
```
gsoltis: Can you label this `null`? In this context it's a lot like a `boolean` with no clear meaning.

Or add something like `const fieldMask = null` and then pass that?

dev/src/util.ts
```diff
@@ -43,3 +48,49 @@ export function autoId(): string {
...
+ * an optional ReadOptions object.
+ */
+export function parseGetAllArguments(
+    validator: AnyDuringMigration,
+    documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
+    {documents: DocumentReference[], fieldMask: FieldPath[]|null} {
+  let documents: DocumentReference[];
+  let readOptions: ReadOptions|undefined = undefined;
+
+  const usesVarags = !Array.isArray(documentRefsOrReadOptions[0]);
```
gsoltis: This looks wrong to me. The function signature says that `documentRefsOrReadOptions` is of type `Array<DocumentReference|ReadOptions>`, so I would expect that the first element must be of type `DocumentReference|ReadOptions` and could never be an array.

dev/src/util.ts
```diff
@@ -43,3 +48,49 @@ export function autoId(): string {
...
+    if (documentRefsOrReadOptions.length > 0 &&
+        isPlainObject(
+            documentRefsOrReadOptions[documentRefsOrReadOptions.length - 1])) {
+      readOptions = documentRefsOrReadOptions.pop() as ReadOptions;
+      documents = documentRefsOrReadOptions as DocumentReference[];
+    } else {
+      documents = documentRefsOrReadOptions as DocumentReference[];
+    }
+  } else {
+    // Support an array of document references as the first argument for
```
gsoltis: Is there a test for this legacy support? I would not expect a call that hits this condition to typecheck.

dev/test/index.ts
```diff
@@ -883,7 +879,8 @@ describe('getAll() method', () => {
     const overrides = {batchGetDocuments: () => stream(found('documentId'))};
 
     return createInstance(overrides).then(firestore => {
-      return firestore.getAll(firestore.doc('collectionId/documentId'))
+      return (firestore as InvalidApiUsage)
+          .getAll([firestore.doc('collectionId/documentId')])
```
gsoltis: If this is expected to work, we shouldn't have to cast to `InvalidApiUsage`, right? That would imply that we might be breaking someone's typescript project?

dev/src/util.ts
gsoltis: Found the test, and it looks like it wouldn't typecheck without a cast.

dev/src/util.ts
schmidt-sebastian: The non-TypeScript versions of this client supported a call to getAll with an array of DocumentReferences. I did not want to break this for no reason, especially since we have supported this through the prior TypeScript versions. There is already a comment on line 80, but I moved it up and made it much more explicit.

dev/src/util.ts
schmidt-sebastian: Yeah, it's meant to support users that still use JS directly *shudder* :)

dev/test/index.ts
schmidt-sebastian: This was never supported for TypeScript users, only for JavaScript users.

This change does however break the no-argument call to `getAll()`. Before, it was allowed to call `getAll()` with no references, now one needs to be provided. This is a compile-time error and we are bumping the minor version.

We should consider removing the array support when we go GA: https://b.corp.google.com/issues/119353405

PR #450, opened 2018-11-11
Initialize Serializer directly

schmidt-sebastian:
This removes some of the indirection around the Serializer class. It uses slightly more memory, but I think it's cleaner.


dev/src/index.ts
```diff
@@ -204,17 +204,17 @@ export class Firestore {
...
-   * `settings()`.
+   * The configuration options for the GAPIC client.
    * @private
    */
-  private _settingsFrozen = false;
+  _settings: Settings = {};
 
   /**
-   * The configuration options for the GAPIC client.
+   * Whether the initialization settings can still be changed by invoking
```
var-const: Nit: now that `initializationSettings` have been renamed to just `settings`, should the word "initialization" be removed here as well?

schmidt-sebastian: I kind of prefer the explicitness here. Leaving as is.

Issue #436, opened 2018-11-07
Document with a number type property fails, if number is zero


timojuonoja:
#### Environment details

  - OS: Linux
  - Node.js version: 8.10.0
  - npm version: 5.6.0
  - `@google-cloud/firestore` version: 0.18.0

#### Steps to reproduce

  1. `collection('col1').doc('doc1').set({ number: 0 })`
  2. Fails with `Error: Invalid use of type "undefined" as a Firestore argument.`

schmidt-sebastian:
Do you mind sharing more of you code? I am not able to reproduce this:

```
    const ref = firestore.doc('col/doc');
    return ref.set({ number: 0 });
```

Logs:
```
Firestore (0.18.0) 2018-11-09T00:55:07.001Z 9ZfUT [Firestore.request]: Sending request: {"database":"projects/.../databases/(default)","writes":[{"update":{"name":"projects/.../databases/(default)/documents/col/doc","fields":{"number":{"integerValue":0}}}}]}
Firestore (0.18.0) 2018-11-09T00:55:07.613Z 9ZfUT [Firestore.request]: Received response: {"writeResults":[{"transformResults":[],"updateTime":{"seconds":"1541724798","nanos":668307000}}],"commitTime":{"seconds":"1541724906","nanos":682288000}}
```

timojuonoja:
Whoops, maybe the problem is not then number zero. I was just assuming that because I could not find any other reason for `"undefined"`.

Here is test code:
```
import faces from 'facesjs'

export const test = https.onCall(async () => {
  try {
    const data = { face: faces.generate() }
    console.log("Data", JSON.stringify(data))
    await db.collection('test').doc('doc').set(data)
    return {}
  } catch (err) {
    console.log(err)
    throw new https.HttpsError('internal')
  }
})
```

And here is the result as attachment:
![2018-11-09 08 37 42](https://user-images.githubusercontent.com/7699788/48247266-56f51980-e3fb-11e8-8133-98dc4d3ffdb8.png)

timojuonoja:
But I realized that it is even better to stringify the content for storing that onto firestore. So my code works fine now, but I just wanted you to know about the problem.

timojuonoja:
@schmidt-sebastian Ok. I'm using firestore for the first time, so therefore I have still lot to learn. Thanks for the reply.

PR #451, opened 2018-11-12
Slightly better error messages

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/436 and https://b.corp.google.com/issues/78932093

This PR:
- adds the field path to the error message if a field-level validation fails
- uses consistent quotes in all errors (changed single quotes to double)
- enforces that all errors are verified as they are raised by removing the RegExp matching and matching by exact value instead

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -295,8 +295,9 @@ export class Firestore {
    */
   constructor(settings?: Settings) {
     this._validator = new Validator({
-      ArrayElement: (name, value) =>
-          validateFieldValue(name, value, /* depth */ 0, /*inArray=*/true),
+      ArrayElement: (name, value) => validateFieldValue(
+          name, value, /* path */ undefined, /*level=*/0,
+          /*inArray=*/true),
       DeletePrecondition: precondition =>
           validatePrecondition(precondition, /* allowExists= */ true),
       Document: validateDocumentData,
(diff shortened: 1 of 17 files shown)


dev/src/index.ts
```diff
@@ -289,8 +289,9 @@ export class Firestore {
    */
   constructor(settings?: Settings) {
     this._validator = new Validator({
-      ArrayElement: (name, value) =>
-          validateFieldValue(name, value, /* depth */ 0, /*inArray=*/true),
+      ArrayElement: (name, value) => validateFieldValue(
+          name, value, /* path */ undefined, /*level=*/0,
```
mikelehen: super nit: path=

dev/src/index.ts
```diff
@@ -1261,51 +1262,67 @@ follow these steps, YOUR APP MAY BREAK.`);
  *
  * @private
  * @param val JavaScript value to validate.
+ * @param path The field path to validate.
  * @param options Validation options
- * @param depth The current depth of the traversal.
+ * @param level The current depth of the traversal. This is used to decided
```
mikelehen: decided=>decide

dev/src/path.ts
```diff
@@ -477,20 +481,29 @@ export class FieldPath extends Path<FieldPath> {
    */
   static validateFieldPath(fieldPath: string|FieldPath) {
     if (!(fieldPath instanceof FieldPath)) {
-      if (!is.string(fieldPath)) {
+      if (fieldPath === undefined) {
```
mikelehen: FWIW this seems to contradict the type signature of the method, which is weird / confusing.  Perhaps we should be handling undefined higher up, or we should fix the signature to indicate that we don't actually know what fieldPath is?

dev/src/index.ts
schmidt-sebastian: Fixed

dev/src/index.ts
schmidt-sebastian: Fixed

PR #497, opened 2018-12-27
Cleanup: Move parseGetAllArguments

schmidt-sebastian:

This PR moves the `getAll()` argument validation to transaction.ts, in order to avoid a new cyclic dependency. It also updates the validation to no longer use `is` for its type checking and to use the new `AllowOptional` interface.


schmidt-sebastian:

PR #499, opened 2018-12-28
Cleanup: Remove validator

schmidt-sebastian:

This removes the Validator class and replaces the remaining invocations with free-standing methods.


schmidt-sebastian:

PR #500, opened 2018-12-28
Cleanup: Remove 'any'

schmidt-sebastian:

This PR removes all usages of any (or enables selective ones via tslint directive). `any` remains for APIs to return user data to the user, but for APIs that accept user data, `unknown` is used instead.


schmidt-sebastian:

PR #502, opened 2018-12-28
Fix destructuring for getAll

### LINKED ISSUE #501: Typings: Cannot destructure array anymore when calling getAll
Since PR #443 has been merged, the signature for `Firestore.getAll()` has changed from

```javascript
getAll(...documents: DocumentReference[]): Promise<DocumentSnapshot[]>
```

to

```javascript
getAll(documentRef: DocumentReference, ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>): Promise<DocumentSnapshot[]>
```

with this change, we cannot use array destructuring anymore to "prepare" the refs to fetch in advance.

For example, I used to do the following:

```javascript
// step 1 - select the documents to fetch
const userIDs = ... 
const usersRef = admin.firestore().collection("users")
const userRefs: DocumentReference[] = userIDs.map(id => usersRef.doc(id))

// step 2 - Grab-them-all
const snaps = await admin.firestore().getAll(...userRefs)
``` 

this won't work anymore (I tried to update TS but it did not fix it), the only workaround is to do something like:

```javascript
const snaps = await admin.firestore().getAll(
  // extract the first item
  userRefs[0],
  // and then process the remaining elements
  ...(userRefs.slice(1))
)
```

#### Environment details

  - OS: Mojave
  - Node.js version: 10.12.0 (TS 3.2.2)
  - npm version: 6.4.1
  - `@google-cloud/firestore` version: 0.19.0

#### Steps to reproduce

  1. Create a project with Typescript support
  2. Type the code above (with `.getAll(...userRefs)`) → it won't compile

Edit: added #502 to restore previously-working behavior

Regards

ushu:
This is a proposal to fix #501 : it literally restores the previous array-based signature, while keeping support for `ReadOptions`-typed arguments.

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -147,8 +147,7 @@ declare namespace FirebaseFirestore {
      * snapshots.
      */
     getAll(
-        documentRef: DocumentReference,
-        ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>
+        ...documentRefs: Array<DocumentReference|ReadOptions>
     ): Promise<DocumentSnapshot[]>;
 
     /**


ushu:
I signed the CLA docs

schmidt-sebastian:
Hi ushu@! Thanks for doing this! While this should work for TypeScript users, it would be nice if we also cleaned up the SDK code itself (https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L707 and https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/transaction.ts#L161). Do you want to take a stab at that as well? Let us know if you are busy and we can take over :)

Note that the original intent of this signature was that to make it clear that the order of arguments is "DocumentReferences followed by field mask". This is probably not worth breaking our users over though... Thanks for catching this!

mikelehen:
Superseded by #515.

PR #508, opened 2018-12-28
Cleanup: Enum validation

schmidt-sebastian:

This moves the validation of acceptable "enum" values in the Query API into the shared validation.ts file. To do this, I used the validation code from the Web SDK. As this code prints the acceptable value in the error message, I removed support for non-documented API shorthands such as `=` (instead of `==`).


schmidt-sebastian:

PR #498, opened 2018-12-27
Cleanup: Enforce Import ordering

schmidt-sebastian:
This is part of #512 but can be reviewed independently.

This PR cleans up all the imports for #510 and uses `tslint` to enforce ordering. I manually moved the imports around to group them into cohesive blocks before invoking the linter.

PR #510, opened 2018-12-28
Finish TypeScript Migration (BREAKING CHANGE)

schmidt-sebastian:
This PR finishes the Typescript migration of the Firestore Server SDK. As part of this:

- I changed the Validator class to use free-standing functions. This gets rid of the dynamic method creation, which doesn't really work with strong typing
- I changed the client to use `unknown` instead of `any` for the user-facing types (only internally, not in the types we shipped)
- I brought the client more in line with the Web SDK and only support the direction/query operator spelling that is outlined in the type definition  (for example, "ASC" no longer works). This lets me use a common property validation function (and is the only breaking change)
- I removed the dependency on `is` to be able to use TypeScript's typechecking.

This is the PR that I am planning on merging, and it's also the PR that I am would like to address all feedback in (to reduce merge conflicts). Since it is not very easy to review, I split this out into a series of PRs that are more tightly scoped:

- #495 Cleanup: Argument count validation
- #496 Cleanup: Validate primitive types
- #497 Cleanup: Move parseGetAllArguments
- #498 Cleanup: Enforce Import ordering
- #499 Cleanup: Remove validator
- #500 Cleanup: Remove 'any' 
- #503 Cleanup: Remove `is`
- #504 Cleanup: Add strict types for the transaction test
- #505 Cleanup: Field value validation
- #506 Cleanup: Misc
- #507 Cleanup: Error messages for custom objects
- #508 Cleanup: Enum validation

The individual PRs are not meant to compile/pass tests. There should not be anything in this larger PR (which should pass CI, fingers crossed) that is not part of the other PRs.

This is fairly low priority, so please take your time to review.

Fixes: b/117464450 and b/119350730


schmidt-sebastian:
@JustinBeckwith / Node team: 

This PR fails CI with the following errors:

`
/Users/mrschmidt/GitHub/googleapis/nodejs-firestore/dev/src/serializer.ts
Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead. (array-type)
`

Note that `gts check` passes (`gts --version` reports 0.8.0, even though 0.9.0 is installed), but `npm run gts check` does not (`npm rum gts --version` reports 6.2.0). `gts fix` actually converts `Array<unknown>` to `unknown[]`.

PR #512, opened 2018-12-29
Finish TypeScript Migration

schmidt-sebastian:
This PR finishes the Typescript migration of the Firestore Server SDK. As part of this:

- I changed the Validator class to use free-standing functions. This gets rid of the dynamic method creation, which doesn't really work with strong typing
- I changed the client to use `unknown` instead of `any` for the user-facing types (only internally, not in the types we shipped)
- I brought the client more in line with the Web SDK and only support the direction/query operator spelling that is outlined in the type definition  (for example, "ASC" no longer works). This lets me use a common property validation function (and is the only breaking change)
- I removed the dependency on `is` to be able to use TypeScript's typechecking.

This is the PR that I am planning on merging, and it's also the PR that I am would like to address all feedback in (to reduce merge conflicts). Since it is not very easy to review, I split this out into a series of PRs that are more tightly scoped:

- ~~#495 Cleanup: Argument count validation~~ (approved)
- ~~#496 Cleanup: Validate primitive types~~ (approved)
- ~~#497 Cleanup: Move parseGetAllArguments~~ (approved)
- #498 Cleanup: Enforce Import ordering
- #499 Cleanup: Remove validator
- #500 Cleanup: Remove 'any' 
- ~~#503 Cleanup: Remove `is`~~ (approved)
- ~~#504 Cleanup: Add strict types for the transaction test~~ (approved)
- ~~#505 Cleanup: Field value validation~~ (approved)
- ~~#506 Cleanup: Misc~~ (approved)
- ~~#507 Cleanup: Error messages for custom objects~~ (approved)
- #508 Cleanup: Enum validation (<-- this is the breaking change)
- 549d90e Moves the field value validation to the serializer file

The individual PRs are not meant to compile/pass tests. There should not be anything in this larger PR  that is not listed above.

This is fairly low priority, so please take your time to review.

Fixes: b/117464450 and b/119350730


schmidt-sebastian:
@JustinBeckwith / Node team: 

This PR fails CI with the following errors:

`
/Users/mrschmidt/GitHub/googleapis/nodejs-firestore/dev/src/serializer.ts
Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead. (array-type)
`

Note that `gts check` passes (`gts --version` reports 0.8.0, even though 0.9.0 is installed), but `npm run gts check` -- as run by the CI -- does not (`npm rum gts --version` reports 6.2.0). `gts fix` actually converts `Array<unknown>` to `unknown[]`.

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+              arg,
+              'precondition')} "lastUpdateTime" is not a Firestore Timestamp.`);
+    }
+  }
+
+  if (conditions > 1) {
+    throw new Error(`${
+        invalidArgumentMessage(
+            arg, 'precondition')} Input contains more than one condition.`);
+  }
```
mikelehen: Should this be checking for invalid preconditions?  E.g. if I typo and do `{exist: true}` instead of `{exists: true}`

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+}
+
+
+/**
+ * Validates the use of 'value' as an update Precondition.
+ *
+ * @private
+ * @param arg The argument name or argument index (for varargs methods).
+ * @param value The object to validate.
+ * @param options Optional validation options specifying whether the valued can
```
mikelehen: "valued" => "value" here and elsewhere.

dev/src/validate.ts
```diff
@@ -14,194 +14,313 @@
...
 
 /**
- * Formats the given word as plural conditionally given the preceding number.
+ * Options to allow argument omission.
  *
  * @private
  */
-function formatPlural(num, str) {
-  return `${num} ${str}` + (num === 1 ? '' : 's');
+export interface AllowOptional {
```
mikelehen: nit: I don't love `AllowOptional` and `AllowRange` as names.  Normally I like to name classes / interfaces as nouns.  I think maybe `OptionalOption` and `RangeOptions` would be better, though admittedly `OptionalOption` is a bit awkward. :)

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+ * Validates the use of 'value' as a Precondition and enforces that 'exists'
+ * and 'lastUpdateTime' use valid types.
+ *
+ * @private
+ * @param arg The argument name or argument index (for varargs methods).
+ * @param value The object to validate
+ * @param allowExist Whether to allow the 'exists' preconditions.
+ */
+function validatePrecondition(
+    arg: string|number, value: unknown, allowExist: boolean): void {
```
mikelehen: allowExist => allowExists ? (you use "/*allowExists=*/" elsewhere)

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+          invalidArgumentMessage(
+              arg,
+              'precondition')} "lastUpdateTime" is not a Firestore Timestamp.`);
+    }
+  }
+
+  if (conditions > 1) {
+    throw new Error(`${
+        invalidArgumentMessage(
+            arg, 'precondition')} Input contains more than one condition.`);
```
mikelehen: This error and above use "condition" instead of "precondition". Is that intentional? We should probably use whatever leads to the most consistency across our APIs / docs / errors.

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+      }
+    }
+
+    if ('merge' in setOptions && 'mergeFields' in setOptions) {
+      throw new Error(`${
+          invalidArgumentMessage(
+              arg,
+              'set() option')} You cannot specify both "merge" and "mergeFields".`);
+    }
+  }
```
mikelehen: Again, should we be checking for extra invalid items in setOptions?  FWIW- the client SDK has `validateOptionNames()` to do that.  You could also consider borrowing `validateNamedOptionalType()` to replace your manual type-checking of 'merge', 'mergeFields' etc.  Up to you though.

dev/src/write-batch.ts
```diff
@@ -589,15 +575,198 @@ export class WriteBatch {
...
+/**
+ * Validates the use of 'value' as SetOptions and enforces that 'merge' is a
+ * boolean.
+ *
+ * @private
+ * @param arg The argument name or argument index (for varargs methods).
+ * @param value The object to validate.
+ * @param options Optional validation options specifying whether the valued can
+ * be omitted.
+ * @returns 'true' if the input is a valid SetOptions object.
```
mikelehen: Should be `@throws ...`.  I think several other methods are wrong too.

dev/test/document.ts
```diff
@@ -957,29 +958,29 @@ describe('set document', () => {
       firestore.doc('collectionId/documentId').set({foo: 'bar'}, 'foo');
     })
         .to.throw(
-            'Argument "options" is not a valid SetOptions. Input is not an object.');
+            'Argument "options" is not a valid set() option. Input is not an object.');
```
mikelehen: This reads a bit weird.  Maybe "is not a valid set() options object." ?  though admittedly that's a bit verbose.

dev/src/document.ts
```diff
@@ -349,7 +348,7 @@ export class DocumentSnapshot {
    *   console.log(`Retrieved data: ${JSON.stringify(data)}`);
    * });
    */
-  data(): DocumentData|undefined {
+  data(): {[field: string]: any}|undefined {  // tslint:disable-line no-any
```
mikelehen: Are all of these remaining instances of "any" expected to stay forever? Or do you intend further cleanup passes?  In my ideal world we'd either remove them or comment them. :-)  This one may be intentional so that our users don't have to do instanceof checks on the data they get back?  I'm not sure though.

dev/src/reference.ts
```diff
@@ -1942,81 +1921,77 @@ export class CollectionReference extends Query {
...
+ * @throws when the direction is invalid
  */
-export function validateFieldOrder(str: string): boolean {
-  if (!is.string(str) || !is.defined(directionOperators[str])) {
-    throw new Error('Order must be one of "asc" or "desc".');
-  }
-
-  return true;
+export function validateQueryOrder(arg: string|number, op: unknown): void {
+  validateEnumValue(arg, op, Object.keys(directionOperators), {optional: true});
```
mikelehen: Per discussions with Gil, etc. during the push to GA, I think we want to unroll the breaking change we made here (and elsewhere).  And in fact we may want to relax the mobile SDK to match.

dev/src/reference.ts
```diff
@@ -60,7 +58,6 @@ const comparisonOperators:
     {[k: string]: api.StructuredQuery.FieldFilter.Operator} = {
       '<': 'LESS_THAN',
       '<=': 'LESS_THAN_OR_EQUAL',
-      '=': 'EQUAL',
       '==': 'EQUAL',
```
mikelehen: Per discussions with Gil, etc. during the push to GA, I think we want to continue to allow '=' to avoid the breaking change.

dev/src/document.ts
mikelehen: > by using `any` as a return type we indicate that the user has to do some type checking of their own

Sorry for the nit, but that's not actually what `any` does. 😀  `any` makes it so you *don't* have to do any type checking (you can do anything you want with it and TypeScript won't complain).  If we want to force the user to do type-checking, then we should return `unknown` (new in TS 3.x) instead, which you can't do anything with unless you first do type-checking on it.

That said, using any here is consistent with the mobile SDK (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L24), so I'm fine with this.

dev/src/reference.ts
```diff
@@ -785,6 +786,8 @@ export class QuerySnapshot {
    *   });
    * });
    */
+  // We deliberately use `any` to allow users to use arbitrary objects as the
+  // callback's context.
```
mikelehen: I think you can use `unknown` here since we don't care what the type is, we just pass it through.

dev/src/validate.ts
```diff
@@ -165,7 +169,7 @@ export function validateBoolean(
  */
 export function validateNumber(
     arg: string|number, value: unknown,
-    options?: AllowOptional&AllowRange): void {
+    options?: RequiredArgumentOptions&NumericRangeOptions): void {
```
mikelehen: Is this really how clang-format formats this?  Yuck. 😒

dev/test/util/helpers.ts
```diff
@@ -14,20 +14,22 @@
  * limitations under the License.
  */
 
+const v1beta1 = require('../../src/v1beta1');
```
mikelehen: unused?

schmidt-sebastian: Removed. Bye bye v1beta1

dev/src/document.ts
schmidt-sebastian: Uhm, that's what I meant (kind of, but not really).

Changed to
```
  // We deliberately use `any` in the external API to not impose type-checking
  // on end users.
```

dev/src/reference.ts
schmidt-sebastian: Yes, that works. Not sure how `.call()` is defined, but the linter doesn't complain.

PR #515, opened 2019-01-08
Overload getAll function to allow array destructuring

### LINKED ISSUE #501: Typings: Cannot destructure array anymore when calling getAll

cxam:
Fixes #501

This PR overloads the getAll function in Firestore and Transaction class to allow passing in a destructured array of documents.

- [x] Tests and linter pass
- [x] Code coverage does not decrease
- [x] Appropriate docs were updated

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -703,6 +703,8 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
+  getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
+      Promise<DocumentSnapshot[]>;
   getAll(
       documentRef: DocumentReference,
       ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -703,16 +703,13 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
-  getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
-      Promise<DocumentSnapshot[]>;
-  getAll(
-      documentRef: DocumentReference,
-      ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
-      Promise<DocumentSnapshot[]> {
+  getAll(...documentRefsOrReadOptions: [
+    DocumentReference, ...Array<DocumentReference|ReadOptions>
+  ]): Promise<DocumentSnapshot[]> {
     this._validator.minNumberOfArguments('Firestore.getAll', arguments, 1);
 
-    const {documents, fieldMask} = parseGetAllArguments(
-        this._validator, [documentRef, ...moreDocumentRefsOrReadOptions]);
+    const {documents, fieldMask} =
+        parseGetAllArguments(this._validator, [...documentRefsOrReadOptions]);
     return this.getAll_(documents, fieldMask, requestTag());
   }
 
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -703,13 +703,12 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
-  getAll(...documentRefsOrReadOptions: [
-    DocumentReference, ...Array<DocumentReference|ReadOptions>
-  ]): Promise<DocumentSnapshot[]> {
+  getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
+      Promise<DocumentSnapshot[]> {
     this._validator.minNumberOfArguments('Firestore.getAll', arguments, 1);
 
     const {documents, fieldMask} =
-        parseGetAllArguments(this._validator, [...documentRefsOrReadOptions]);
+        parseGetAllArguments(this._validator, documentRefsOrReadOptions);
     return this.getAll_(documents, fieldMask, requestTag());
   }
 
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -687,9 +687,12 @@ export class Firestore {
   /**
    * Retrieves multiple documents from Firestore.
    *
-   * @param {DocumentReference} documentRef A `DocumentReference` to receive.
-   * @param {Array.<DocumentReference|ReadOptions>} moreDocumentRefsOrReadOptions
-   * Additional `DocumentReferences` to receive, followed by an optional field
+   * The first argument is required and must be of type `DocumentReference`
+   * followed by any additional `DocumentReference` documents. If used, the
+   * optional `ReadOptions` must be the last argument.
+   *
+   * @param {Array.<DocumentReference|ReadOptions>} documentRefsOrReadOptions
+   * The `DocumentReferences` to receive, followed by an optional field
    * mask.
    * @returns {Promise<Array.<DocumentSnapshot>>} A Promise that
    * contains an array with the resulting document snapshots.
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -77,6 +77,18 @@ describe('Firestore class', () => {
         });
   });
 
+  it('getAll() supports array destructuring', () => {
+    const ref1 = randomCol.doc('doc1');
+    const ref2 = randomCol.doc('doc2');
+    return Promise.all([ref1.set({foo: 'a'}), ref2.set({foo: 'a'})])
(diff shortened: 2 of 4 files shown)


dev/src/index.ts
```diff
@@ -703,6 +703,8 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
+  getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
+      Promise<DocumentSnapshot[]>;
   getAll(
       documentRef: DocumentReference,
       ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
```
mikelehen: Interesting...  I thought the way overloads work in Typescript is that you list all the overloads and then you define the function so that its type is a superset of all of the overloads, but the function itself doesn't actually count as an overload.  So this would mean that:

1. This isn't actually defining multiple overloads, since the function implementation itself doesn't count as an overload.
2. I am surprised TypeScript even accepts this because the overload `getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>)` would accept `getAll()` or `getAll(readOptions)` but the actual function (`getAll(documentRef: DocumentReference, ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>)` does not.  So the actual function definition is not a superset of all the overloads.

So I'm kinda' confused what is going on here.  But taking a step back I think overloads may not actually help us (sorry!).  I think the ideal function signature would be something like:

```
getAll(documentRef: DocumentReference, ...moreDocumentRefs: Array<DcoumentReference>, readOptions?: ReadOptions)
```

But TypeScript doesn't support this.  The next best signature is the one we have checked in right now, but that apparently breaks array destructuring (maybe because the array isn't guaranteed to have at least 1 element to match the leading `documentRef` argument?).

So I think that means that the best we can do is basically what you did originally and forego overloads and just make the signature:

```
getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
      Promise<DocumentSnapshot[]>;
```

This is overly permissive, since it allows `getAll()`, `getAll(readOptions)`, `getAll(doc, readOptions, doc, readOptions), etc. which are all invalid...  but I think it's the best we can do with TypeScript right now (there's an open issue that might help us if it's ever implemented: https://github.com/Microsoft/TypeScript/issues/1360)

cxam: Thanks for the feedback. You're right, unfortunately that's the best we can do right now with what's available. I will update the PR with an alternative solution for the original signature that will be less permissive (won't allow `getAll()` at least). Let me know what you think.

cxam:
@mikelehen I have just pushed an update to the PR that uses a tuple type along with the rest element. What this means is that the first parameter is required and must be of type `DocumentRefence` followed by the rest. This is just an improvement over the original signature to prevent calling `getAll()` without any matching parameters or an incorrect one (`getAll(ReadOptions)`).

dev/src/index.ts
```diff
@@ -703,14 +703,13 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
-  getAll(
-      documentRef: DocumentReference,
-      ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
-      Promise<DocumentSnapshot[]> {
+  getAll(...documentRefsOrReadOptions: [
+    DocumentReference, ...Array<DocumentReference|ReadOptions>
+  ]): Promise<DocumentSnapshot[]> {
```
mikelehen: I'm a little surprised that this behaves any differently than the original code.  In fact, I am somewhat suspicious.  I just noticed that our tests aren't actually being properly type-checked (maybe because we haven't landed https://github.com/googleapis/nodejs-firestore/pull/512 yet).  So I'm not sure your tests are actually validating anything. 😢 

Would it be possible for you to validate this change in your original project / app where you hit the destructuring problem?  (e.g. by manually applying the fix to the .d.ts file)

dev/system-test/firestore.ts
```diff
@@ -1383,6 +1408,22 @@ describe('Transaction class', () => {
...
+    return Promise.all([ref1.set({}), ref2.set({})])
+        .then(() => {
+          return firestore.runTransaction(updateFunction => {
+            return updateFunction.getAll(...[ref1, ref2]).then(docs => {
+              return Promise.resolve(docs.length);
+            });
+          });
+        })
+        .then(res => {
+          expect(res).to.equal(2);
```
mikelehen: This test is actually failing for me.

dev/src/index.ts
```diff
@@ -703,14 +703,13 @@ export class Firestore {
...
     this._validator.minNumberOfArguments('Firestore.getAll', arguments, 1);
 
-    const {documents, fieldMask} = parseGetAllArguments(
-        this._validator, [documentRef, ...moreDocumentRefsOrReadOptions]);
+    const {documents, fieldMask} =
+        parseGetAllArguments(this._validator, [...documentRefsOrReadOptions]);
```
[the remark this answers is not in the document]
cxam: Fixed.

dev/src/index.ts
cxam: Just validated, you're right, doesn't work as intended when I manually applied the fix on my app. I was too trusting of VSCode to type check, wasn't aware of #512, sorry. Will push the updated PR with the single argument.

dev/system-test/firestore.ts
cxam: Should be fixed with the revert.

cxam:
No worries! Just updated the PR with single signature. This should hopefully maintain compatibility albeit being overly permissive.

dev/src/index.ts
```diff
@@ -703,14 +703,12 @@ export class Firestore {
    *   console.log(`Second document: ${JSON.stringify(docs[1])}`);
    * });
    */
-  getAll(
-      documentRef: DocumentReference,
-      ...moreDocumentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
+  getAll(...documentRefsOrReadOptions: Array<DocumentReference|ReadOptions>):
```
schmidt-sebastian: Thanks for doing this :) One nit from my end: Can you also update the JSDoc for these methods?

cxam: Updated and verified. I will also include an additional comment regarding the required order of arguments.

cxam:
@schmidt-sebastian JSDoc updated in latest commit. I have also fixed the failing test for `Transaction class - getAll() supports array destructuring with field mask`.

Also, I noticed that `npm test` doesn't actually run the system tests. I understand that this is due to the credentials needed to run this properly as it requires access to a Google Cloud Project. Should the contributing guide be updated to include instructions on how to run these (i.e. setting of `GCLOUD_PROJECT` and `GOOGLE_APPLICATION_CREDENTIALS` environment variables)?

mikelehen:
@JustinBeckwith @crwilcox Either of you know if there's a trick to making the CI run (for an externally contributed PR)?

cxam:
@mikelehen I think last time this worked when @kinwa91 assigned the `kokoro:force-run` label.

kinwa91:
Anyone with `write` access can apply the `kokoro:force-run` label to jumpstart the test.

mikelehen:
@cxam @kinwa91 Nice, thanks!

PR #516, opened 2019-01-11
feat: update to v1 protos

mikelehen:
Update the nodejs-firestore client to use the v1 protos and endpoint. Mostly this is a no-op, though I did have to tweak the way we serialize empty document paths because the backend now requires `/project/[project]/database/[database]/documents` for empty document paths instead of `/project/[project]/database/[database]`

The v1beta1 gapic client is preserved for now and can still be used, but the manually-curated API uses v1.


JustinBeckwith:
@crwilcox @jadekler Are we removing the `v1beta1` library in other languages?

jadekler:
@JustinBeckwith We decided not to. @mikelehen could you leave the v1beta1 gapic as-is, and just switch the manual to using v1? This helps users currently relying on the beta (because they remain unbroken for a little while longer), but mostly just makes the process a bit more seamless.

mikelehen:
I can undelete dev/src/v1beta1 (and dev/protos/google/firestore/v1beta/ too?) but I'm unclear what benefit this provides.  We have to delete it before we go 1.x, right?  And so if users depend on it directly, they'll need to stick with a 0.x version.

mikelehen:
@JustinBeckwith Per discussions, I've updated this PR to keep the v1beta1 gapic client and expose both v1beta1 and v1 gapic clients to developers, but the manual SDK uses v1 internally.

@kinwa91 The docs check you added in #511 is failing because of two broken links that came from the v1 protos. I see you added steps to synth.py to automatically fix them, but when I re-run synth.py I'm actually seeing it revert your changes to .kokoro/docs.sh.  Am I doing something wrong?

mikelehen:
I went ahead and checked in just the proto changes necessary to make CI pass, but I had to revert some of the other changes synth.py made.

mikelehen:
@kinwa91 v1beta1 has been revived. Can you approve?  Thanks!

kinwa91:
synthtool hasn't been released to reflect the changes to docs.sh. I will cut a release today. LGTM

Issue #525, opened 2019-01-21
Enable noImplicitAny?


joecervino:
I was wondering if this library is ready to be worked on to enable noImplciitAny. I see a TODO in the tsconfig file asking to remove the "false" assignment for noImplicitAny once source is finished migrating to TS

mikelehen:
Unfortunately not. There's a large outstanding PR to migrate a bunch more code to TypeScript: https://github.com/googleapis/nodejs-firestore/pull/512  Any work done to remove noImplicitAny would need to happen after that PR lands.

PR #553, opened 2019-02-22
refactor(typescript): enable noImplicitAny

schmidt-sebastian:
This removes `no-implicit-any` from the SDK.

Fixes: https://github.com/googleapis/nodejs-firestore/issues/525

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -21,9 +21,8 @@ import * as extend from 'extend';
 import * as through2 from 'through2';
 
 import {google} from '../protos/firestore_proto_api';
-import {documentFromJson, timestampFromJson} from './convert';
-import {DocumentSnapshot, DocumentSnapshotBuilder} from './document';
-import {GeoPoint} from './geo-point';
+import {fieldsFromJson, timestampFromJson} from './convert';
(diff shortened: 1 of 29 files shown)


dev/src/index.ts
```diff
@@ -461,41 +457,84 @@ export class Firestore {
...
+   *
+   * This API is used by Google Cloud Functions.
+   *
+   * @private
+   * @param document The Firestore 'Document' proto in JSON format.
+   * @param readTime An ISO-8601 formatted string indicating the time this
+   * document was read.
+   * @param encoding 'json' to indicate the format of `document` and `readTime`.
+   * @returns The QueryDocumentSnapshot with the document's data.
    */
```
mikelehen: Hrm.  Duplicating all the docs for each overload like this seems a bit verbose and likely to get out-of-sync over time.  Was there a particular motivation?  I don't feel strongly, so feel free to leave as-is.

schmidt-sebastian: This was mostly intended to get IntelliJ to stop warning me about missing docs for some of the arguments (since the overloads use different argument names). Since I wasn't 100% convinced of this when I did it, I removed this part of the changes. I did keep the overloads though.

Issue #109, opened 2018-01-18
Add Generic Types


bmass02:
## Proposal
Add Generic Types with defaults to DocumentReference, CollectionReference, Query, etc.
```typescript
export class DocumentReference<D = DocumentData, U = UpdateData, P = CollectionReference<D, U>|null, S = {[subCollectionId: string]: CollectionReference}> {
```
D and U are pretty straight forward, but P would be for the parent type and S would be for the typing of subcollections. This would still be backwards compatible as providing no types would default to what is currently typed.

Currently if I want to achieve any of this I have to create wrapper interfaces for any Firestore classes I want to use and retype all of the methods, then type the collections/documents with those wrappers.

I'd be happy to make a PR with the necessary changes if you think this change is beneficial.

## Reasons
1. Typed data returned from DocumentSnapshot (i.e. eliminates the need to import types and explicitly type variables containing the data.)
2. Enforcement of parent types and subcollections and their types
3. Enforcement of update rules (i.e. if certain values should only be written on create, or if one value is changed, another should also be updated)
4. Easier to extend Firestore classes for modeling purposes (only as interface since constructors are private)

```typescript
const chatRooms = firestore.collection<ChatRoom, ChatRoomUpdateRules>('rooms')
const myChatRoomRef = chatRooms.doc(myChatRoomId)
const myChatRoomDoc = await myChatRoomRef.get()
const myChatRoomInfo = myChatRoomDoc.data() //myChatRoomInfo is of type ChatRoom

const members = myChatRoomDoc.ref.collection<Member, MemberUpdateRules>('members')
const certainMembersQuery = members.where('posts', '>=', 3).orderBy('posts').limit(10)
certainMembersQuery.onSnapshot((membersSnapshot) => {
  membersSnapshot.docChanges.forEach((change) => {
    const memberInfo = change.doc.data() //memberInfo is of type Member
  })
  membersSnapshot.docs.forEach((doc) => {
    const memberInfo = doc.data() //memberInfo is of type Member
    const otherChatRoomRef = doc.ref.parent.parent.parent.doc(otherChatRoomId) // Maintains typings of parents
  })
})
```

### Enforcing subcollections
```typescript
import * as admin from 'firebase-admin'
import * as Event from './event'

interface IOrganizationModel {
  ...
}

interface IOrganizationUpdateRules {
  ...
}

interface IOrganizationSubcollections {
  events: Event
}

[...]

PR #554, opened 2019-02-24
Make DocumentSnapshot .data() generic

polRk:
```
functions.firestore
  .document('agencies/{agencyID}/tasks/{taskID}')
  .onCreate((snapshot:  FirebaseFirestore.DocumentSnapshot<{name: string}>, context) => true);

 FirebaseFirestore.DocumentSnapshot<{name: string}>
```

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -676,7 +676,7 @@ declare namespace FirebaseFirestore {
    * access will return 'undefined'. You can use the `exists` property to
    * explicitly verify a document's existence.
    */
-  export class DocumentSnapshot {
+  export class DocumentSnapshot<D = DocumentData> {
     protected constructor();
 
     /** True if the document exists. */
@@ -713,7 +713,7 @@ declare namespace FirebaseFirestore {
      *
      * @return An Object containing all fields in the document.
      */
-    data(): DocumentData | undefined;
+    data(): D | undefined;
 
     /**
      * Retrieves the field specified by `fieldPath`.


schmidt-sebastian:
Thank you for sending this over! This is a feature request that comes around every once in a while (see https://github.com/googleapis/nodejs-firestore/issues/109), but we have so far decided to keep things as is. Out client doesn't know the type of the data that it is returning, and hence any "casting" that we do is not safe. Since there are no guarantees, we would rather make this explicit by returning `any` from `DocumentSnapshot.data()`.

polRk:
But you can add non required generic types , like 
```
class A <D = number> {
}

new A<string>();
new A();
```
If i know, the type of the data, i want use generic. 
And in pull request , i added support generic types to some functions.

polRk:
Add returning `any` - It's a bad practice .

Robula:
This change would be awesome, `data()` _really_ needs to be generic so we can avoid casting all together.

mikelehen:
Just to be clear, there fundamentally has to be a cast here.  The question is whether Firestore would do this for you implicitly or whether the developer has to do it explicitly.  Our current position is that it's better if the developer does it  explicitly so that they are aware there's a non-type-safe operation going on.  

FWIW- Returning `any` is consistent with e.g. the typings TypeScript provides for IndexedDb (https://github.com/Microsoft/TypeScript/blob/84087d0e0a795eee3b1a8cb1988e1285b13cdd5e/lib/lib.dom.d.ts#L9165).

fenixsolorzano:
+1, Dealing with explicit casting is kinda awful, sometimes we end up using strings in our code just to safely do casting:

```ts
const myType: MyTypeInterface = {
  fieldA: snapshot.get('fieldA'),
  fieldB: snapshot.get('fieldB'),
} 
```

It would be awesome if at least the functions that require field names or strings would support some kind of typed parameters by using `keyof T`...

PR #565, opened 2019-03-05
Don't call stream.end() on Watch ended by server

schmidt-sebastian:
This is part of https://github.com/googleapis/nodejs-firestore/pull/564

The changes in #564 to make `formattedName()` async changed a lot of the timing in the Watch code and uncovered an issue with the way we close streams. We should only call `stream.end()` if the SDK closes the stream. If the server closes the stream we don't call end() - if we do, we get an exception (that was previously swallowed).

This PR also introduces subclasses for Watch that provide the Target Proto and the Comaprator. This makes it possible to construct these after initialization time and to return a Promise.


schmidt-sebastian:
I'm gonna leave this unmerged for a bit to give @mikelehen a chance to comment.

PR #568, opened 2019-03-05
Serialize documents at commit time

schmidt-sebastian:
This makes PR #564 smaller. Since DocumentSnapshot.fromObject() and toProto() becomes async, this PR moves all serialization in a WriteBatch to the commit() call.


schmidt-sebastian:
Closing for now. This code is still part of #564, but can probably now be reviewed as part of that PR.

PR #578, opened 2019-03-11
feature: Adding CollectionGroup queries

schmidt-sebastian:
CollectionGroup queries add support for queries to Firestore that query documents by collection ID rather than by collection path. With this feature, users can now query for documents in collections that share a name, no matter where these collection are located.

This the Node port of https://github.com/googleapis/google-cloud-java/pull/4652. The integration tests are ported from https://github.com/firebase/firebase-js-sdk/pull/1440

Note: This PR is blocked on a backend release.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -867,15 +867,87 @@ interface QueryCursor {
   values: unknown[];
 }
 
-/** Internal options to customize the Query class. */
-interface QueryOptions {
-  startAt?: QueryCursor;
-  startAfter?: QueryCursor;
-  endAt?: QueryCursor;
-  endBefore?: QueryCursor;
-  limit?: number;
-  offset?: number;
-  projection?: api.StructuredQuery.IProjection;
+/**
+ * Internal class representing custom Query options.
+ *
+ * These options are immutable. Modified options can be created using `with()`.
+ * @private
+ */
+export class QueryOptions {
+  constructor(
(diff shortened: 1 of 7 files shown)


dev/src/reference.ts
```diff
@@ -867,15 +867,76 @@ interface QueryCursor {
...
+ *
+ * These options are immutable. Modified options can be created using `with()`.
+ * @private
+ */
+export class QueryOptions {
+  constructor(
+      readonly parentPath: ResourcePath, readonly collectionId: string,
+      readonly allDescendants: boolean, readonly fieldFilters: FieldFilter[],
+      readonly fieldOrders: FieldOrder[], readonly startAt?: QueryCursor,
+      readonly endAt?: QueryCursor, readonly limit?: number,
```
var-const: It looks like `startAfter` and `endBefore` were omitted, what is the reason? Looks like they were unused previously?

var-const:
I'm really sorry, I somehow managed to miss this PR entirely. I usually respond no later than the next day on code reviews; feel free to ping me if I don't.

dev/src/reference.ts
schmidt-sebastian: This is actually meant to mimic the proto, where `startAfter` and `endBefore` are just different settings in the QueryCurso (`startAt` sets before to `true`, whereas `startAfter` sets it to `false`).

PR #581, opened 2019-03-14
internal: Update Protos

schmidt-sebastian:
Pull in the newest Proto files and adding the new dependency on "Descriptor.proto"

There are no changes to the client.


schmidt-sebastian:
Closing as this may no longer be needed.

PR #610, opened 2019-05-06
build: patch Windows container, fixing Node 10

bcoe:
temporary patch broken Windows container, such that tests can be run on Node 10


samples/README.md
```diff
@@ -40,8 +40,18 @@ View the [source code](https://github.com/googleapis/nodejs-firestore/blob/maste
 
 __Usage:__
 
+<<<<<<< HEAD
```
callmehiphop: @bcoe looks like this guy snuck in here!

bcoe: Good catch! fixing.

PR #618, opened 2019-05-10
fix(deps): drop dependency on through2

JustinBeckwith:


dev/src/index.ts
```diff
@@ -1372,15 +1372,18 @@ export class Firestore {
...
-          const logStream = through2.obj(function(this, chunk, enc, callback) {
-            logger(
-              'Firestore.readStream',
-              requestTag,
-              'Received response: %j',
-              chunk
-            );
-            this.push(chunk);
-            callback();
+          const logStream = new PassThrough({
```
wilhuff: This can be a constructed by a `responseLogger` function that takes `methodName` and `requestTag`.

dev/src/index.ts
```diff
@@ -16,7 +16,7 @@
 
 import * as bun from 'bun';
 import {CallOptions} from 'google-gax';
-import * as through2 from 'through2';
+import {PassThrough, Transform} from 'stream';
```
wilhuff: Unused import.

Question, though: `PassThrough` is documented as a trivial `Transform` that passes its input through to its output. By overriding `_transform`, aren't we invalidating the purpose of the refined type? Why not make these all `Transform`?

dev/src/watch.ts
```diff
@@ -347,7 +346,7 @@ abstract class Watch {
 
     // We may need to replace the underlying stream on reset events.
     // This is the one that will be returned and proxy the current one.
-    const stream = through2.obj();
+    const stream = new PassThrough({objectMode: true});
```
wilhuff: This is a case where `PassThrough` is appropriate.

package.json
```diff
@@ -79,6 +78,7 @@
     "protobufjs": "^6.8.6",
     "proxyquire": "^2.0.1",
     "source-map-support": "^0.5.6",
+    "through2": "^3.0.0",
```
wilhuff: There aren't that many uses in tests so we could get rid of this too, though getting it out of the production code is a win enough.

dev/src/index.ts
JustinBeckwith: That....  is a fair question 😆

dev/src/index.ts
JustinBeckwith: Apologies - I do not know this code base terribly well :) I was just trying to do a 1 for 1 replacement of `through2` with Transform streams, now that we've dropped node.js 6.  Could you spell out the line of code you think I should be using here?

package.json
JustinBeckwith: It looks like the GAPIC generator we use under the hood is bringing it in for tests. I'm happy to go chase that down in another PR, but one step at a time :)

Just to set expectations - this module is still brought in via transitive dependencies, but again, one step at a time :)

dev/src/index.ts
wilhuff: Oh, I was just suggesting that these two could be combined. We can do that in another pass.

JustinBeckwith:
Meh, not worth it :) Guess I'll leave this one be.

PR #620, opened 2019-05-13
[docs] Further improve quickstart.js

schmidt-sebastian:
To avoid confusion, I think we should remove the `Document created` console output since `firestore.doc()` merely returns a reference. It does not talk to the backend.


schmidt-sebastian:
Thanks for the test fix :) I was unaware this test existed.

PR #623, opened 2019-05-15
fix: Support more than 100 long-lived streams

schmidt-sebastian:
Fixes firebase/firebase-admin-node#499

This is a different approach to https://github.com/googleapis/nodejs-firestore/pull/614 and solves the same problem. It re-uses more of the original code to achieve the same result. Instead of releasing the client back to the pool when a stream is marked as established, we instead keep it around until the stream is closed.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1454,7 +1454,6 @@ export class Firestore {
         resultStream.on('error', lifetime.resolve);
 
         await this._initializeStream(resultStream, requestTag, request);
-
         result.resolve(resultStream);
       }).catch(err => {
         lifetime.resolve();
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -46,7 +46,7 @@ import {
   ReadOptions,
   Settings,
 } from './types';
-import {requestTag} from './util';
+import {Deferred, requestTag} from './util';
 import {
   validateBoolean,
   validateFunction,
@@ -58,7 +58,6 @@ import {
 import {WriteBatch} from './write-batch';
 
 import api = google.firestore.v1;
-import {Deferred} from '../test/util/helpers';
 
 export {
   CollectionReference,
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1144,18 +1144,15 @@ export class Firestore {
    * @returns The given Stream once it is considered healthy.
    */
   private _initializeStream(
-    releaser: () => void,
     resultStream: NodeJS.ReadableStream,
     requestTag: string
   ): Promise<NodeJS.ReadableStream>;
   private _initializeStream(
-    releaser: () => void,
     resultStream: NodeJS.ReadWriteStream,
     requestTag: string,
     request: {}
   ): Promise<NodeJS.ReadWriteStream>;
   private _initializeStream(
-    releaser: () => void,
     resultStream: NodeJS.ReadableStream | NodeJS.ReadWriteStream,
     requestTag: string,
     request?: {}
@@ -1167,7 +1164,7 @@ export class Firestore {
      * Whether we have resolved the Promise and returned the stream to the
      * caller.
      */
-    let streamInitialized = false;
+    let streamReleased = false;
 
     /**
      * Whether the stream end has been reached. This has to be forwarded to the
@@ -1176,7 +1173,7 @@ export class Firestore {
     let endCalled = false;
 
     return new Promise((resolve, reject) => {
-      const streamReady = () => {
+      const releaseStream = () => {
         if (errorReceived) {
           logger(
             'Firestore._initializeStream',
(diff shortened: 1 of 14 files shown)


dev/src/index.ts
```diff
@@ -1176,7 +1174,7 @@ export class Firestore {
     let endCalled = false;
 
     return new Promise((resolve, reject) => {
-      const streamReady = () => {
+      const releaseStream = () => {
```
wilhuff: I don't like this name. "released" can be confused with releasing back into the pool. I certainly was confused that way when reading this.

This promise represents that the stream is ready for use or initialized or something like that.

dev/src/index.ts
```diff
@@ -1416,11 +1423,15 @@ export class Firestore {
...
+
+    this._clientPool.run(gapicClient => {
+      // While we return the stream to the callee early, we don't want to
+      // release the GAPIC client until the callee has finished processing the
+      // stream.
+      const lifetime = new Deferred<void>();
 
-    return this._retry(attempts, requestTag, () => {
-      return Promise.resolve().then(() => {
+      this._retry(attempts, requestTag, async () => {
```
wilhuff: Isn't this subtly different from the original code? Doesn't `Promise.resolve().then(...)` give the runtime the opportunity to go to the next tick? Did the original code intend that?

dev/src/index.ts
```diff
@@ -1436,14 +1447,21 @@ export class Firestore {
...
-        return this._initializeStream(
-          releaser,
-          resultStream,
-          requestTag,
-          request
-        );
-      });
+        return this._initializeStream(resultStream, requestTag, request);
+      })
+        .then(stream => {
```
wilhuff: async/await?

dev/src/index.ts
```diff
@@ -1359,36 +1354,47 @@ export class Firestore {
...
-          logger('Firestore.readStream', requestTag, 'Received error:', err);
-          reject(err);
-        }
-      }).then(stream => this._initializeStream(releaser, stream, requestTag));
+          this.push(chunk);
+          callback();
+        });
+
+        const resultStream = bun([stream, logStream]);
+        resultStream.on('close', lifetime.resolve);
```
wilhuff: Are streams guaranteed to emit the `close` event? What happens in the case of an error?

dev/system-test/firestore.ts
```diff
@@ -913,6 +914,37 @@ describe('DocumentReference class', () => {
...
+            } else if (snapshot.size === 1) {
+              documentResults[i].resolve();
+            }
+          });
+      }
+
+      await Promise.all(emptyResults.map(d => d.promise));
+      ref.set({i: 1337});
+      await Promise.all(documentResults.map(d => d.promise));
+      unsubscribeCallbacks.forEach(c => c());
```
wilhuff: This test verifies that all 150 listeners succeed but doesn't verify that everything has been properly released to the pool.

Is it possible to check that pool.size is 150 once the listeners are started and then get back to zero after?

wilhuff: I worry that this test can succeed even if you remove the line that resolves the lifetime promise.

wilhuff:
This is now failing tests under node 8:

```
  1) Query watch
       re-opens on unexpected stream end:
      Expected message of type 'end' but got 'open' with '{}'.
      + expected - actual
      -open
      +end
      at DeferredListener.await (build/test/watch.js:143:41)
      at StreamHelper.await (build/test/watch.js:186:43)
      at watchHelper.await.then (build/test/watch.js:604:41)
      at <anonymous>
  2) Query watch
       "after each" hook for "re-opens on unexpected stream end":
      AssertionError: expected 1 to equal 0
      + expected - actual
      -1
      +0
      at Object.verifyInstance (build/test/util/helpers.js:87:56)
      at Context.afterEach (build/test/watch.js:544:19)
```

schmidt-sebastian:
The test failure for Node 8 was due to a flake, which I fixed.

jakeleventhal:
@wilhuff @schmidt-sebastian I just upgraded to `@google-cloud/firestore@2.0.0` and since upgrading, I've been getting all of these warnings that say:
`(node:2870) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connectivityStateChanged listeners added. Use emitter.setMaxListeners() to increase limit`

I read through the release notes and found that this thread is probably the reason that I am seeing these warnings. I tried downgrading back to `v1.3.0` and no longer saw the messages in my logs. I believe that it makes sense to take care of the warning that I'm seeing within the `@google-cloud/firestore` package itself, as the user doesn't really have a good way to know when the listeners will be added.

What are your guys' thoughts?

schmidt-sebastian:
!@jakeleventhal Thanks for letting us know! I think I have seen this warning once before as well, so it may indeed be related to this change (which adds a lot of listeners). I filed https://github.com/googleapis/nodejs-firestore/issues/661 to track this.

PR #624, opened 2019-05-15
refactor(deprecation)!: remove deprecated getCollections()

schmidt-sebastian:
BREAKING CHANGE: This removes the deprecated getCollections() method (deprecated in Oct 2018). This will be part of the v2 release.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -813,19 +813,6 @@ export class Firestore {
   }
 
   /**
-   * Fetches the root collections that are associated with this Firestore
-   * database.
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
Just FYI - updated the PR title and desc to more closely match what I think? conventional commits wants.  @bcoe should peek too.

bcoe:
editing commit title party, woo :stuck_out_tongue: I'd probably do something like what I just changed it to ... it's a refactor, removing something deprecated, which breaks things.

PR #625, opened 2019-05-15
refactor(deprecation)!: remove legacy support for array arguments

schmidt-sebastian:
BREAKING CHANGE: This removes the ability to call getAll() and `new FieldPath()` with an array as its first argument instead of calling it with a list of arguments.

### DIFF
--- a/dev/src/path.ts
+++ b/dev/src/path.ts
@@ -522,20 +522,23 @@ export class FieldPath extends Path<FieldPath> {
    * });
    */
   constructor(...segments: string[]) {
-    const elements: string[] = Array.isArray(segments[0])
-      ? ((segments[0] as unknown) as string[])
-      : segments;
+    if (Array.isArray(segments[0])) {
+      throw new Error(
+        'The FieldPath constructor no longer supports an array as its first argument. ' +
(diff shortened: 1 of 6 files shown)


dev/src/path.ts
```diff
@@ -522,20 +522,23 @@ export class FieldPath extends Path<FieldPath> {
    * });
    */
   constructor(...segments: string[]) {
-    const elements: string[] = Array.isArray(segments[0])
-      ? ((segments[0] as unknown) as string[])
-      : segments;
+    if (Array.isArray(segments[0])) {
```
mikelehen: Remove `string[]` from the param type definition above?

dev/src/transaction.ts
```diff
@@ -433,30 +433,23 @@ export function parseGetAllArguments(
   let documents: DocumentReference[];
```
mikelehen: I can't comment on unchanged code, but does the getAll() parameter definition need to be updated somehow?
```
   * @param {Array.<DocumentReference|ReadOptions>} documentRefsOrReadOptions
   * The `DocumentReferences` to receive, followed by an optional field
   * mask.
```

PR #628, opened 2019-05-15
docs: Remove @private from public classes

schmidt-sebastian:
The new JSDoc now hides classes with `@private` constructors. This PR removes these annotations, which restores the previous behavior (the constructors were always visible in the doc).


jkwlui:
should we get rid of `@hideconstructor` if the intended behaviour is to make constructors visible?

schmidt-sebastian:
I don't want these constructors to be visible, but I don't think there is an option to hide them. `hideconstructor` used to work at some point.

schmidt-sebastian:
> I don't want these constructors to be visible, but I don't think there is an option to hide them. `hideconstructor` used to work at some point.

Actually, it seems to work right now.

jkwlui:
So `@private` hides the entire class after `jsdoc@v3.6.2`, but we can actually do away with just `@hideconstructor`?

JustinBeckwith:
@schmidt-sebastian @jkwlui 👋 Whatup with this one?

schmidt-sebastian:
Hey :) This will allow us to unpin the JSDoc version and update to the latest version. JSDoc 3.5.1 hides the class if the constructor is marked `@private`.

manwithsteelnerves:
@schmidt-sebastian Any possibility for making Transaction constructor public?
I'm trying to implement a common library where I'm using databases additional to firestore and many follow below structure.
```
Begin
....updates
Commit 
```

Making transaction private makes the code blocks us to follow the standard transaction workflow. Please suggest.

schmidt-sebastian:
The Transaction class isn't really usable on its own, and making it appear so might confuse some end users. If a user sees that `Transaction` is easily constructible, they might prefer calling the Transaction constructor rather than `runTransaction`. If you want to make the class easier to construct on make the `requestTag` and optional argument, then I would be more than willing to accept your PR. You would still have to cast to `any` to instantiate the class though.

manwithsteelnerves:
@schmidt-sebastian  Thanks for the inputs.
Transaction currently takes two arguments
1. Firestore instance
2. RequestTag

I see requestTag can be optional as it can be generated from util class if not specified. I would prefer it to be completely removed as already internally its accessing the requestId method from util file when using runTransaction.

Regarding first variable, firestore instance, I think it will be required. Please correct me if i'm wrong.

Additionally, need to add a runTransaction which takes optional argument - maxAttempts so that we can make use of the code from existing runTransaction method by refactoring.

Incase if you can confirm the above approach is fine, I can work on it.

Separating transaction instance has the following benefits
1. It can be easily integrated with existing orm libraries easily which helps for more reachability
2. It avoids passing methods for a cleaner code. Just => Create, call transaction CRUD operations, Run.

schmidt-sebastian:
Your suggestions sound reasonable, as long we intend to expose them without changes to our public API.

PR #629, opened 2019-05-16
refactor!: Removing default export

schmidt-sebastian:
This PR removes the default export hack that we introduced in order to migrate to Typescript without a major version bump.


schmidt-sebastian:
After offline discussion with @JustinBeckwith, we decided to not proceed with this change. This change will break far too many users and the perceived benefit is low.

PR #631, opened 2019-05-17
build: add new kokoro config for coverage and release-please

bcoe:


.nycrc
```diff
@@ -10,7 +10,6 @@
     "**/samples",
     "**/scripts",
     "**/src/**/v*/**/*.js",
-    "**/protos",
```
schmidt-sebastian: :( This brings down our test coverage from 96% to 60%ish. @bcoe - Can you point me at the source of this change so that I can update the configuration?

PR #642, opened 2019-05-19
refactor: modernize sample test

JustinBeckwith:


samples/package.json
```diff
@@ -9,13 +8,13 @@
     "node": ">=8"
```
schmidt-sebastian: Should we update the engines field here as well?

bcoe:
while you're at it, if you toss `nyc` in front of the `sample-test` stanza in package.json, hour coverage should go up.

JustinBeckwith:
Any reason not to use c8 in this case?

bcoe:
@JustinBeckwith no, I'd be excited to continue rolling it out; it will be slightly awkward as we phase out `.nycrc` for `.c8rc`; I was thinking maybe I should make `c8` load `.nycrc`.

PR #649, opened 2019-05-21
deps: Remove dependency on lodash.merge

schmidt-sebastian:
See https://github.com/googleapis/nodejs-firestore/pull/648


schmidt-sebastian:
We also need to re-generate the v1beta1 client.

PR #650, opened 2019-05-21
chore: Include v1beta1 in synth.py

schmidt-sebastian:


.kokoro/samples-test.sh
```diff
@@ -31,27 +31,12 @@ if [ -f .kokoro/pre-samples-test.sh ]; then
...
+npm link ../
+npm install
+cd ..
 
-    npm run samples-test
-fi
-
-# codecov combines coverage across integration and unit tests. Include
-# the logic below for any environment you wish to collect coverage for:
-COVERAGE_NODE=10
```
bcoe: this is odd, it seems to revert a bunch of configuration, which is landed upstream; I wonder if perhaps you're running an older version of `synthtool` -- mind running a local version checked out from master?

schmidt-sebastian: Re-running from Master removed all these changes.

bcoe:
@schmidt-sebastian this looks good to me :+1: I might be tempted to call it a `feat:` rather than a chore, at the end of the day, once this is released, our customers have access to the new `v1beta1` API surface, am I correct?

schmidt-sebastian:
@bcoe The API and functionality should not have changed (ignoring the addition of a second export that could theoretically be used). I see a bunch of code that got moved around in v1beta1, but I don't see any breaking changes or feature additions. Let me know if that lowers the bar to "chore", otherwise I can make this a feature PR.

PR #651, opened 2019-05-22
feat: Support listDocuments()/listCollections() via the Firestore Emulator

ryanpbrewster:
These custom headers can be passed in via an `customHeaders` field in the constructor, and are plumbed all the way through to the gRPC metadata in `CallOptions`.

With this change it is much simpler to connect to the Firestore emulator:
```
const grpc = require("@grpc/grpc-js");
const Firestore = require("@google-cloud/firestore");
const firestore = new Firestore({
  projectId: 'firestore-emulator-sample',
  servicePath: 'localhost',
  port: 8080,
  sslCreds: grpc.credentials.createInsecure(),
  customHeaders: {
    "Authorization": "Bearer owner"
  }
});
```

I thought this was the right level of abstraction: not too general (only constant headers provided at instantiation time) but not too specific (allows arbitrary headers, rather than only the Authorization header, or some special purpose emulator field).


dev/src/index.ts
```diff
@@ -1038,9 +1038,12 @@ export class Firestore {
...
   private createCallOptions(): CallOptions {
-    const gaxHeaders: CallOptions = {otherArgs: {headers: {}}};
-    gaxHeaders.otherArgs!.headers[CLOUD_RESOURCE_HEADER] = this.formattedName;
-    return gaxHeaders;
+    const headers: any = {};
+    headers[CLOUD_RESOURCE_HEADER] = this.formattedName;
+    if (this._settings.additionalHeaders) {
+      Object.assign(headers, this._settings.additionalHeaders);
+    }
+    return {otherArgs: { headers }};
```
schmidt-sebastian: This is pretty much what I had stashed away in my local client as well.

Can we make this a little more ES-6y by changing this to:

```
 return {
      otherArgs: {
        headers: {
          [CLOUD_RESOURCE_HEADER]: this.formattedName,
          ...this._settings.customHeaders
        }
      }
    };
```

This will remove the usage of `any`. Note that I also changed the  `additionalHeaders` name to `customHeaders`, which is just a slight preference.

ryanpbrewster: holy wow, you're a wizard

schmidt-sebastian:
@ryanpbrewster Can you run `npm run fix` from the repo root?

ryanpbrewster:
done

nfarina:
Very glad to have this! All my unit tests run against the Firestore emulator and after each one I do a Jest snapshot of the entire DB. Previously I had to "discover" known collections by parsing my rules file 😆

PR #655, opened 2019-05-23
fix: retry on abort and limit retry count to 10

thebrianchen:
Fixes https://github.com/firebase/firebase-admin-node/issues/478

### DIFF
--- a/dev/src/watch.ts
+++ b/dev/src/watch.ts
@@ -431,6 +431,7 @@ abstract class Watch {
         .backoffAndWait()
         .catch(err => {
           closeStream(err);
+          return Promise.reject(err);
         })
         .then(async () => {
           if (!isActive) {
--- a/dev/src/watch.ts
+++ b/dev/src/watch.ts
@@ -446,7 +446,7 @@ abstract class Watch {
 
           // Note that we need to call the internal _listen API to pass additional
           // header values in readWriteStream.
-          this._firestore
+          return this._firestore
             .readWriteStream('listen', request, this._requestTag, true)
             .then(backendStream => {
               if (!isActive) {
@@ -470,8 +470,7 @@ abstract class Watch {
               });
               currentStream!.pipe(stream);
               currentStream!.resume();
-            })
-            .catch(closeStream);
+            });
         })
         .catch(closeStream);
     };
--- a/dev/src/watch.ts
+++ b/dev/src/watch.ts
@@ -41,12 +41,6 @@ import api = google.firestore.v1;
 const WATCH_TARGET_ID = 0x1;
 
 /*!
- * The default maximum number of retries that will be attempted by backoff
- * before stopping all retry attempts.
- */
-export const DEFAULT_MAX_RETRY_ATTEMPTS = 10;
-
-/*!
  * The change type for document change events.
  */
 // tslint:disable-next-line:variable-name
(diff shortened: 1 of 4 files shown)


dev/src/watch.ts
```diff
@@ -426,6 +433,15 @@ abstract class Watch {
      * Initializes a new stream to the backend with backoff.
      */
     const initStream = () => {
+      if (this._backoff.getRetryCount() > DEFAULT_MAX_RETRY_ATTEMPTS) {
+        closeStream(
+          new Error(
+            'Exceeded maximum number of retries before any ' +
+              'response was received.'
```
schmidt-sebastian: This could be one line. We don't really have a fixed line length and 80 character wrap is so 1994.

dev/src/watch.ts
```diff
@@ -426,6 +433,15 @@ abstract class Watch {
...
+          )
+        );
+        return;
+      }
       this._backoff.backoffAndWait().then(async () => {
```
schmidt-sebastian: What do you think about encapsulating `retryCount` in `backoffAndWait()` and failing the Promise that `backoffAndWait()` returns?

dev/test/watch.ts
```diff
@@ -819,6 +824,36 @@ describe('Query watch', () => {
...
 
+  it('stops attempts after maximum retry attempts', () => {
+    const err = new GrpcError('GRPC Error');
+    err.code = Number(10 /* ABORTED */);
+    return watchHelper.runFailedTest(
+      collQueryJSON(),
+      async () => {
+        let chain = Promise.resolve();
+        // Retry for the maximum of retry attempts.
+        for (let i = 0; i < DEFAULT_MAX_RETRY_ATTEMPTS; i++) {
```
schmidt-sebastian: You can just use `await` in the for  loop, which should allow you to get rid of the Promise chain.

dev/src/watch.ts
```diff
@@ -40,6 +40,12 @@ import api = google.firestore.v1;
  */
 const WATCH_TARGET_ID = 0x1;
 
+/*!
+ * The default maximum number of retries that will be attempted by backoff
+ * before stopping all retry attempts.
+ */
+export const DEFAULT_MAX_RETRY_ATTEMPTS = 10;
```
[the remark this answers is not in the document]
thebrianchen: Ohh didn't know that!

thebrianchen:
Thanks for the review mrschmidt!

dev/src/watch.ts
```diff
@@ -426,50 +427,55 @@ abstract class Watch {
...
-            'Watch.onSnapshot',
-            this._requestTag,
-            'Not initializing inactive stream'
-          );
-          return;
-        }
+      this._backoff
+        .backoffAndWait()
+        .catch(err => {
+          closeStream(err);
```
schmidt-sebastian: After you deal with a Promise in `.catch()` it is no longer considered rejected. While an error in the backoff timer is handled by closing the stream, the code as written then immediately tries to re-establish it. This doesn't happen here because `closeStream()` sets `isActive` to false, but I would prefer if we did not rely on this.

thebrianchen: By this, you mean to pass along the Promise rejection right?

schmidt-sebastian: If you just revert your change and call `.then` before `.catch`, everything will work again. Otherwise you have to do:

```
.catch(e => {
   ...
   return Promise.reject(e);
}
```

schmidt-sebastian:
I pushed a small cleanup to this branch to remove the duplicate .catch() clause. Let me know if you disagree otherwise I will merge this tomorrow. Thanks!

PR #658, opened 2019-05-24
fix: Allow CollectionReference.doc(undefined)

Arthie:
This PR Fixes Error below when passing `undefined` as parameter to `CollectionReference.doc()`

`CollectionReference.doc(undefined)` gives this error:
`'Value for argument "documentPath" is not a valid ' + 'resource path. Path must be a non-empty string.'`

undefined should be a valid parameter for `CollectionReference.doc()` according to the types because the **documentPath** argument is an optional string === `string | undefined`.

- [X] Tests and linter pass

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2073,7 +2073,7 @@ export class CollectionReference extends Query {
    * console.log(`Reference with auto-id: ${documentRefWithAutoId.path}`);
    */
   doc(documentPath?: string): DocumentReference {
-    if (arguments.length === 0) {
+    if (documentPath === undefined) {
       documentPath = autoId();
     } else {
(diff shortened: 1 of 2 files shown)


Arthie:
I signed it!

schmidt-sebastian:
> Thanks for this fix! Our CI is a bit behind these days, but this should get merged and released soon.
> 
> BTW: This issue also exists in the Web SDK. I filed an internal issue to clean this up.

I have some bad news :/ After discussing this with the team, we decided that the original invariant of this code was worth keeping. Namely, we want to prevent our users from accidentally passing undefined. With that said, our TypeScript types do not reflect this. I am proposing a fix for this here: https://github.com/googleapis/nodejs-firestore/pull/658

PR #660, opened 2019-05-28
docs: Update docs in CONTRIBUTING and test commands

thebrianchen:


.kokoro/samples-test.sh
```diff
@@ -40,7 +40,7 @@ if [ -f samples/package.json ]; then
     npm install
     cd ..
 
-    npm run samples-test
+    npm run test:samples
```
JustinBeckwith: These files are auto-generated, and we put a *huge* premium on consistency here.

schmidt-sebastian: We tried to make this SDK more consistent with the test patterns we use in `firebase-js-sdk`. That being said, breaking (or worse: fixing) 72 repos was not our goal :) 

If I am not mistaken, then this repo is the only repo that uses the conformance tests. We should be able to achieve some of the goals of this PR (have more consistency) by just updating the conformance test runner to 'conformance-test' or by just inlining it with the unit tests.

schmidt-sebastian:
Further things we should mention:
- Docs generation and docs testing
- Running the system tests and how to set up the environment variable

schmidt-sebastian:
@thebrianchen Should we close this for now? https://github.com/googleapis/nodejs-firestore/pull/780 merges your latest changes.

PR #662, opened 2019-05-28
fix: Add overloads for doc() and doc(id:string)

schmidt-sebastian:
This addresses the same issue as https://github.com/googleapis/nodejs-firestore/pull/658, but instead of relaxing our `CollectionReference.doc()` API, it fixes the types to reflect the invariant that we enforce via our argument validation. As in older releases, we still prevent users from passing "undefined", but with this PR this should generate a compile time error.

This is based on the following little experiment:

```
interface Foo {
    bar(): void;
    bar(baz: string): void;
}
class Foobar implements Foo {
    bar(baz?: string) {}
}
function createFoo(): Foo {
    return new Foobar();
}
const f = createFoo();
f.bar();
f.bar("foobar");
f.bar(undefined); // Compile error
```

Note that I am marking this as a "fix" rather than as a "breaking change" as any user that tried to use "undefined" before would have been broken already (and gotten a validation error). This PR will therefore go out with the next release, which is a minor version bump.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2055,6 +2055,8 @@ export class CollectionReference extends Query {
     });
   }
 
+  doc(): DocumentReference;
+  doc(documentPath: string): DocumentReference;
   /**
    * Gets a [DocumentReference]{@link DocumentReference} instance that
    * refers to the document at the specified path. If no path is specified, an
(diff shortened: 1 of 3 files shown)


dev/src/reference.ts
```diff
@@ -2055,6 +2055,8 @@ export class CollectionReference extends Query {
     });
   }
 
+  doc(): DocumentReference;
+  doc(documentPath: string): DocumentReference;
```
schmidt-sebastian: This weird arrangement here is so that our JSDoc generation still works. The TypeScript compiler still validates that we do not pass "undefined" (hence the test change).

PR #672, opened 2019-06-03
docs: Adding example for FieldValue.isEqual

schmidt-sebastian:
Fixes b/129393978

### DIFF
--- a/dev/src/field-value.ts
+++ b/dev/src/field-value.ts
@@ -171,6 +171,28 @@ export class FieldValue {
    *
    * @param {*} other The value to compare against.
    * @return {boolean} true if this `FieldValue` is equal to the provided value.
+   *
+   * @example
+   * let fieldValues = [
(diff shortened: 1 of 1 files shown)


dev/src/field-value.ts
```diff
@@ -171,6 +171,28 @@ export class FieldValue {
    *
    * @param {*} other The value to compare against.
    * @return {boolean} true if this `FieldValue` is equal to the provided value.
+   *
+   * @example
+   * let fieldValues = [
```
mikelehen: const?

mikelehen: (very optional)

schmidt-sebastian: For better or for worse, we don't use `const` in any of the other samples. Leaving as is.

schmidt-sebastian: For better or for worse, we don't use `const` in any of the other samples. Leaving as is.

Issue #674, opened 2019-06-04
Add a settings flag for creating insecure credentials


samtstern:
**Is your feature request related to a problem? Please describe.**

When using this library to connect to the Cloud Firestore emulator, you have to change the firestore settings like this:

```js
firestore.settings({
        projectId: 'samstern-projectdoesnotexist',
        port: 8080,
        servicePath: 'localhost',
        sslCreds: grpc.credentials.createInsecure(),
    });
```

In order to provide `sslCreds` I have to import `grpc` and make sure that I resolve to the same version of grpc as the one used by this library.  This is fairly inconvenient.

For instance with the change from `grpc` to `grpc-js` recently I had to change my testing code, but the change was not obvious since I was depending on an implementation detail of this library.

 **Describe the solution you'd like**

I would prefer that this library provide a new flag in `firestore.settings` like `insecureCredentials: true` or `sslEnabled: false` that would do the same thing under the hood.

Firestore already has this on other platforms.  For instance on Android:

```java
  FirebaseFirestoreSettings settings =
        new FirebaseFirestoreSettings.Builder()
                .setHost("10.0.2.2:8080")
                .setSslEnabled(false)
                .build();
```

 **Describe alternatives you've considered**

It's possible that we could infer this from `servicePath: localhost` but I don't need that much magic.

 **Additional context**

The Functions emulator built into the Firebase CLI automatically points the Firestore SDK at the emulator, and we have run into a bunch of issues related to needing to resolve `grpc` so this would help us a lot.

samtstern:
cc @abeisgoat @ryanpbrewster

schmidt-sebastian:
@samtstern Do we still need to do this if we take care of https://github.com/firebase/firebase-admin-node/issues/472?

samtstern:
@schmidt-sebastian definitely less urgent if that lands, but I think environment variables are pretty messy and would prefer to have something like this also.

PR #680, opened 2019-06-05
feat: Support host, ssl and FIRESTORE_EMULATOR_HOST for endpoint configuration

schmidt-sebastian:
This PR ports the "host" and "ssl" settings from the Web SDK and adds support for the FIRESTORE_EMULATOR_HOST that the emulator sets at startup.

FIRESTORE_EMULATOR_HOST is already supported by other Server SDKs, such as Go.

Fixes: https://github.com/googleapis/nodejs-firestore/issues/674
Fixes: https://github.com/firebase/firebase-admin-node/issues/472

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -449,7 +449,7 @@ export class Firestore {
       validateHost('settings.host', settings.host);
       if (settings.servicePath !== undefined) {
         throw new Error(
-          'Cannot set both "settings.hos" and "settings.servicePath".'
+          'Cannot set both "settings.host" and "settings.servicePath".'
         );
       }
       if (settings.apiEndpoint !== undefined) {
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -863,8 +863,8 @@ export class Firestore {
     requestTag: string,
     transactionId?: Uint8Array
   ): Promise<DocumentSnapshot[]> {
-    const requestedDocuments = new Set();
-    const retrievedDocuments = new Map();
+    const requestedDocuments = new Set<string>();
+    const retrievedDocuments = new Map<string, DocumentSnapshot>();
 
     for (const docRef of docRefs) {
       requestedDocuments.add(docRef.formattedName);
@@ -947,12 +947,13 @@ export class Firestore {
               const orderedDocuments: DocumentSnapshot[] = [];
               for (const docRef of docRefs) {
                 const document = retrievedDocuments.get(docRef.path);
-                if (document === undefined) {
+                if (document !== undefined) {
+                  orderedDocuments.push(document);
+                } else {
                   reject(
                     new Error(`Did not receive document for "${docRef.path}".`)
                   );
                 }
-                orderedDocuments.push(document);
               }
               resolve(orderedDocuments);
             });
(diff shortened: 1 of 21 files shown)


dev/src/index.ts
```diff
@@ -407,11 +440,30 @@ export class Firestore {
...
+    if (settings.projectId !== undefined) {
       validateString('settings.projectId', settings.projectId);
       this._projectId = settings.projectId;
     }
 
+    if (settings.host !== undefined) {
+      validateHost('settings.host', settings.host);
+      if (settings.servicePath !== undefined) {
+        throw new Error(
+          'Cannot set both "settings.hos" and "settings.servicePath".'
```
thebrianchen: settings.host

dev/src/validate.ts
```diff
@@ -151,6 +152,40 @@ export function validateString(
...
 }
 
+/**
+ * Validates that 'value' is a host.
+ *
+ * @private
+ * @param arg The argument name or argument index (for varargs methods).
+ * @param value The input to validate.
+ * @param options Options that specify whether the host can be omitted.
+ * whether it can contain a port.
```
thebrianchen: sentence fragment: is "whether it can contain a port" supposed to be with options?

dev/src/index.ts
schmidt-sebastian: Done

dev/src/validate.ts
schmidt-sebastian: Done

dev/src/index.ts
```diff
@@ -338,7 +340,29 @@ export class Firestore {
...
 
-    this.validateAndApplySettings(Object.assign({}, settings, libraryHeader));
+    if (process.env.FIRESTORE_EMULATOR_HOST) {
+      validateHost(
+        'FIRESTORE_EMULATOR_HOST',
+        process.env.FIRESTORE_EMULATOR_HOST
+      );
+
+      const url = new URL('http://' + process.env.FIRESTORE_EMULATOR_HOST);
+      const host = url.hostname;
```
JustinBeckwith: In every other library, we're trying to consistently support the `apiEndpoint` property in the options that are passed into the client constructor.  Can we support this as well here as well as `host`?

schmidt-sebastian: If `host` is not specified, `apiEndpoint` and `servicePath` are passed through and work as they do in other languages. If `host` and either `apiEndpoint` or `servicePath` is specified, we fail initialization.

samtstern:
I think we use a different name for the environment variable:
https://github.com/firebase/firebase-tools/blob/bf1c1dedabe5e64c85147514705bd59c36501b6e/src/emulator/firestoreEmulator.ts#L14

cc @ryanpbrewster  and @yuchenshi

schmidt-sebastian:
> I think we use a different name for the environment variable:
> https://github.com/firebase/firebase-tools/blob/bf1c1dedabe5e64c85147514705bd59c36501b6e/src/emulator/firestoreEmulator.ts#L14
> 
> cc @ryanpbrewster and @yuchenshi

Hm. The debug log from the emulator says it's FIRESTORE_EMULATOR_HOST:

```
 If you are using a library that supports the FIRESTORE_EMULATOR_HOST environment variable, run:

   export FIRESTORE_EMULATOR_HOST=[::1]:8080,127.0.0.1:8080
```

PR #683, opened 2019-06-05
fix: Store large numbers as doubles

thebrianchen:
Fixes firebase/firebase-admin-node#548

### DIFF
--- a/dev/src/serializer.ts
+++ b/dev/src/serializer.ts
@@ -24,7 +24,7 @@ import {DocumentReference, Firestore} from './index';
 import {FieldPath, QualifiedResourcePath} from './path';
 import {Timestamp} from './timestamp';
 import {ApiMapValue, DocumentData, ValidationOptions} from './types';
-import {isEmpty, isObject, isSafeInteger} from './util';
+import {isEmpty, isObject} from './util';
 import {customObjectMessage, invalidArgumentMessage} from './validate';
 
 import api = proto.google.firestore.v1;
@@ -117,7 +117,7 @@ export class Serializer {
     }
 
     if (typeof val === 'number') {
-      if (isSafeInteger(val)) {
+      if (Number.isSafeInteger(val)) {
         return {
           integerValue: val as number,
         };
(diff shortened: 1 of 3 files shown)


dev/test/document.ts
```diff
@@ -170,6 +170,28 @@ describe('serialize document', () => {
...
   });
 
+  it('serializes large numbers into doubles', () => {
+    const overrides: ApiOverride = {
+      commit: (request, options, callback) => {
+        requestEquals(
+          request,
+          set({
+            document: document('documentId', 'someNumber', {
+              doubleValue: 18014398509481984,
```
schmidt-sebastian: Can you add a comment to explain what makes this value special? Alternatively, you could also use a more descriptive name than "someNumber".

dev/src/serializer.ts
```diff
@@ -116,17 +116,16 @@ export class Serializer {
...
-        integerValue: val as number,
-      };
-    }
-
-    // Integers are handled above, the remaining numbers are treated as doubles
     if (typeof val === 'number') {
-      return {
-        doubleValue: val as number,
-      };
+      if (isSafeInteger(val)) {
```
schmidt-sebastian: You can directly use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger since we are using ES2016: https://github.com/google/gts/blob/master/tsconfig-google.json#L7

dev/src/util.ts
```diff
@@ -14,6 +14,24 @@
...
+// Untyped Number alias we can use to check for ES6 methods / properties.
+// tslint:disable-next-line:no-any variable-name
+const NumberAsAny = Number as any;
+
+/**
+ * Minimum safe integer in Javascript because of floating point precision.
+ * Added to not rely on ES6 features.
+ */
+export let MIN_SAFE_INTEGER: number =
+  NumberAsAny.MIN_SAFE_INTEGER || -(Math.pow(2, 53) - 1);
```
schmidt-sebastian: Feel free to ignore since I think you can just remove the changes in this file: You shouldn't have to use `NumberAsAny`. I am surprised `Number. MIN_SAFE_INTEGER` doesn't work for you.

PR #684, opened 2019-06-06
fix: Split settings.host into servicePath and port

schmidt-sebastian:
I think https://github.com/googleapis/nodejs-firestore/pull/680 doesn't split the `settings.host` correctly into `port` and `servicePath`. We only do this for FIRESTORE_EMULATOR_HOST.


schmidt-sebastian:
cc @thebrianchen

PR #696, opened 2019-06-19
fix: Unset servicePath when FIRESTORE_EMULATOR_HOST is set

schmidt-sebastian:
We currently assert that only one of servicePath and host (or apiEndpoint and host) is set. This causes a problem when a user or a library consumer specifies a servicePath and sets the FIRESTORE_EMULATOR_HOST variable. Since we want to "blindly" accept FIRESTORE_EMULATOR_HOST as an override, we should unset servicePath and apiEndpoint.

See https://github.com/firebase/firebase-tools/blob/589387b79a9c07f4811ab35201ced1bb84e66e72/src/emulator/functionsEmulatorRuntime.ts#L425 for code that sets servicePath and https://github.com/firebase/firebase-admin-node/issues/575 for the user issue.


schmidt-sebastian:
@ryanpbrewster Can you sanity-check this?

dev/src/reference.ts
```diff
@@ -1048,7 +1048,9 @@ export class Query {
         const fieldValue = documentSnapshot.get(fieldOrder.field);
         if (fieldValue === undefined) {
           throw new Error(
-            `Field "${fieldOrder.field}" is missing in the provided DocumentSnapshot. ` +
+            `Field "${
+              fieldOrder.field
+            }" is missing in the provided DocumentSnapshot. ` +
```
ryanpbrewster: unrelated?

schmidt-sebastian: Yes. I guess prettier got updated.

schmidt-sebastian: Had to reinstall prettier but it's gone now.

ChromeQ:
I am still experiencing this error after updating to the latest version and after a bit of debugging I found the reason.
So the changes you made remove the `servicePath` and `apiEndpoint` when `process.env.FIRESTORE_EMULATOR_HOST` is set: https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L360
This prevents the errors being thrown inside `validateAndApplySettings` but on L470 it sets it again and then saves all settings to the instance `this._settings` (L480)

This means that anytime `this.settings` are later used or merged it contains the servicePath, and any calls to `validateAndApplySettings` throws the errors again.
I am hitting this issue inside firebase-tools functions emulator: https://github.com/firebase/firebase-tools/blob/master/src/emulator/functionsEmulatorRuntime.ts#L490 which calls settings()

Although it does not pass any userSettings in my case but it might here https://github.com/firebase/firebase-tools/blob/master/src/emulator/functionsEmulatorRuntime.ts#L505 and finally the `this._settings` are merged which contains `servicePath`
https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L438

Hope this helps and is easy enough to follow?

schmidt-sebastian:
@ChromeQ Sorry that I seem to be solving your issues one baby step at a time. I opened a follow up PR that hopefully unblocks you. If you have a chance, do you mind giving it a spin?

ChromeQ:
Of course, I'll see if I can patch mine with the PR changes and let you
know.

ChromeQ:
I've added in the `delete settings.host;` and there are no errors being thrown anymore, so that has worked.

Unfortunately though, none of my functions are picking up the emulator running firestore. See my steps below...

In one process I run `firebase emulators:start --only firestore`
This is a success and outputs `FIRESTORE_EMULATOR_HOST=localhost:5002`
In another separate process I run `FIRESTORE_EMULATOR_HOST=localhost:5002 firebase serve --only functions`
There are no more errors with your patch changes but I am warned: `Ignoring trigger "onCompetitionCreate" because the Cloud Firestore emulator is not running.`

However it does seem to work without errors or warnings when I run `firebase emulators:start --only firestore,functions` so I will be using that going forward. :tada: :+1: 

It seems as though we cannot mix emulators:start and serve, which is probably fine.
Interestingly when I use `firebase serve --only firestore,functions` it breaks as port 8080 is already used, it isn't occupied when I start this command so something may be initialising twice

Thanks for persisting and being so quick on this issue @schmidt-sebastian :smiley:

ryanpbrewster:
You got it. The Firestore and Functions emulators will only talk to each other if they're started at the same time from the same command.

PR #697, opened 2019-06-20
fix: Support non-ISO-8859-1 Collection Names in Queries

schmidt-sebastian:
Updates google-gax to 1.1.2 to pull in https://github.com/googleapis/gax-nodejs/pull/521  and adds tests to verify that we can query non-Ascii collection.


schmidt-sebastian:
go/grpc-request-params should answer all your questions.

PR #700, opened 2019-06-22
fix: Use new stream rather than pipe

### LINKED ISSUE #661: GRPC-JS: Possible EventEmitter memory leak detected
See https://github.com/googleapis/nodejs-firestore/pull/623#issuecomment-496343759

> I just upgraded to @google-cloud/firestore@2.0.0 and since upgrading, I've been getting all of these warnings that say:
> (node:2870) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connectivityStateChanged listeners added. Use emitter.setMaxListeners() to increase limit

thebrianchen:
Fixes #661.

We use a single long-lived stream for all of our logic. If the backend streams drops, we unpipe the old stream and pipe and new stream. After many re-connects, it seems to cause EventEmitter leaks (#661). By using new streams rather than piping/unpiping a long-lived stream, the `MaxListenersExceededWarning` should no longer appear.

### DIFF
--- a/dev/src/watch.ts
+++ b/dev/src/watch.ts
@@ -590,7 +590,7 @@ abstract class Watch {
         this.backoff.reset();
       }
     } else if (proto.documentChange) {
-      logger('Watch.onSnapshot', this.requestTag, 'Processing change event');
+      logger('Watch.onData', this.requestTag, 'Processing change event');
 
       // No other targetIds can show up here, but we still need to see
       // if the targetId was in the added list or removed list.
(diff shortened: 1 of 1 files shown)


dev/src/watch.ts
```diff
@@ -687,6 +540,113 @@ abstract class Watch {
...
   }
 
+  /**
+   * Handles 'data' events and closes the stream if the response type is
+   * invalid.
+   * @private
+   */
+  private onData(proto: api.IListenResponse): void {
+    if (proto.targetChange) {
+      logger('Watch.onSnapshot', this.requestTag, 'Processing target change');
```
schmidt-sebastian: Please update the log function names.

dev/src/watch.ts
```diff
@@ -687,6 +540,113 @@ abstract class Watch {
...
+          this.closeStream(Error('Unexpected target ID sent by server'));
+        }
+      } else if (change.targetChangeType === 'REMOVE') {
+        let code = 13;
+        let message = 'internal error';
+        if (change.cause) {
+          code = change.cause.code!;
+          message = change.cause.message!;
+        }
+        // @todo: Surface a .code property on the exception.
```
schmidt-sebastian: I think we may have now earned a Moma badge by moving this TODO around more than 10 times :)

PR #717, opened 2019-07-22
fix(deps): drop dependency on through2

JustinBeckwith:


schmidt-sebastian:
Do the comments from https://github.com/googleapis/nodejs-firestore/pull/618#event-2337855128 still apply here?

JustinBeckwith:
🤦‍♂

PR #729, opened 2019-08-08
internal: Exclude conformance tests from code coverage

schmidt-sebastian:


schmidt-sebastian:
@bcoe: If I am not mistaken, conventional commits using the commit message from the merge commit, which is the PR title. Is that correct? It's a bit of a pain to update a PR that from an edit on GitHub, so I am inclined to leave the original commit message as is.

JustinBeckwith:
This is the file we need to change:
https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/.nycrc

After that's in, we need to do a hammer PR to make the change (happy to do that).  

Also - for the conventional commits bit I think the `chore:` prefix or `test:` prefix would be the right answer :)

PR #733, opened 2019-08-16
fix: add logging to client pool

schmidt-sebastian:
This PR logs when we create or delete instances in the client pool, helping us debug https://github.com/firebase/firebase-admin-node/issues/620

Almost everything here is related to plumbing through the "request tag"

### DIFF
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -161,6 +161,7 @@ export class ClientPool<T> {
    * Deletes clients that are no longer executing operations. Keeps up to one
    * idle client to reduce future initialization costs.
    *
+   * @return Number of clients deleted.
    * @private
    */
   private garbageCollect(): number {
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -16,6 +16,8 @@
 
 import * as assert from 'assert';
 
+import {logger} from './logger';
+
 /**
  * An auto-resizing pool that distributes concurrent operations over multiple
  * clients of type `T`.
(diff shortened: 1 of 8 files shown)


dev/src/pool.ts
```diff
@@ -156,5 +174,6 @@ export class ClientPool<T> {
         }
       }
     });
+    return idleClients - 1;
```
hiranya911: Is this supposed to represent the number of clients deleted? Can you add a comment to clarify that?

dev/src/pool.ts
```diff
@@ -79,15 +88,23 @@ export class ClientPool<T> {
...
+  private release(requestTag: string, client: T): void {
     let requestCount = this.activeClients.get(client) || 0;
     assert(requestCount > 0, 'No active request');
 
     requestCount = requestCount! - 1;
     this.activeClients.set(client, requestCount);
 
     if (requestCount === 0) {
-      this.garbageCollect();
+      const deletedCount = this.garbageCollect();
```
hiranya911: Should we log something for deletedCount == 0 case too?

dev/src/write-batch.ts
```diff
@@ -527,11 +527,11 @@ export class WriteBatch {
     // Note: We don't call `verifyNotCommitted()` to allow for retries.
     this._committed = true;
 
-    await this._firestore.initializeIfNeeded();
-
+    const tag = (commitOptions && commitOptions.requestTag) || requestTag();
     const explicitTransaction = commitOptions && commitOptions.transactionId;
```
hiranya911: Might want to push this a couple of lines down, until it's actually needed.

PR #735, opened 2019-08-21
fix: remove hasOwnProperty method

### LINKED ISSUE #732: Serializer does not check object for a prototype before calling hasOwnProperty()
Latest version.

https://github.com/googleapis/nodejs-firestore/search?q=hasOwnProperty&unscoped_q=hasOwnProperty

The bigger issue is that if I try to save an object into the firestore was created with `Object.create(null)`, the save fails due to the object not having hasOwnProperty. It is quite difficult to debug as a result

Related reasoning and background and example code: 

* https://github.com/graphql/express-graphql/issues/177
* https://github.com/graphql/graphql-js/pull/504
* https://github.com/apollographql/apollo-server/issues/3149

praveenqlogic01:
Fixes #732  (it's a good idea to open an issue first for discussion)

removed use of hasOwnProperty function in application.

created two generic function to list out all own properties from object and to find out value of given property in object.

export const getPropValue = <T, K extends keyof T>(obj: T, key: K) => {
return obj[key];
};

export const getObjectOwnProperties = (obj: T) => {
if (obj) {
return Object.keys(obj);
}
return [];
};

### DIFF
--- a/dev/conformance/runner.ts
+++ b/dev/conformance/runner.ts
@@ -29,6 +29,8 @@ import {
   FieldPath,
   FieldValue,
   Firestore,
+  getObjectOwnProperties,
+  getPropValue,
   Query,
   QueryDocumentSnapshot,
   QuerySnapshot,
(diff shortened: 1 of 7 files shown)


thebrianchen:
Merged into https://github.com/googleapis/nodejs-firestore/pull/736

PR #736, opened 2019-08-22
fix: serialization does not work with null objects

### LINKED ISSUE #732: Serializer does not check object for a prototype before calling hasOwnProperty()

thebrianchen:
Fixes #732 

Thanks to @praveenqlogic01 for spearheading the fix!

### DIFF
--- a/dev/conformance/runner.ts
+++ b/dev/conformance/runner.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017 Google Inc. All Rights Reserved.
+ * Copyright 2019 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


lookfirst:
@thebrianchen you kick ass man. thanks so much!

PR #742, opened 2019-08-30
fix: retrying 13 INTERNAL RPC errors

schmidt-sebastian:
Retrying INTERNAL errors for Idempotent calls should help address https://github.com/firebase/firebase-functions/issues/536

### DIFF
--- a/dev/src/v1/firestore_client_config.json
+++ b/dev/src/v1/firestore_client_config.json
@@ -4,6 +4,7 @@
       "retry_codes": {
         "idempotent": [
           "DEADLINE_EXCEEDED",
+          "INTERNAL",
           "UNAVAILABLE"
         ],
         "non_idempotent": []


alexander-fenster:
@schmidt-sebastian Note that this file is auto-generated. It's OK to try this change here to see if it helps, but the ground truth is https://github.com/googleapis/googleapis/blob/82809578652607c8ee29d9e199c21f28f81a03e0/google/firestore/v1/firestore_gapic.yaml#L41-L42 (and it, in turn, gets auto-synced from `third_party` internal folder, which is the real place where this change must be done).

alexander-fenster:
(what I mean is that the change made here will disappear after `synthtool` run - I'm approving it if you'd like to test it out here fast, but we won't see it here for long :) )

schmidt-sebastian:
> @schmidt-sebastian Note that this file is auto-generated. It's OK to try this change here to see if it helps, but the ground truth is https://github.com/googleapis/googleapis/blob/82809578652607c8ee29d9e199c21f28f81a03e0/google/firestore/v1/firestore_gapic.yaml#L41-L42 (and it, in turn, gets auto-synced from `third_party` internal folder, which is the real place where this change must be done).

Which, in turn, is generated from https://cs.corp.google.com/piper///depot/google3/google/firestore/admin/v1/firestore_gapic.yaml :)
I will commit this and release this ASAP. Changing the main GAPIC file has to be coordinated with the backend, which I will take care of after.

PR #772, opened 2019-10-01
docs: update docs for system test

thebrianchen:
pls look sebastian


CONTRIBUTING.md
```diff
@@ -52,4 +52,13 @@ accept your pull requests.
 
         npm run fix
 
+1. Run backend tests:
```
schmidt-sebastian: Can you merge this with the previous test section and differentiate between "unit" and "system" tests (rather than backend).

Issue #768, opened 2019-09-29
HUGE memory leak problem


jakeleventhal:
#### Environment details

  - OS: Mac OS 10.15 Beta (19A573a)
  - Node.js version: v12.10.0
  - npm version: 6.10.3
  - `@google-cloud/firestore` version: 2.3.0

I'm still experiencing a memory leak potentially related to #661. The following is from my yarn.lock:
```
"@grpc/grpc-js@^0.5.2":
  version "0.5.2"
  resolved "https://registry.yarnpkg.com/@grpc/grpc-js/-/grpc-js-0.5.2.tgz#8e39876033749e1611fe9fa181d9535077805abe"
  integrity sha512-NE1tP/1AF6BqhLdILElnF7aOBfoky+4ZOdZU/0NmKo2d+9F9QD8zGoElpBk/5BfyQZ3u1Zs+wFbDOFpVUzDx1w==
  dependencies:
    semver "^6.0.0"

"@grpc/proto-loader@^0.5.0", "@grpc/proto-loader@^0.5.1":
  version "0.5.1"
  resolved "https://registry.yarnpkg.com/@grpc/proto-loader/-/proto-loader-0.5.1.tgz#48492b53cdda353110b51a4b02f465974729c76f"
  integrity sha512-3y0FhacYAwWvyXshH18eDkUI40wT/uGio7MAegzY8lO5+wVsc19+1A7T0pPptae4kl7bdITL+0cHpnAPmryBjQ==
  dependencies:
    lodash.camelcase "^4.3.0"
    protobufjs "^6.8.6"
```
_Quick note: because of the grpc problems that keep appearing here, we should probably put grpc version as part of the issue template_

So it is clear that I'm actually using the "correct" grpc. My application makes many calls to firestore and I'm experiencing massive memory leaks.
<img width="1673" alt="Screen Shot 2019-09-28 at 6 53 37 PM" src="https://user-images.githubusercontent.com/15133828/65825228-779fab00-e229-11e9-95c5-1c6b2c922ab3.png">

I'm not actually seeing the error message described previously (`MaxListenersExceededWarning`), however, the memory usage on my application slowly and steady is increasing. I've taken several heap snapshots from local tests over the course of several hours and when comparing snapshots from a few hours apart, I notice that the top culprits for memory allocations are all related to grpc.

Here you can clearly see that the top 4 culprits for the memory delta between snapshots 12 and 4 are vastly more than everything else
[...]

PR #779, opened 2019-10-08
fix: close GRPC channel when we dispose of clients

schmidt-sebastian:
See https://github.com/googleapis/nodejs-firestore/issues/768, especially https://github.com/googleapis/nodejs-firestore/issues/768#issuecomment-539654138

This is a "Request for comments" PR if this is a viable strategy to dispose of unused GRPC clients, in the hope that this will mitigate memory leaks. If feasible, some of the code in this PR needs to be pushed upstream to the Gapic generator.


schmidt-sebastian:
FYI @thebrianchen

dev/src/index.ts
```diff
@@ -401,7 +401,8 @@ export class Firestore {
 
         logger('Firestore', null, 'Initialized Firestore GAPIC Client');
         return client;
-      }
+      },
+      /* clientDestructor= */ (client: GapicClient) => client.close()
```
alexander-fenster: hmmm - do we really have `.close()`? :)

schmidt-sebastian: Not yet - but this PR shows how we could add it. We still have to add it to the Gapic generator though (if you are okay with the general direction).

schmidt-sebastian: @alexander-fenster Do you know what the next sets of steps would be if we wanted to add this support?

schmidt-sebastian:
@alexander-fenster PR rebased against the Typescript branch and updated.

PR #805, opened 2019-12-06
feat!: convert Gapic client to TypeScript

alexander-fenster:
Converting the auto-generated parts of the library to TypeScript using the new https://github.com/googleapis/gapic-generator-typescript.


dev/src/v1/firestore_client_config.json
```diff
@@ -2,42 +2,37 @@
...
+          "rpc_timeout_multiplier": 1,
+          "max_rpc_timeout_millis": 20000,
           "total_timeout_millis": 600000
         }
       },
       "methods": {
         "GetDocument": {
           "timeout_millis": 60000,
-          "retry_codes_name": "idempotent",
+          "retry_codes_name": "d_e_a_d_l_i_n_e_e_x_c_e_e_d_e_d_i_n_t_e_r_n_a_l_u_n_a_v_a_i_l_a_b_l_e",
```
alexander-fenster: This is a scary auto-generated name, looks like a bug - I'll figure out

dev/protos/firestore_proto_api.d.ts
```diff
@@ -1,17 +1,3 @@
-// Copyright 2019 Google LLC
```
alexander-fenster: Was the license added manually? I just used the update script.

dev/protos/update.sh
```diff
@@ -50,6 +50,14 @@ mkdir -p "${PROTOS_DIR}/google/firestore/v1"
 cp googleapis/google/firestore/v1/*.proto \
    "${PROTOS_DIR}/google/firestore/v1/"
 
+mkdir -p "${PROTOS_DIR}/google/firestore/v1beta1"
```
alexander-fenster: Now we need all the protos there.

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+   *
+   *   Optional. If not specified, an ID will be assigned by the service.
+   * @param {google.firestore.v1.Document} request.document
+   *   Required. The document to create. `name` must not be set.
+   * @param {google.firestore.v1.DocumentMask} request.mask
+   *   The fields to return. If not set, returns all fields.
+   *
+   *   If the document has a field that is not present in this mask, that field
+   *   will not be returned in the response.
+   * @param {object} [options]
```
schmidt-sebastian: optionsOrCallback?

Should the documentation be on the "overload" rather than the catch-all implementation?

Also... uhm... it's 2020 soon. We should drop callbacks :)

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+    options.otherArgs = options.otherArgs || {};
+    options.otherArgs.headers = options.otherArgs.headers || {};
+    options.otherArgs.headers[
+      'x-goog-request-params'
+    ] = gax.routingHeader.fromParams({
+      document_name: request.document!.name || '',
+    });
+    return this._innerApiCalls.updateDocument(request, options, callback);
+  }
+  deleteDocument(
```
schmidt-sebastian: Nit: Empty line before.

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+      callSettings
+    );
+  }
+
+  /**
+   * Terminate the GRPC channel and close the client.
+   *
+   * The client will no longer be usable and all future behavior is undefined.
+   */
+  close(): Promise<void> {
```
schmidt-sebastian: WHEEEE. I take all my comments. back. Perfection has been reached. :)

dev/src/v1/firestore_client_config.json
```diff
@@ -2,42 +2,37 @@
   "interfaces": {
     "google.firestore.v1.Firestore": {
       "retry_codes": {
+        "non_idempotent": [],
         "idempotent": [
           "DEADLINE_EXCEEDED",
-          "INTERNAL",
```
schmidt-sebastian: We need to fix this for the main client for sure.

dev/src/v1/firestore_client_config.json
```diff
@@ -2,42 +2,37 @@
...
-          "total_timeout_millis": 600000
-        },
-        "streaming": {
-          "initial_retry_delay_millis": 100,
-          "retry_delay_multiplier": 1.3,
-          "max_retry_delay_millis": 60000,
-          "initial_rpc_timeout_millis": 60000,
-          "rpc_timeout_multiplier": 1.0,
-          "max_rpc_timeout_millis": 60000,
+          "initial_rpc_timeout_millis": 20000,
```
schmidt-sebastian: All timeouts should stay the same (60s)

dev/src/v1/firestore_client_config.json
schmidt-sebastian: :)

dev/src/v1/index.ts
```diff
@@ -12,11 +12,12 @@
...
+import {FirestoreAdminClient} from './firestore_admin_client';
+import {FirestoreClient} from './firestore_client';
 
-const FirestoreClient = require('./firestore_client');
-const FirestoreAdminClient = require('./firestore_admin_client');
+export {FirestoreClient, FirestoreAdminClient};
 
-FirestoreClient.FirestoreClient = FirestoreClient;
-FirestoreClient.FirestoreAdminClient = FirestoreAdminClient;
+// Doing something really horrible for reverse compatibility with original JavaScript exports
```
schmidt-sebastian: Didn't get quoted on this one. I guess it's better that way...

dev/src/v1beta1/firestore_client.ts
```diff
@@ -0,0 +1,1322 @@
+// Copyright 2019 Google LLC
```
schmidt-sebastian: Note: I skipped this file.

dev/test/gapic-firestore-v1.ts
```diff
@@ -11,76 +11,134 @@
...
   it('has servicePath', () => {
     const servicePath = firestoreModule.v1.FirestoreClient.servicePath;
     assert(servicePath);
   });
-
   it('has apiEndpoint', () => {
     const apiEndpoint = firestoreModule.v1.FirestoreClient.apiEndpoint;
     assert(apiEndpoint);
   });
-
```
schmidt-sebastian: Nit: Keep empty lines (again, don't fix before merging)

dev/test/gapic-firestore-v1.ts
```diff
@@ -11,76 +11,134 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
 
-'use strict';
```
schmidt-sebastian: It would be interesting to know if the old system tests pass against the new client. Did you test that by any chance?

synth.py
```diff
@@ -6,44 +6,38 @@
...
+s.copy(v1_library, "dev", excludes=["package.json", "README.md", "src/index.ts", "src/v1/index.ts",
+    "tsconfig.json", "tslint.json", "linkinator.config.json", "webpack.config.js"])
 
 # Fix dropping of google-cloud-resource-header
 # See: https://github.com/googleapis/nodejs-firestore/pull/375
 s.replace(
-    "dev/src/v1beta1/firestore_client.js",
+    "dev/src/v1beta1/firestore_client.ts",
     "return this\._innerApiCalls\.listen\(options\);",
     "return this._innerApiCalls.listen({}, options);",
```
schmidt-sebastian: Is this still needed (or rather: does the original bug still exist in the TS generated client?). It would seem to be that this should now be caught by the type system.

tsconfig.json
```diff
@@ -1,15 +1,17 @@
 {
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
-    "allowJs": true,
-    "declaration": false, // TODO: Remove when all source are migrated to TS
```
schmidt-sebastian: I just read 16k lines when all I needed was this.

tsconfig.json
```diff
@@ -1,15 +1,17 @@
...
-    "outDir": "build"
+    "resolveJsonModule": true,
+    "outDir": "build",
+    "lib": [
+      "es2016",
+      "dom"
```
schmidt-sebastian: Is "dom" needed? I thought that was merely for browser support.

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+
+  constructor(opts?: ClientOptions) {
+    // Ensure that options include the service address and port.
+    const staticMembers = this.constructor as typeof FirestoreClient;
+    const servicePath =
+      opts && opts.servicePath
+        ? opts.servicePath
+        : opts && opts.apiEndpoint
+        ? opts.apiEndpoint
+        : staticMembers.servicePath;
```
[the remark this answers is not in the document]
alexander-fenster: We don't generate 3.7 code (with optional chaining and null coalescing). Even worse, we are going to rollback libraries to 3.6 because of some compatibility issues with 3.5. https://github.com/googleapis/google-auth-library-nodejs/issues/844 has some details.

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+      'rollback',
+      'runQuery',
+      'write',
+      'listen',
+      'listCollectionIds',
+    ];
+
+    for (const methodName of firestoreStubMethods) {
+      const innerCallPromise = this._firestoreStub.then(
+        stub => (...args: Array<{}>) => {
```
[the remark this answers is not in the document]
alexander-fenster: Does it give us any benefits? `{}` at least says it's an object (and it really is – it's either a protobuf message or an options object).

(I don't know, I was having rather hard time TypeScript'ing this piece of code so these types are the best I've got so far, but they surely can be improved!)

dev/src/v1/firestore_client.ts
alexander-fenster: It's 2020 indeed, and TypeScript still cannot process documentation on overloads :)  If we move the comment block a few lines up so it was before the overloads, it gets completely lost when transpiling to JS, and jsdoc does not get anything at all. Unfortunately, we still use jsdoc on the compiled code to build HTML docs, we haven't found any better solution for TypeScript so far.

As for callbacks - we do advertise `async`/`await` but we still accept callbacks, and dropping them would be a semver major. Actually, dropping callbacks altogether might be a good idea for this January (together with Node 8 deprecation).

dev/src/v1/firestore_client_config.json
alexander-fenster: They are coming from `firestore_grpc_service_config.json`, cc: @noahdietz here

tsconfig.json
alexander-fenster: Yep. The thing is, we support browser now (kind of). We check for `typeof window !== "undefined"` in the generated client (to enable gRPC-fallback if we are in a browser), and to make it compile, we need `dom`.

dev/src/v1/firestore_client_config.json
noahdietz: The `initial_rpc_timeout` is not a supported field anymore with the new config, @alexander-fenster this is one of the default values you decided to set it to. Also, this specific field is not the "RPC timeout" or the total_timeout/timeout_millis. The source for that is the "timeout" field int he grpc_service_config which is currently set to 60s

https://github.com/googleapis/googleapis/blob/b10e4547017ca529ac8d183e839f3c272e1c13de/google/firestore/v1/firestore_grpc_service_config.json

dev/protos/firestore_proto_api.js
```diff
@@ -2409,6 +2395,42 @@ $root.google = (function() {
             return BytesValue;
         })();
 
+        protobuf.FieldMask = (function() {
```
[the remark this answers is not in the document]
schmidt-sebastian: This was addressed.

dev/src/v1/firestore_client.ts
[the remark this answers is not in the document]
schmidt-sebastian: I don't have a strong preference - unknown feels like we know nothing about the type, which is almost true (other than the fact that it is a Proto object).

dev/src/v1/firestore_client_config.json
schmidt-sebastian: FWIW, the `googleapis` definition is correct: https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore_grpc_service_config.json#L39

dev/src/v1/firestore_client.ts
```diff
@@ -0,0 +1,1318 @@
...
+import {Transform} from 'stream';
+import * as protosTypes from '../../protos/firestore_proto_api';
+import * as gapicConfig from './firestore_client_config.json';
+
+const version = require('../../../package.json').version;
+
+/**
+ *  The Cloud Firestore service.
+ *
+ *  This service exposes several types of comparable timestamps:
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed in https://critique.corp.google.com/#review/284683228

dev/src/v1/firestore_admin_client_config.json
```diff
@@ -2,21 +2,25 @@
   "interfaces": {
     "google.firestore.admin.v1.FirestoreAdmin": {
       "retry_codes": {
+        "non_idempotent": [],
         "idempotent": [
           "DEADLINE_EXCEEDED",
-          "INTERNAL",
           "UNAVAILABLE"
```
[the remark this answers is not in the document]
schmidt-sebastian: Looks like these are just default codes that are not used. Resolving.

dev/src/v1/firestore_client_config.json
schmidt-sebastian: These seem to be default configurations that are unused. Resolving.

dev/src/v1/firestore_admin_client_config.json
```diff
@@ -2,21 +2,25 @@
...
         ],
-        "non_idempotent": []
+        "d_e_a_d_l_i_n_e_e_x_c_e_e_d_e_d_i_n_t_e_r_n_a_l_u_n_a_v_a_i_l_a_b_l_e": [
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed by https://github.com/googleapis/gapic-generator-typescript/pull/172

schmidt-sebastian:
The system tests should pass once https://github.com/googleapis/gapic-generator-typescript/pull/173 is merged and released.

dev/src/v1/firestore_client_config.json
schmidt-sebastian: https://github.com/googleapis/gapic-generator-typescript/pull/172

alexander-fenster:
I'm guessing we first need to update `synth.py` to make changes for your latest fixes, then run the generator again.

schmidt-sebastian:
Re-ran synthool and pushed changes. The only remaining blockers (in my opinion) are the setting for initial_rpc_timeout_millis and max_rpc_timeout_millis. Should we drop them?

schmidt-sebastian:
Updated all the dependencies since the CI requires a newer version of `gts` than in `package.json` and the generated clients require a newer version of Google Gax (for `gax.OperationsClient`).

alexander-fenster:
Dropping of `initial_rpc_timeout_millis` and others will be done in gax (the exported type requires those unused parameters to be set). I will first remove them from there, then from the generator. All generated libraries will be updated within the nightly synthtool run.

I think we are good to go here! I suggest we treat this PR as a breaking change just in case (feel free to merge the PR then), otherwise, if you think it's just a minor, remove `!` from the PR title before merging :)  It's up to you.

alexander-fenster:
https://github.com/googleapis/gapic-generator-typescript/pull/175 will reset the parameters to 60000. We can merge now (this change will come within a nightly synthtool run) or wait for this change to make its way into the Docker image (will happen on Monday).

willbattel:
This PR is linked in the v3.0.0 release notes as a breaking change. Is there something we need to alter to comply with this change? The only thing I could imagine this affecting is the typings of some exposed objects. In v2.X we imported a few specific types with
```
import { GeoPoint, FieldValue, Timestamp, DocumentReference } from '@google-cloud/firestore'
```
but that shouldn't have to change in v3, correct?

schmidt-sebastian:
We bumped the major version because of the risk with this PR. We are not aware of any specific breakages.

PR #808, opened 2019-12-06
feat!: remove deprecated timestampInSnapshots setting

schmidt-sebastian:
The next release will be breaking (see https://github.com/googleapis/nodejs-firestore/pull/805). I would like to use this opportunity to clean up the code and remove timestampInSnapshots

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -258,7 +258,7 @@ export class Firestore {
    * The configuration options for the GAPIC client.
    * @private
    */
-  _settings: Settings = {};
+  private _settings: Settings = {};
 
   /**
    * Whether the initialization settings can still be changed by invoking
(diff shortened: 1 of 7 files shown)


dev/src/index.ts
```diff
@@ -420,12 +406,6 @@ export class Firestore {
   settings(settings: Settings): void {
     validateObject('settings', settings);
     validateString('settings.projectId', settings.projectId, {optional: true});
-    validateBoolean(
-      'settings.timestampsInSnapshots',
-      // tslint:disable-next-line deprecation
-      settings.timestampsInSnapshots,
-      {optional: true}
-    );
```
mikelehen: Am I reading the code correctly that we don't have any particular validation to prevent "extra" settings? (i.e. if a user typos 'projectId' as 'projectID' we'll just silently ignore it rather than let them know that they messed up?)

It would be nice if we had validation like the web SDK (https://github.com/firebase/firebase-js-sdk/blob/57a0f2bd78afba64b292125522f1071b386a920d/packages/firestore/src/api/database.ts#L181) and then now that we remove timestampsInSnapshots, people will get a very clear error if they're using it.

schmidt-sebastian: No, we don't have this validation. The Server SDK supports a ton of options that are consumed by either our code, Google Gax or GRPC. I would be surprised if anyone knew the complete set of settings we support, hence I don't think we can validate them here.

PR #814, opened 2019-12-12
sample: sample for distributed counters

laljikanjareeya:
Sample for Distributed Counters concepts.

Samples for Distributed Counters in other languages:
https://firebase.google.com/docs/firestore/solutions/counters

Continuation of: [Go samples](https://github.com/GoogleCloudPlatform/golang-samples/pull/959), [Python samples](https://github.com/GoogleCloudPlatform/python-docs-samples/pull/2412).

### DIFF
--- a/samples/solution-counters.js
+++ b/samples/solution-counters.js
@@ -19,18 +19,21 @@ async function main() {
   // [START increment_counter]
   function incrementCounter(docRef, numShards) {
     const shardId = Math.floor(Math.random() * numShards);
-    const shardRef = docRef.collection("shards").doc(shardId.toString());
+    const shardRef = docRef.collection('shards').doc(shardId.toString());
     return shardRef.set({ count: FieldValue.increment(1) }, { merge: true });
   }
   // [END increment_counter]
 
   // [START get_count]
   async function getCount(docRef) {
-    const snapshotList = await docRef.collection("shards").select('count').get();
+    const querySnapshot = await docRef.collection('shards').get();
+    const documents = querySnapshot.docs;
 
-    return snapshotList.docs.reduce((count, documentSnapshot) => {
-      return count + documentSnapshot.data().count;
-    }, 0)
+    let count = 0;
+    for (const doc of documents) {
+      count += doc.get('count');
+    }
+    return count;
   }
   // [END get_count]
 
--- a/samples/solution-counters.js
+++ b/samples/solution-counters.js
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2019, Google, Inc.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+const { Firestore } = require('@google-cloud/firestore');
+
+// [START create_counter]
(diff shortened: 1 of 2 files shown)


samples/solution-counters.js
```diff
@@ -0,0 +1,63 @@
...
+    const shardRef = docRef.collection("shards").doc(shardId.toString());
+    return shardRef.set({ count: FieldValue.increment(1) }, { merge: true });
+  }
+  // [END increment_counter]
+
+  // [START get_count]
+  async function getCount(docRef) {
+    const snapshotList = await docRef.collection("shards").get();
+
+    return snapshotList.docs.reduce((count, documentSnapshot) => {
```
schmidt-sebastian: This is pretty clever but IMHO not the easiest code to understand in a sample.

Optional.

samples/solution-counters.js
```diff
@@ -0,0 +1,63 @@
...
+    }, 0)
+  }
+  // [END get_count]
+
+  // [START delete_Docs]
+  async function deleteDocs(docRef) {
+    const shardsCollectionRef = docRef.collection('shards');
+    const shardDocs = await shardsCollectionRef.get();
+    shardDocs.forEach(async (doc) => {
+      await shardsCollectionRef.doc(doc.id).delete();
```
schmidt-sebastian: This is much slower than it needs to be since it blocks for the processing of each step. You could accumulate all Promises in a list and then wait for all Promises to resolve in the end via `Promise.all()`.

Please fix.

samples/solution-counters.js
```diff
@@ -0,0 +1,63 @@
...
+    });
+  }
+  // [END delete_Docs]
+
+  // Create a new client
+  const firestore = new Firestore();
+  const docRef = firestore.doc('distributed_counter_samples/distributed_counter');
+  const numberOfShards = 10;
+  // Increase the document count
+  incrementCounter(docRef, numberOfShards).then(async () => {
```
schmidt-sebastian: Missing return?

Please fix.

samples/test/solution-counters.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+'use strict';
+
+const { execSync } = require('child_process');
+const { assert } = require('chai');
+const exec = cmd => execSync(cmd, { encoding: 'utf8' });
+
+describe('distributed counter', () => {
+    it('should increase, get counter and delete the docs', () => {
+        const output = exec('node solution-counters.js');
```
schmidt-sebastian: As eluded to in the other review, this style of testing is very fragile. I can probably live with it since it is used in the other Quickstart test.

samples/solution-counters.js
laljikanjareeya: optimize the query for getting count.

samples/solution-counters.js
laljikanjareeya: fixed!!

samples/solution-counters.js
```diff
@@ -0,0 +1,65 @@
...
+  function incrementCounter(docRef, numShards) {
+    const shardId = Math.floor(Math.random() * numShards);
+    const shardRef = docRef.collection("shards").doc(shardId.toString());
+    return shardRef.set({ count: FieldValue.increment(1) }, { merge: true });
+  }
+  // [END increment_counter]
+
+  // [START get_count]
+  async function getCount(docRef) {
+    const snapshotList = await docRef.collection("shards").select('count').get();
```
schmidt-sebastian: If you keep this, please use consistent quotation marks.

schmidt-sebastian: BTW, this returns a single QuerySnapshot, not a list.

samples/solution-counters.js
schmidt-sebastian: This makes the code even harder to follow...

What about:
```
const querySnapshot = await docRef.collection("shards").get();
const documents = querySnapshot.docs();

let count = 0;
for (const doc of documents) {
  count += doc.get("count");
}
return count;
```

While "reduce" is nice and works here, it is non-trivial and much harder to understand than a simple loop.

PR #824, opened 2019-12-18
feat: add ability to close channels

### LINKED ISSUE #769: Expose an API to close channels

Running more than one instance of Firestore in isolated, disposable contexts is not possible, as the underlying connection makes it impossible for these clients to be cleared from memory.

An example would be to use [Node.js VM](https://nodejs.org/api/vm.html).

I'd like to see an API to close all open streams, so the client can be garbage collected.

Hacking my way into `@grpc/grpc-js`, which exposes this API.

 **Additional context**

Here's an example of what it can cause :)

Screen Shot 2019-09-30 at 21 39 01

thebrianchen:
Fixes #769.


dev/src/index.ts
```diff
@@ -267,6 +267,13 @@ export class Firestore {
    */
   private _settingsFrozen = false;
 
+  /**
+   * Whether is client has been terminated. Once the Firestore instance is
+   * terminated, terminated cannot be set to false again, and all subsequent
```
schmidt-sebastian: This is a bit confusing to read. Can we just remove "terminated cannot be set to false again"?

dev/src/index.ts
```diff
@@ -1007,6 +1025,21 @@ export class Firestore {
       });
   }
 
+  /**
+   * Terminates the Firestore client and closes all open streams.
+   * @return A Promise that resolves when the client is terminated.
+   */
+  async terminate(): Promise<void> {
+    await this._clientPool.terminate();
+    this._terminated = true;
```
schmidt-sebastian: If you keep `_terminated`, you should flip these two statements. Otherwise, an operation can run during termination.

dev/src/index.ts
```diff
@@ -1007,6 +1025,21 @@ export class Firestore {
       });
   }
 
+  /**
+   * Terminates the Firestore client and closes all open streams.
+   * @return A Promise that resolves when the client is terminated.
```
schmidt-sebastian: Nit: We usually leave an empty line before the `@` section.

dev/src/index.ts
```diff
@@ -1007,6 +1025,21 @@ export class Firestore {
...
+   */
+  async terminate(): Promise<void> {
+    await this._clientPool.terminate();
+    this._terminated = true;
+  }
+
+  private verifyNotTerminated(): void {
```
schmidt-sebastian: I don't think this method is needed - the logic you added to ClientPool does this just as well, and it is used in all code paths (such as WriteBatch.commit()). I think you can remove this method and the _terminated state from this file.

dev/src/index.ts
```diff
@@ -1007,6 +1025,21 @@ export class Firestore {
       });
   }
 
+  /**
+   * Terminates the Firestore client and closes all open streams.
+   * @return A Promise that resolves when the client is terminated.
+   */
+  async terminate(): Promise<void> {
```
schmidt-sebastian: Please remember to add this to our public types.

dev/src/index.ts
thebrianchen: oops, thanks for catching that

dev/src/index.ts

dev/src/index.ts
```diff
@@ -1007,6 +1007,15 @@ export class Firestore {
       });
   }
 
+  /**
+   * Terminates the Firestore client and closes all open streams.
+   *
+   * @return A Promise that resolves when the client is terminated.
+   */
+  async terminate(): Promise<void> {
+    await this._clientPool.terminate();
```
schmidt-sebastian: Nit: This doesn’t need to be Async anymore and you can just return the Promise.

PR #828, opened 2019-12-19
feat: add support for Typescript Custom Mapping

thebrianchen:
Fixes https://github.com/googleapis/nodejs-firestore/issues/856

Added `withConverter()` methods on `CollectionReference`, `DocumentReference`, and `Query` that allow developers to use Firestore classes with generics. 

Example usage:
```
class Post {
  constructor(readonly title: string, readonly author: string) {}
  toString(): string {
    return this.title + ', by ' + this.author;
  }
}

const PostConverter = {
  toFirestore(post: Post): DocumentData {
    return {title: post.title, author: post.author};
  },
  fromFirestore(data: DocumentData): Post {
    return new Post(data.title, data.author);
  }
};

const postSnap = await firebase.firestore().collection('posts').withConverter(PostConverter).doc().get();
const post = postSnap.data();
if (post !== undefined) {
  post.title; // string
  post.toString(); // Should be defined with your overload
  post.someNonExistentProperty; // TS error
}
```


dev/test/document.ts
```diff
@@ -2020,3 +2022,28 @@ describe('listCollections() method', () => {
...
+
+  beforeEach(() => {
+    return createInstance().then(firestoreInstance => {
+      firestore = firestoreInstance;
+    });
+  });
+
+  afterEach(() => verifyInstance(firestore));
+
+  it('for DocumentReference.withConverter()', async () => {
```
schmidt-sebastian: Note that your tests do not pass since you need to initialize the mock layer for unit test. This would look something like this:


```
describe('withConverter() support', () => {
  it('for DocumentReference.withConverter()', async () => {
    const doc = document('documentId', 'author', 'author', 'title', 'post');
    const overrides: ApiOverride = {
      commit: (request, options, callback) => {
        const expectedRequest = set({
          document: doc,
        });
        requestEquals(request, expectedRequest);

        callback(null, writeResult(1));
      },
      batchGetDocuments:  () => {
        const stream = through2.obj();
        setImmediate(() => {
          stream.push({found: doc, readTime: {seconds: 5, nanos: 6}});
          stream.push(null);
        });

        return stream;
      }
    };

    return createInstance(overrides).then(async firestore => {
      const docRef = firestore
          .collection('collectionId')
          .doc('documentId')
          .withConverter(postConverter);

      await docRef.set(new Post('post', 'author'));
      const postData = await docRef.get();
      const post = postData.data();
      expect(post).to.not.equal(undefined);
      expect(post!.byline()).to.equal('post, by author');
    });
  });
});
```

thebrianchen: Done.

dev/src/watch.ts
```diff
@@ -434,7 +451,7 @@ abstract class Watch {
    * invalid.
    * @private
    */
-  private onData(proto: api.IListenResponse): void {
+  private onData<T>(proto: api.IListenResponse): void {
```
schmidt-sebastian: This looks like it is not needed.

dev/src/write-batch.ts
```diff
@@ -286,19 +287,19 @@ export class WriteBatch {
...
+    documentRef: DocumentReference<T>,
+    data: T,
     options?: SetOptions
   ): WriteBatch {
     validateSetOptions('options', options, {optional: true});
     const mergeLeaves = options && options.merge === true;
     const mergePaths = options && options.mergeFields;
-
     validateDocumentReference('documentRef', documentRef);
+    let firestoreData: DocumentData = documentRef._converter.toFirestore(data);
```
schmidt-sebastian: You should be able to drop `: DocumentData`.

dev/src/reference.ts
```diff
@@ -1940,6 +1992,19 @@ export class Query {
...
+   * Applies a custom data converter to this Query, allowing you to use your
+   * own custom model objects with Firestore. When you call get() on the
+   * returned Query, the provided converter will convert between Firestore
+   * data and your custom type U.
+   *
+   * @param converter Converts objects to and from Firestore.
+   * @return A Query<U> that uses the provided converter.
+   */
+  withConverter<U>(converter: FirestoreDataConverter<U>): Query<U> {
+    return new Query<U>(this.firestore, this._queryOptions, converter);
```
schmidt-sebastian: Here and everywhere: Should we validate that a `converter` is provided?

dev/system-test/firestore.ts
```diff
@@ -288,7 +289,7 @@ describe('DocumentReference class', () => {
       })
       .then(doc => {
         const data = doc.data()!;
-        expect(data.pathValue.path).to.equal(
+        expect((data.pathValue as DocumentReference).path).to.equal(
```
schmidt-sebastian: This should probably return a DocumentReference<unknown> and needs to be fixed in the Serializer (serializer.ts, line 55)

dev/test/collection.ts
```diff
@@ -14,15 +14,21 @@
 
 import {expect} from 'chai';
 
+import through2 = require('through2');
```
schmidt-sebastian: Prefer using `from` for eventual TreeShaking support.

dev/system-test/firestore.ts
```diff
@@ -288,7 +289,7 @@ describe('DocumentReference class', () => {
...
           allSupportedTypesObject.pathValue.path
         );
         delete data.pathValue;
```
schmidt-sebastian: You should really have a System test that exercises the new surface area.

dev/test/watch.ts
```diff
@@ -141,8 +141,7 @@ function snapshot(
   ref: DocumentReference,
   data: DocumentData
 ): QueryDocumentSnapshot {
-  const snapshot = new DocumentSnapshotBuilder();
-  snapshot.ref = ref;
+  const snapshot = new DocumentSnapshotBuilder(ref);
   snapshot.fieldsProto = ref.firestore._serializer!.encodeFields(data);
   snapshot.readTime = new Timestamp(0, 0);
   snapshot.createTime = new Timestamp(0, 0);
```
schmidt-sebastian: Please also update test/typescript.ts to make sure the new types are what we expect. We may drop this file at some point, but we should keep it updated as long as it exists.

dev/test/collection.ts
```diff
@@ -169,4 +175,51 @@ describe('Collection interface', () => {
...
+    return createInstance(overrides).then(async firestore => {
+      const docRef = firestore
+        .collection('collectionId')
+        .withConverter(postConverter)
+        .doc('documentId');
+
+      await docRef.set(new Post('post', 'author'));
+      const postData = await docRef.get();
+      const post = postData.data();
+      expect(post).to.not.equal(undefined);
```
schmidt-sebastian: Nit: to.not.be.undefined

dev/test/collection.ts
```diff
@@ -169,4 +175,51 @@ describe('Collection interface', () => {
...
+  });
+
+  it('drops the converter when calling CollectionReference<T>.parent()', () => {
+    return createInstance().then(async firestore => {
+      const postsCollection = firestore
+        .collection('users/user1/posts')
+        .withConverter(postConverter);
+
+      const usersCollection = postsCollection.parent;
+      expect(usersCollection!.isEqual(firestore.doc('users/user1'))).to.be.true;
```
schmidt-sebastian: This tests relies on the fact that we compare converters in `isEqual`. You can keep the test as is, but you should change the semantics of `isEqual` so that the test would fail if we don't drop the converter.

dev/test/document.ts
```diff
@@ -14,6 +14,7 @@
 
 import {expect} from 'chai';
 import {GoogleError, Status} from 'google-gax';
+import through2 = require('through2');
```
schmidt-sebastian: Same comment about using `import from` (which is the ES6 style module import).

dev/test/collection.ts
```diff
@@ -169,4 +175,51 @@ describe('Collection interface', () => {
...
+    };
+
+    return createInstance(overrides).then(async firestore => {
+      const docRef = firestore
+        .collection('collectionId')
+        .withConverter(postConverter)
+        .doc('documentId');
+
+      await docRef.set(new Post('post', 'author'));
+      const postData = await docRef.get();
```
schmidt-sebastian: Should this use CollectionReference.get() since this test is meant to exercise the CollectionReference API? (You can also drop it since you are already testing this in query.ts)

dev/test/document.ts
```diff
@@ -2004,3 +2007,51 @@ describe('listCollections() method', () => {
...
+    return createInstance(overrides).then(async firestore => {
+      const docRef = firestore
+        .collection('collectionId')
+        .doc('documentId')
+        .withConverter(postConverter);
+
+      await docRef.set(new Post('post', 'author'));
+      const postData = await docRef.get();
+      const post = postData.data();
+      expect(post).to.not.equal(undefined);
```
schmidt-sebastian: Nit: to.not.be.undefined

dev/test/query.ts
```diff
@@ -621,6 +627,38 @@ describe('query interface', () => {
...
+        .collection('collectionId')
+        .doc('documentId')
+        .set({title: 'post', author: 'author'});
+      const posts = await firestore
+        .collection('collectionId')
+        .where('title', '==', 'post')
+        .withConverter(postConverter)
+        .get();
+      expect(posts.size).to.equal(1);
+      expect(posts.docs[0].data()!.byline()).to.equal('post, by author');
```
schmidt-sebastian: You should be able to remove the bang since queries only return documents that have data (via QueryDocumentSnapshot).

dev/test/util/helpers.ts
```diff
@@ -334,3 +334,19 @@ export function stream<T>(...elements: Array<T | Error>): Duplex {
 export function response<T>(result: T): Promise<[T, unknown, unknown]> {
   return Promise.resolve([result, undefined, undefined]);
 }
+
+export class Post {
+  constructor(readonly title: string, readonly author: string) {}
+  byline(): string {
```
schmidt-sebastian: Nit: toString()?

Optional.

dev/test/watch.ts
```diff
@@ -141,8 +141,7 @@ function snapshot(
   ref: DocumentReference,
   data: DocumentData
 ): QueryDocumentSnapshot {
-  const snapshot = new DocumentSnapshotBuilder();
-  snapshot.ref = ref;
+  const snapshot = new DocumentSnapshotBuilder(ref);
```
schmidt-sebastian: Please add a test for onSnapshot() with generics.

types/firestore.d.ts
```diff
@@ -290,7 +345,7 @@ declare namespace FirebaseFirestore {
      * @param data The object data to serialize as the document.
```
schmidt-sebastian: I think the generic type for Transaction.getAll() is missing

types/firestore.d.ts
```diff
@@ -651,7 +706,7 @@ declare namespace FirebaseFirestore {
      * @return An unsubscribe function that can be called to cancel
      * the snapshot listener.
      */
-    onSnapshot(onNext: (snapshot: DocumentSnapshot) => void,
+    onSnapshot(onNext: (snapshot: DocumentSnapshot<T>) => void,
```
schmidt-sebastian: This is also not exercised  by any tests.

types/firestore.d.ts
```diff
@@ -1143,15 +1223,28 @@ declare namespace FirebaseFirestore {
...
+    /**
+     * Applies a custom data converter to this CollectionReference, allowing you
+     * to use your own custom model objects with Firestore. When you call add()
+     * on the returned CollectionReference instance, the provided converter will
+     * convert between Firestore data and your custom type U.
+     *
+     * @param converter Converts objects to and from Firestore.
+     * @return A CollectionReference<U> that uses the provided converter.
+     */
+    withConverter<U>(
```
schmidt-sebastian: As eluded to above, I don't think this is necessary.

dev/src/index.ts
```diff
@@ -1024,11 +1029,19 @@ export class Firestore {
...
-              const orderedDocuments: DocumentSnapshot[] = [];
+              const orderedDocuments: Array<DocumentSnapshot<T>> = [];
+
               for (const docRef of docRefs) {
                 const document = retrievedDocuments.get(docRef.path);
                 if (document !== undefined) {
-                  orderedDocuments.push(document);
+                  // Recreate the DocumentSnapshot with the DocumentReference
+                  // containing the original converter.
+                  const finalDoc = new DocumentSnapshotBuilder(docRef);
```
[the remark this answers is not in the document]
thebrianchen: you're only complimenting yourself here :P

dev/src/reference.ts
```diff
@@ -216,8 +223,12 @@ export class DocumentReference implements Serializable {
    *   console.log(`Found ${results.size} matches in parent collection`);
    * }):
    */
-  get parent(): CollectionReference {
-    return new CollectionReference(this._firestore, this._path.parent()!);
+  get parent(): CollectionReference<T> {
+    return new CollectionReference(
```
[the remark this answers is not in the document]
thebrianchen: I don't think you have to specify it in this case because `this._converter` is typed to `T` already, so it's automatically inferred.

Or is this a more style / code readability issue?

dev/src/reference.ts
```diff
@@ -437,7 +448,11 @@ export class DocumentReference implements Serializable {
 
     const writeBatch = new WriteBatch(this._firestore);
     return writeBatch.update
-      .apply(writeBatch, [this, dataOrField, ...preconditionOrValues])
+      .apply(writeBatch, [
+        this as DocumentReference<unknown>,
```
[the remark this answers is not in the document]
thebrianchen: I needed it before when I using `writeBatch.update.apply`, but changing it to `update()` removes the need to cast.

dev/src/reference.ts
```diff
@@ -437,7 +448,11 @@ export class DocumentReference implements Serializable {
...
+        dataOrField,
+        ...preconditionOrValues,
```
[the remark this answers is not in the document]
thebrianchen: Done.

dev/src/index.ts
```diff
@@ -695,20 +696,22 @@ export class Firestore {
       );
     }
 
-    const document = new DocumentSnapshotBuilder();
-
+    let ref;
```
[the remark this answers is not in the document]

Question: if the IDE doesn't display the variables as `any`, does that mean that the transpiler has deduced them successfully, or is there some other way to check?

dev/src/reference.ts
```diff
@@ -1146,7 +1185,7 @@ export class Query {
     const options = this._queryOptions.with({
       fieldFilters: this._queryOptions.fieldFilters.concat(fieldFilter),
     });
-    return new Query(this._firestore, options);
+    return new Query(this._firestore, options, this._converter);
```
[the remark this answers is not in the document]
thebrianchen: I considered, but that would require QueryOption to be typed with `T`, and that seemed more ugly. Thoughts?

dev/src/reference.ts
```diff
@@ -1663,8 +1706,17 @@ export class Query {
...
           if (result.document) {
-            const document = result.document;
-            docs.push(document);
+            const document: QueryDocumentSnapshot<T> = result.document;
+            const finalDoc = new DocumentSnapshotBuilder(
+              document.ref.withConverter(this._converter)
+            );
+            // Recreate the QueryDocumentSnapshot with the DocumentReference
+            // containing the original converter.
+            finalDoc.fieldsProto = document.fieldsProto;
```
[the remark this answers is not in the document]
thebrianchen: I tried moving it to `_stream()`, but we run into problems since we mock the output of `stream()` in the integration tests, so I think we would be mocking out functionality in that case.

Also, we currently do the conversion for `getAll()` after we receive it in the stream.

dev/src/reference.ts
thebrianchen: Would this be for checking JS code? We don't currently validate converters in the web SDK. Do we usually verify all user inputs beyond the type system?

dev/src/reference.ts
```diff
@@ -2148,12 +2217,31 @@ export class CollectionReference extends Query {
...
+   * to use your own custom model objects with Firestore. When you call add()
+   * on the returned CollectionReference instance, the provided converter will
+   * convert between Firestore data and your custom type U.
+   *
+   * @param converter Converts objects to and from Firestore.
+   * @return A CollectionReference<U> that uses the provided converter.
+   */
+  withConverter<U>(
+    converter: FirestoreDataConverter<U>
+  ): CollectionReference<U> {
```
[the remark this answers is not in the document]
thebrianchen: How does that work? `Query<U>` doesn't match `this`, which is `Query<T>`.  I'm running into the following error when I change the return type to `this` (am I missing something?):
```
Type 'Query<U>' is not assignable to type 'this'
```

dev/src/watch.ts
```diff
@@ -43,7 +56,8 @@ const WATCH_TARGET_ID = 0x1;
 /*!
  * Sentinel value for a document remove.
  */
-const REMOVED = {} as DocumentSnapshotBuilder;
+// tslint:disable-next-line:no-any
+const REMOVED = {} as DocumentSnapshotBuilder<any>;
```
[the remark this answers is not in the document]
thebrianchen: REMOVED is used in `changeMap` ([here](https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/watch.ts#L152)), so it needs to be `DocumentSnapshotBuilder`.

dev/src/watch.ts
```diff
@@ -281,10 +298,10 @@ abstract class Watch {
         }
       } else if (this.docMap.has(name)) {
         value.readTime = readTime;
-        updates.push(value.build() as QueryDocumentSnapshot);
+        updates.push(value.build() as QueryDocumentSnapshot<T>);
```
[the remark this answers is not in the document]
thebrianchen: Can you elaborate on what you mean?

`DocumentSnapshotBuilder` is already typed with `T` where, and the cast is used to to cast the return of `build()` from `QueryDocumentSnapshot<T> | DocumentSnapshot<T>`.

dev/src/watch.ts
thebrianchen: Good catch, thanks!

dev/system-test/firestore.ts
thebrianchen: Hmm, I removed the cast to DocumentReference and it still works fine. I think I probably forgot to remove it.

dev/test/util/helpers.ts
thebrianchen: I just realized that the problem with using

dev/test/collection.ts
thebrianchen: I don't think `CollectionReference.get()` exists? 😮 

This test makes susre that calling `CollectionReference.doc()` returns a DocumentReference that has the converter.

dev/test/collection.ts
thebrianchen: Added the `this.converter === other.converter` check for isEqual.

dev/src/reference.ts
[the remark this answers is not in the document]
thebrianchen: Talked in person: replacing with `this` won't work. Marking as resolved.

dev/test/watch.ts
thebrianchen: Done.

types/firestore.d.ts
thebrianchen: After talking in person, it is necessary, see above.

types/firestore.d.ts
thebrianchen: Done.

dev/src/index.ts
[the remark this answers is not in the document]
schmidt-sebastian: VSCode likely gets its type information from the transpiler, so seeing the correct type is a good indication. Another thing we could do is turn on `no-implicit-any`, but that will probably result in a number of errors initially.

dev/src/reference.ts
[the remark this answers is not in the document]
schmidt-sebastian: Since we are explicitly specifying the return type, I would prefer if we also specify the type of the generic. To me this looks like it is using the default value (T=DocumentData), which is not always correct.

dev/src/reference.ts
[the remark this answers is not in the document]
schmidt-sebastian: I just played with it. Despite the additional generics, I think it is a bit cleaner. It keeps query modification encapsulated, which was the original goal behind QueryOptions. This is totally optional, but if you have the bandwidth to play around with `QueryOptions<T>` I would be more than happy to review :)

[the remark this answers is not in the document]
schmidt-sebastian: My experiment: https://gist.github.com/schmidt-sebastian/5473b71d05553ca8db01553e66979805

dev/src/reference.ts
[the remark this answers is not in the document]
schmidt-sebastian: I was able to get this to work with no changes in the tests: https://gist.github.com/schmidt-sebastian/f021aa6928876ed240a622ca0b3b3aa4. This encapsulates the conversion within the method, which matches the behavior of getAll() (IMHO).

dev/src/document.ts
```diff
@@ -52,14 +49,20 @@ export class DocumentSnapshotBuilder {
   /** The time when this document was last updated. */
   updateTime?: Timestamp;
 
+  /**
```
schmidt-sebastian: Nit: Since this doesn't describe what "constructor" is doing, this comment should use `/* */` or `//`. As it stands currently, the comment becomes part of the API documentation.

dev/src/document.ts
```diff
@@ -116,14 +118,13 @@ export class DocumentSnapshot {
    * if the document does not exist).
    */
   constructor(
-    ref: DocumentReference,
-    fieldsProto?: ApiMapValue,
+    ref: DocumentReference<T>,
+    readonly fieldsProto?: ApiMapValue,
```
schmidt-sebastian: I am undecided whether this should be `fieldsProto` or `_ fieldsProto`. This is a public field for internal consumption.

dev/src/index.ts
```diff
@@ -647,31 +648,31 @@ export class Firestore {
...
     encoding: 'json'
-  ): QueryDocumentSnapshot;
-  snapshot_(
+  ): QueryDocumentSnapshot<T>;
+  snapshot_<T = DocumentData>(
     documentOrName: api.IDocument | {[k: string]: unknown} | string,
     readTime?: google.protobuf.ITimestamp | string,
     encoding?: 'json' | 'protobufJS'
-  ): DocumentSnapshot {
+  ): DocumentSnapshot<T> {
```
schmidt-sebastian: These typings are a little off if I am not mistaken. Don't we always return `DocumentSnapshot<DocumentData>`?

dev/src/transaction.ts
```diff
@@ -419,10 +415,10 @@ export class Transaction {
...
  * an optional ReadOptions object.
  */
-export function parseGetAllArguments(
-  documentRefsOrReadOptions: Array<DocumentReference | ReadOptions>
-): {documents: DocumentReference[]; fieldMask: FieldPath[] | null} {
-  let documents: DocumentReference[];
+export function parseGetAllArguments<T>(
+  documentRefsOrReadOptions: Array<DocumentReference<T> | ReadOptions>
+): {documents: Array<DocumentReference<T>>; fieldMask: FieldPath[] | null} {
+  let documents: Array<DocumentReference<T>>;
```
schmidt-sebastian: The way `getAll()` is implemented, it should be straightforward to support references of different types. Can you see if we can add an overload that supports `any`? We may be able to just redeclare `getAll<T>` as `getAll<T = any>`, as long as the compiler still types `getAll(doc<Foo>, doc<Foo>)` as returning `DocSnap<Foo>`.

dev/src/watch.ts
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good. You could also move it into Watch and type it `<T>` but that is probably not worth it.

dev/src/watch.ts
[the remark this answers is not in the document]
schmidt-sebastian: Oh. I forgot that we need to cast to get rid of `DocumentSnapshot`. Never mind.

dev/test/collection.ts
```diff
@@ -14,15 +14,21 @@
 
 import {expect} from 'chai';
 
+import * as through2 from 'through2';
```
schmidt-sebastian: Nit: Move up to the previous block since it is an external import.

dev/test/collection.ts
schmidt-sebastian: There is a `Query.get()` that is available via the `CollectionReference` API. 

If you want to test `CollectionReference.doc() ` can you update the name of the test?

dev/test/typescript.ts
```diff
@@ -33,6 +33,7 @@ import Precondition = FirebaseFirestore.Precondition;
 import SetOptions = FirebaseFirestore.SetOptions;
 import Timestamp = FirebaseFirestore.Timestamp;
 import Settings = FirebaseFirestore.Settings;
+import {defaultConverter} from '../src/types';
```
schmidt-sebastian: Would it be possible to create your own converter, which will also test the types on `FirestoreDataConverter`?

dev/src/transaction.ts
thebrianchen: After our conversation, we've decided not to cast to `any` by default since the user would have to cast to `<any>` themselves anyways.

dev/src/reference.ts
```diff
@@ -957,10 +996,14 @@ export class QueryOptions {
...
    */
-  with(settings: Partial<QueryOptions>): QueryOptions {
+  with(
+    settings: Partial<QueryOptions<T>>,
+    converter = defaultConverter as FirestoreDataConverter<T>
+  ): QueryOptions<T> {
     return new QueryOptions(
       coalesce(settings.parentPath, this.parentPath)!,
       coalesce(settings.collectionId, this.collectionId)!,
+      converter,
```
schmidt-sebastian: Hm. This is scary. This resets the converter implicitly, even when it is passed in. If possible, can you drop the `converter` argument and change `settings` to `Partial<Omit<QueryOptions<T>, 'converter'>>`. See https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk

dev/src/types.ts
```diff
@@ -92,6 +92,77 @@ export type UnaryMethod<Req, Resp> = (
...
+   * Called by the Firestore SDK to convert Firestore data into an object of
+   * type T.
+   */
+  fromFirestore(data: DocumentData): T;
+}
+
+/**
+ * A default converter to use when none is provided.
+ */
+export const defaultConverter = {
```
schmidt-sebastian: Do you mind explicitly typing this as FirestoreDataConverter?

Also add `@private` to the JSDoc.

dev/src/types.ts
```diff
@@ -92,6 +92,77 @@ export type UnaryMethod<Req, Resp> = (
...
+ *   .doc().get();
+ * const post = postSnap.data();
+ * if (post !== undefined) {
+ *   post.title; // string
+ *   post.toString(); // Should be defined
+ *   post.someNonExistentProperty; // TS error
+ * }
+ * ```
+ */
+export interface FirestoreDataConverter<T> {
```
schmidt-sebastian: Since this is a type, this does not actually show up in our JSDocs. Can you move this example to the `withConverter()` methods? Without this interface or this example, these methods are not very explanatory. 

You can look at the docs output yourself by running `npm run docs`. This spits the docs into ./docs/

dev/test/document.ts
```diff
@@ -15,8 +15,10 @@
 import {expect} from 'chai';
 import {GoogleError, Status} from 'google-gax';
 
+import * as through2 from 'through2';
```
schmidt-sebastian: Move import into previous block.

dev/test/document.ts
```diff
@@ -2004,3 +2008,51 @@ describe('listCollections() method', () => {
...
+
+  beforeEach(() => {
+    return createInstance().then(firestoreInstance => {
+      firestore = firestoreInstance;
+    });
+  });
+
+  afterEach(() => verifyInstance(firestore));
+
+  it('for DocumentReference.withConverter().get()', async () => {
```
schmidt-sebastian: s/DocumentReference.withConverter()/DocumentReference.get() since withConverter() is already part of the test name (via the describe).

dev/test/typescript.ts
```diff
@@ -33,6 +33,7 @@ import Precondition = FirebaseFirestore.Precondition;
 import SetOptions = FirebaseFirestore.SetOptions;
 import Timestamp = FirebaseFirestore.Timestamp;
 import Settings = FirebaseFirestore.Settings;
+import {FirestoreDataConverter} from '@google-cloud/firestore';
```
schmidt-sebastian: Does `import FirestoreDataConverter from FirebaseFirestore.FirestoreDataConverter` not work?

Also, if it does - please sort the imports.

schmidt-sebastian:
Can you also make the description of the PR a bit more expressive so that users can you it as a guide for how this feature works?

dev/test/typescript.ts
thebrianchen: Forgot to export it in the `index.ts` file, but now it does.

dev/src/reference.ts
thebrianchen: Done.

PR #833, opened 2019-12-20
fix: update terminate test to not be racey

thebrianchen:
Is this more along the lines of what you're talking about? Realized there's not really a need for the snapshot listener, since it was just a manual check to make sure nothing errored out.


thebrianchen:
Closing since it's already being implemented in #845.

PR #837, opened 2019-12-26
feat: allow specifying how many idle GRPC channels to keep

schmidt-sebastian:
This PR lets users influence the memory profile of their Firestore usage:
- They can set the minimum client pool size to 0 to close all GRPC channels when Firestore is completely idle
- They can use a larger number to reduce the startup costs of new GRPC channels.

This PR aims to preserve the current behavior by always leaving at least one GRPC channel if no option is provided.

b/146363799

### DIFF
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -67,13 +67,14 @@ export class ClientPool<T> {
    */
   private acquire(requestTag: string): T {
     let selectedClient: T | null = null;
-    let selectedRequestCount = 0;
+    let selectedClientRequestCount = 0;
 
     for (const [client, requestCount] of this.activeClients) {
-      // Bin pack requests to reduce the maximize the number of idle clients as
-      // operations start to complete
+      // Use the "most-full" client that can still accommodate the request
+      // in order to maximize the number of idle clients as operations start to
+      // complete.
       if (
-        requestCount > selectedRequestCount &&
+        requestCount > selectedClientRequestCount &&
         requestCount < this.concurrentOperationLimit
       ) {
         logger(
@@ -83,11 +84,18 @@ export class ClientPool<T> {
           this.concurrentOperationLimit - requestCount
         );
         selectedClient = client;
-        selectedRequestCount = requestCount;
+        selectedClientRequestCount = requestCount;
       }
     }
 
-    if (!selectedClient) {
+    if (selectedClient) {
+      logger(
+        'ClientPool.acquire',
+        requestTag,
+        'Re-using existing client with %s remaining operations',
+        this.concurrentOperationLimit - selectedClientRequestCount
+      );
+    } else {
       logger('ClientPool.acquire', requestTag, 'Creating a new client');
       selectedClient = this.clientFactory();
       assert(
@@ -96,7 +104,7 @@ export class ClientPool<T> {
       );
     }
 
-    this.activeClients.set(selectedClient, selectedRequestCount + 1);
+    this.activeClients.set(selectedClient, selectedClientRequestCount + 1);
 
     return selectedClient!;
   }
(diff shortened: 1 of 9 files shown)


dev/src/pool.ts
```diff
@@ -99,23 +107,43 @@ export class ClientPool<T> {
...
+
+    // Compute the remaining capacity of the ClientPool. If the capacity exceeds
+    // the total capacity that `maxIdleClients` could hold, garbage collect. We
+    // look at the capacity rather than just at the current request count to
+    // allows us to:
+    // - Use `maxIdleClients:1` to preserve legacy behavior (there is always at
+    //   least one active client as a single client can never exceed the
+    //   concurrent operation limit by itself).
+    // - Use `maxIdleClients:0` to shut down the client pool completely when all
+    //   clients are idle.
```
mikelehen: FWIW I read this comment 2 or 3 times, and read the code once, and the code still makes more sense to me than the comment.  So it might be worth rephrasing / paring down the comment.  I'm not 100% sure what it's trying to get across.  Is it justifying the existence of the maxIdleClients setting (in which maybe the bulk of this comment should go there?)?  Or is there actually some nuance to this code that needs 10 lines of explanation?

dev/src/types.ts
```diff
@@ -72,6 +72,13 @@ export interface Settings {
   /** Whether to use SSL when connecting. */
   ssl?: boolean;
 
+  /**
+   * The maximum number of idle GRPC channels to keep. A smaller number of idle
+   * channels reduces memory usage but increases request latency for clients
+   * with fluctuating request rates. Defaults to 1.
```
mikelehen: Elsewhere you document that setting it to 0 allows the client pool to shut down completely.  Is that a behavior that should be documented here?

dev/src/pool.ts
```diff
@@ -66,8 +69,13 @@ export class ClientPool<T> {
     let selectedClient: T | null = null;
     let selectedRequestCount = 0;
 
-    this.activeClients.forEach((requestCount, client) => {
-      if (!selectedClient && requestCount < this.concurrentOperationLimit) {
+    for (const [client, requestCount] of this.activeClients) {
+      // Bin pack requests to reduce the maximize the number of idle clients as
```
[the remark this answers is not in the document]
schmidt-sebastian: Your comment is not only grammatically correct, but also more helpful. Thanks!

dev/src/pool.ts
```diff
@@ -77,7 +85,7 @@ export class ClientPool<T> {
         selectedClient = client;
```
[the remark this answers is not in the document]
schmidt-sebastian: Good catch. I moved it outside the loop.

dev/src/pool.ts
```diff
@@ -99,23 +107,43 @@ export class ClientPool<T> {
    * @private
    */
   private async release(requestTag: string, client: T): Promise<void> {
-    let requestCount = this.activeClients.get(client) || 0;
+    const requestCount = this.activeClients.get(client) || 0;
     assert(requestCount > 0, 'No active request');
+    this.activeClients.set(client, requestCount! - 1);
```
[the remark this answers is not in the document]
schmidt-sebastian: Uhm. I hope I had my reasons when I added it first, but they are not immediately apparent now. Removed.

dev/src/pool.ts
schmidt-sebastian: I was trying to justify that the code doesn't just count the number of idle clients, but instead counts the number of idle "request slots". If this comment doesn't help, then I am ok with removing it.

dev/src/types.ts
schmidt-sebastian: That's a good suggestion. Added (here and in the other two places...)

PR #841, opened 2019-12-29
refactor: use explicit mocha imports

JustinBeckwith:


JustinBeckwith:
I only work on the super important stuff

Issue #799, opened 2019-11-26
[Bug] Unable to add documents to subcollections using the functions shell


DeveloperUX:
I can't seem to be able to `add` data to subcollections in firestore using the functions shell emulator

#### Environment details

  - OS: Mac OS Catalin
  - Node.js version: v8.15.0
  - npm version: 6.7.0
  - `@google-cloud/firestore` version: "firebase-admin": "8.7.0"

#### Steps to reproduce

  1. Create a simple script that you can run in Firebase shell (ex):
```
admin.initializeApp({...});
const db = admin.firestore();

const run = async () => {
  await db.collection('papers')
      .doc(docId)
      .collection('pens')
      .add({x: 'a'});
};
run();
```
  2. Run the above script through the emulator (ex):
```
firebase functions:shell < script.js
```

**Expected Result:**
This code should run without any issues and add the object `{x: 'a'}` to the `pens` subcollection

**Actual Result:**
This code fails when run through the functions:shell but runs fine when added as a cloud function and run through the cloud function

**Error Stack**
```
(node:90760) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'constructor' of null
    at extractBaseClassName (/.../functions/node_modules/@google-cloud/firestore/build/src/validate.js:35:33)
    at Object.customObjectMessage (/.../functions/node_modules/@google-cloud/firestore/build/src/validate.js:55:26)
    at Object.validateDocumentData (/.../functions/node_modules/@google-cloud/firestore/build/src/write-batch.js:608:36)
    at CollectionReference.add (/.../functions/node_modules/@google-cloud/firestore/build/src/reference.js:1765:23)
    at runB (repl:43:6)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Stackoverflow Question:
https://stackoverflow.com/questions/59013268/unable-to-add-new-documents-to-a-collection-in-firestore-using-functionsshell

DeveloperUX:
I'm guessing this has been recorded as a bug considering that the "bug" label was added. Looking forward to the fix. If I can help let me know

bcoe:
@schmidt-sebastian any thoughts on this one?

schmidt-sebastian:
@DeveloperUX This does not reproduce for me. I wonder if this might be caused by a package that is rewriting some parts of the prototype chain. For what it is worth, the input `{x: 'a'}` should not even invoke `customObjectMessage()`.

```
export function validateDocumentData(
[...]

PR #842, opened 2019-12-30
fix: support Objects created with Object.create({})

schmidt-sebastian:
This changes our plain object check to use the constructor name, and also uses the constructor name in all of our object checks throughout.

fixes: https://github.com/googleapis/nodejs-firestore/issues/799


dev/src/util.ts
```diff
@@ -75,6 +77,23 @@ export function isObject(value: unknown): value is {[k: string]: unknown} {
...
+ * @private
+ * @param input The argument to verify.
+ * @returns 'true' if the input can be a treated as a plain object.
+ */
+export function isPlainObject(input: unknown): input is DocumentData {
+  return (
+    isObject(input) &&
+    (Object.getPrototypeOf(input) === Object.prototype ||
+      Object.getPrototypeOf(input) === null ||
+      input.constructor.name === 'Object')
```
thebrianchen: According to this stackoverflow [post](https://stackoverflow.com/questions/7039260/can-i-use-constructor-name-to-detect-types-in-javascript), `obj.constructor` can be changed by the user, so they recommend using `instanceof` instead.

dev/test/index.ts
```diff
@@ -1073,7 +1073,7 @@ describe('getAll() method', () => {
         expect(actualErrorAttempts).to.deep.eq(expectedErrorAttempts);
       });
     });
-  });
+  }).timeout(5000);
```
thebrianchen: Still needed?

dev/src/util.ts
schmidt-sebastian: The recommendation here does not quite work for us. We don't have access to the user's type and `instanceof Object` returns true for custom types as well as plain objects such as `{}`.

It also looks like the other suggestion `Object.prototype.toString.call` just returns `[object Object]` for custom types, so that is likely out of the question too,

dev/test/index.ts
schmidt-sebastian: This is in master now, so the diff should eventually disappear: https://github.com/googleapis/nodejs-firestore/blob/master/dev/test/index.ts#L1103

dev/test/util.ts
```diff
@@ -0,0 +1,35 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {expect} from 'chai';
+import {isPlainObject} from '../src/util';
+
+describe('util', () => {
+  it('isPlainObject supports Object.create()', () => {
```
thebrianchen: nitting for nits sake: Replace 'supports' with 'allows'?

Issue #789, opened 2019-11-01
Streams do not provide the full Node.js stream API (Lack of Symbol.asyncIterator)


novemberborn:
`Query#stream()` returns a stream that doesn't quite implement the Node.js stream API. For instance `[Symbol.asyncIterator]` is missing.

#### Environment details

  - Node.js version: 12.13.0
  - npm version: 6.12.1
  - `@google-cloud/firestore` version: 2.5.0

#### Steps to reproduce

  1. Create a query and call `stream()` on it
  2. On the resulting object, invoke [`[Symbol.asyncIterator]()`](https://nodejs.org/api/stream.html#stream_readable_symbol_asynciterator)
  3. This crashes, since that property is `undefined`

bcoe:
@novemberborn :wave: small world :smile: thanks for the bug report :+1: have you found a reasonable workaround for the time being for the work you're doing?

novemberborn:
Hey @bcoe! Yea, I've got some code to read the stream and provide an async iterator over it.

PR #843, opened 2019-12-30
feat: add Symbol.asyncInterator to Query.stream()

schmidt-sebastian:
This PR removes the dependency on `bun`, which exposed an old Stream interface that did not expose Symbol.asyncInterator. 

A lot of the code churn in this PR is combining `readStream` and `readWriteStream`, which were almost identical before (readStream sent request to the GAPIC method directly, whereas readWriteStream sent it to the stream). The new method handles all Stream callbacks uniformly and combines both stream modes. 

A disclaimer: I don't really understand Node streams, but we have semi-good test coverage for most error cases.

Fixes https://github.com/googleapis/nodejs-firestore/issues/789

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1233,7 +1233,7 @@ export class Firestore {
             'Received stream error:',
             err
           );
-          // We execute the forwarding of the 'error' event via setTimeout() as
+          // We execute the forwarding of the 'error' event via setImmediate() as
           // V8 guarantees that the Promise chain returned from this method
           // is resolved before any code executed via setImmediate(). This
           // allows the caller to attach an error handler.
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-import * as bun from 'bun';
 import {CallOptions} from 'google-gax';
+import {Duplex, PassThrough} from 'stream';
 import * as through2 from 'through2';
 import {URL} from 'url';
 
@@ -950,7 +950,7 @@ export class Firestore {
     const self = this;
 
     return self
-      .readStream('batchGetDocuments', request, requestTag, true)
+      .requestStream('batchGetDocuments', 'unidirectional', request, requestTag)
       .then(stream => {
         return new Promise<DocumentSnapshot[]>((resolve, reject) => {
           stream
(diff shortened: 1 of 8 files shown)


schmidt-sebastian:
@bcoe It looks like the coverage decrease in this PR is because of the license comment I had to add to external-modules.d.ts.

dev/src/index.ts
```diff
@@ -1166,144 +1166,100 @@ export class Firestore {
...
+            'Firestore._initializeStream',
+            requestTag,
+            'Received stream error:',
+            err
+          );
+          // We execute the forwarding of the 'error' event via setTimeout() as
+          // V8 guarantees that the Promise chain returned from this method
+          // is resolved before any code executed via setTimeout(). This allows
+          // the caller to attach an error handler.
+          setTimeout(() => {
```
alexander-fenster: setImmediate?

dev/src/index.ts
```diff
@@ -1363,130 +1319,70 @@ export class Firestore {
...
    * @param methodName Name of the streaming Veneer API endpoint that
    * takes a request and GAX options.
+   * @param mode Whether this a unidirectional or bidirectional call.
    * @param request The Protobuf request to send.
    * @param requestTag A unique client-assigned identifier for this request.
-   * @param allowRetries Whether this is an idempotent request that can be
-   * retried.
    * @returns A Promise with the resulting read-only stream.
    */
-  readStream(
```
alexander-fenster: Is it a breaking change?

dev/src/index.ts
```diff
@@ -1363,130 +1319,70 @@ export class Firestore {
...
+  requestStream(
     methodName: string,
+    mode: 'unidirectional' | 'bidirectional',
```
alexander-fenster: if it's an externally visible function, I would vote for an enum (but up to you)

dev/system-test/firestore.ts
```diff
@@ -1355,6 +1356,19 @@ describe('Query class', () => {
     });
   });
 
+  it('stream() supports readable[Symbol.asyncIterator]()', async () => {
+    let received = 0;
+    await randomCol.doc().set({foo: 'bar'});
+    await randomCol.doc().set({foo: 'bar'});
+
+    const stream = randomCol.stream();
+    for await (const chunk of stream) {
```
alexander-fenster: Does TypeScript compile it into something that older versions of Node.js can understand? Async iterators only appeared in Node 10, right?

dev/src/index.ts
schmidt-sebastian: I tried this earlier and it broke my unit tests. It works now though, so I updated this call.

dev/src/index.ts
schmidt-sebastian: This is not part of of our public API (we define [our own types](https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts)). I can use an underscore prefix here, but in general the client is pretty inconsistent here.

dev/src/index.ts
schmidt-sebastian: This should be private.

dev/system-test/firestore.ts
schmidt-sebastian: I am a little surprised too, but it seems to work. I stole this test from https://github.com/googleapis/nodejs-storage/pull/799

schmidt-sebastian: Typescript transpiles it away:

```
  it('stream() supports readable[Symbol.asyncIterator]()', async () => {
        var e_1, _a;
        let received = 0;
        await randomCol.doc().set({ foo: 'bar' });
        await randomCol.doc().set({ foo: 'bar' });
        const stream = randomCol.stream();
        try {
            for (var stream_1 = __asyncValues(stream), stream_1_1; stream_1_1 = await stream_1.next(), !stream_1_1.done;) {
                const chunk = stream_1_1.value;
                ++received;
            }
        }
        catch (e_1_1) { e_1 = { error: e_1_1 }; }
        finally {
            try {
                if (stream_1_1 && !stream_1_1.done && (_a = stream_1.return)) await _a.call(stream_1);
            }
            finally { if (e_1) throw e_1.error; }
        }
        chai_1.expect(received).to.equal(2);
    });
```

schmidt-sebastian: Also verifies that the test does indeed work with Symbol.asyncIterator by changing the target level to ES2019.

dev/src/index.ts
alexander-fenster: OK. Since it's TypeScript, we might as well make those methods `private`.

schmidt-sebastian: Unfortunately, these are "module private" and used in other internal classes.

PR #845, opened 2019-12-30
fix: use rejected Promise for terminate()

schmidt-sebastian:
This allows us to catch the rejected Promise, rather than having it show up in test log as an unhandled exception.


schmidt-sebastian:
Sending this to @wilhuff  since @thebrianchen is still OOO. This PR also includes the fixes + pending feedback for https://github.com/googleapis/nodejs-firestore/pull/833

PR #846, opened 2019-12-30
fix: increase test timeout

schmidt-sebastian:
This should make our CI less flaky.

### DIFF
--- a/dev/test/index.ts
+++ b/dev/test/index.ts
@@ -1073,7 +1073,7 @@ describe('getAll() method', () => {
         expect(actualErrorAttempts).to.deep.eq(expectedErrorAttempts);
       });
     });
-  });
+  }).timeout(5000);
 
   it('requires at least one argument', () => {
     return createInstance().then(firestore => {


dev/test/index.ts
```diff
@@ -1073,7 +1073,7 @@ describe('getAll() method', () => {
         expect(actualErrorAttempts).to.deep.eq(expectedErrorAttempts);
       });
     });
-  });
```
alexander-fenster: We normally do `this.timeout(5000);` inside the function, it may be more readable than this, but up to you.

schmidt-sebastian: That doesn't compile for me:

```
test/watch.ts(865,5): error TS7041: The containing arrow function captures the global value of 'this'.
test/watch.ts(865,10): error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.
```

If I want to do this, I have to use `function() {..}` somewhere in my test initialization (e.g. in the `describe` call). Rather than to this, I prefer to leave it as is.

alexander-fenster: Yes, we use `function` instead of arrow-style functions if we need to set `this.timeout()`.  That's OK to leave it as is.

PR #847, opened 2019-12-30
feat: use GAX retry config for streams

schmidt-sebastian:
This PR cleans up our request handling a little bit:
- It removes our internal request retry mechanism for "normal" requests and instead relies on Google Gax to do the retries.
- It changes the retry logic for streaming requests to use the error codes defined in the Service Config (as far as I am aware, Gax does not retry streaming requests)
- It changes the backoff timer for streaming requests to use the backoff timer that Firestore is already using for Watch.
- It drops our own error class in favor of Google Gax's GoogleError.


dev/src/types.ts
```diff
@@ -20,6 +20,123 @@ import {Timestamp} from './timestamp';
 
 import api = google.firestore.v1;
 
+/*!
```
schmidt-sebastian: This block is moved from watch.ts. I eventually want to use the error code definition included in Google Gax.

dev/test/transaction.ts
```diff
@@ -457,21 +459,6 @@ describe('failed transactions', () => {
     ).to.eventually.be.rejectedWith('Final exception');
   });
 
-  it('fails on beginTransaction', () => {
```
schmidt-sebastian: This test no longer works, since the retries are handled in google-gax, which is replaced in the unit tests with dumb implementation that doesn't retry.

dev/src/index.ts
```diff
@@ -1107,62 +1104,55 @@ export class Firestore {
...
           'Retrying request that failed with error:',
-          err
+          lastError
         );
-        return self._retry(attemptsRemaining, requestTag, func, nextDelay);
-      });
+      }
+
+      try {
+        const result = await backoff.backoffAndWait().then(func);
```
alexander-fenster: Mixing `await` and `then` in the same line is kind of confusing. Is it the same as

```ts
await backoff.backoffAndWait();
this._lastSuccessfulRequest = new Date().getTime();
return await func();
```

dev/src/types.ts
alexander-fenster: Gosh, those status codes spread everywhere.

In `google-gax`, we already have `src/status.ts` that basically copies the error codes enum from gRPC. It's not exported though. If we export it from gax, we can reuse it here.

dev/src/util.ts
```diff
@@ -92,3 +96,24 @@ export function isEmpty(value: {}): boolean {
...
+ * RPC.
+ *
+ * @private
+ */
+export function isPermanentRpcError(
+  err: GoogleError,
+  methodName: string,
+  config: ClientConfig
+): boolean {
+  const serviceConfig = config.interfaces!['google.firestore.v1.Firestore']!;
```
alexander-fenster: Just a heads up, this config file will go away soon. By "soon" here I mean half-a-year-ish time. When backend starts supporting returning this configuration via DNS (TXT record I guess), we'll start using it and will stop generating those JSON configs.

dev/src/index.ts
schmidt-sebastian: It's not quite (the successful request should only be set after `await func()`). I cleaned it up.

dev/src/types.ts
schmidt-sebastian: I am hoping that https://github.com/googleapis/gax-nodejs/pull/693 will solve this.

dev/src/util.ts
schmidt-sebastian: Got it. I will find some clever way to get this configuration in 6 months from now.

Or rather - I will find a clever way to push this code into GAX.

dev/src/types.ts
alexander-fenster: https://github.com/googleapis/gax-nodejs/pull/693 is merged and a new version will be released on January 2nd. Thank you!

schmidt-sebastian: Can you also merge https://github.com/googleapis/gax-nodejs/pull/695? Thanks!

alexander-fenster: Done! Release is expected to be done tomorrow in the morning.

PR #849, opened 2020-01-01
refactor: use typed GAPIC methods

schmidt-sebastian:
This PR adds a typed GAPIC layer, but instead of copying the .d.ts files post build, I am using a manually created interface.

The PR still deletes "fake_certificate" and converts all GAPIC usage to Promises. We also continue to lazy-load the client.


schmidt-sebastian:
On import, this currently fails with:

"Exports and export assignments are not permitted in module augmentations."

dev/src/types.ts
```diff
@@ -27,9 +30,62 @@ export interface ApiMapValue {
...
+  ): Duplex;
+  runQuery(request?: api.IRunQueryRequest, options?: CallOptions): Duplex;
+  listDocuments(
+    request: api.IListDocumentsRequest,
+    options?: CallOptions
+  ): Promise<[api.IDocument[], unknown, unknown]>;
+  listCollectionIds(
+    request: api.IListCollectionIdsRequest,
+    options?: CallOptions
+  ): Promise<[string[], unknown, unknown]>;
```
schmidt-sebastian: FYI: I tried to use `Promise<[T, undefined, undefined]>` here, but the paginated methods and the non-paginated methods seem to disagree on return type (undefined vs null).

dev/src/types.ts
```diff
@@ -27,9 +30,62 @@ export interface ApiMapValue {
...
+    options?: CallOptions
+  ): Promise<[api.ICommitResponse, unknown, unknown]>;
+  rollback(
+    request: api.IRollbackRequest,
+    options?: CallOptions
+  ): Promise<[google.protobuf.IEmpty, unknown, unknown]>;
+  batchGetDocuments(
+    request?: api.IBatchGetDocumentsRequest,
+    options?: CallOptions
+  ): Duplex;
```
schmidt-sebastian: I am just returning a Duplex stream since I my unit tests do not create a Cancellable stream.

dev/src/types.ts
alexander-fenster: Oh, this is something to be fixed in gax (later).

PR #851, opened 2020-01-02
build: make update-license script work on Windows

schmidt-sebastian:
This PR changes the update-license script to be a Node JS script instead of bash. 

The intent is to make the script work on Windows so it can be used as part of a larger PR that will export the generated types for the GAPIC client (https://github.com/googleapis/nodejs-firestore/pull/849)

This PR also contains the output from running the script.


dev/protos/firestore_v1_proto_api.js
```diff
@@ -7428,4 +7428,4 @@ $root.google = (function() {
     })();
 
     return google;
-})();
+})();
```
alexander-fenster: Can we have new lines back? (here and everywhere)

scripts/license.js
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+`;
+
+for (const file of process.argv.slice(2)) {
+    const content = fs.readFileSync(file);
+    fs.writeFileSync(file, `${LICENSE_HEADER}\n${content}`);
```
alexander-fenster: maybe just add `\n` here

schmidt-sebastian:
Line breaks added.

PR #853, opened 2020-01-03
build: add typings for generated client

schmidt-sebastian:
This PR abuses the TypeScript build pipeline to add the generated firestore_client.d.ts file to our public types.  It copies the result from the transpile step into the /src folder. This allows us to use a typed GAPIC layer without having to directly import the client (to avoid loading the GAX stack at startup).

To make the Protobuf imports work in the generated types, I move the /type folder under /src. I also removed the fake-certificate test hack, as I no longer see any warnings. This reduces the complexity of the "compile" step a tiny bit.

This is also part of the overall change in https://github.com/googleapis/nodejs-firestore/pull/849 (which is not to be submitted, but may provide some contents).

### DIFF
--- a/dev/test/fake-certificate.json
+++ b/dev/test/fake-certificate.json
@@ -1,8 +1,0 @@
-{
-  "type": "service_account",
-  "project_id": "...",
-  "private_key_id": "...",
-  "private_key": "...",
-  "client_email": "...",
-  "client_id": "..."
-}
--- a/dev/test/index.ts
+++ b/dev/test/index.ts
@@ -47,7 +47,6 @@ const DATABASE_ROOT = `projects/${PROJECT_ID}/databases/(default)`;
 const DEFAULT_SETTINGS = {
   projectId: PROJECT_ID,
   sslCreds: grpc.credentials.createInsecure(),
-  keyFilename: __dirname + '/fake-certificate.json',
 };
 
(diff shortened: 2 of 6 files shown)


schmidt-sebastian:
The code coverage drop is this PR will go away once https://github.com/googleapis/synthtool/pull/362 is pushed everywhere.

alexander-fenster:
This is scary. I'll summon @JustinBeckwith who knows much more TypeScript than I do to review this :) 

I'm thinking of maybe this is not the way how this problem should be solved. It _seems_ like TypeScript supports dynamic `import()`, maybe we should look into that and make gax load when needed (from the client constructor)? I have never tried that but can look into that.

schmidt-sebastian:
> I'm thinking of maybe this is not the way how this problem should be solved. It _seems_ like TypeScript supports dynamic `import()`, maybe we should look into that and make gax load when needed (from the client constructor)? I have never tried that but can look into that.

We are already only importing the client when needed. The problem is the type definition, which is attached to a top-level class member (https://github.com/googleapis/nodejs-firestore/pull/849/files#diff-4de362b91cfe3cd8254c447a5a08b316R275). I hope there is some better way to do this than what I proposed, but I looked at dynamic imports and it only seems to solve part of the problem. Here is to hoping for @JustinBeckwith :)

Issue #829, opened 2019-12-19
GRPC related crash when committing batch updates


jakeleventhal:
#### Environment details

  - OS: OS: Mac OS 10.15 Beta (19A573a)
  - Node.js version: v12.10.0
  - npm version: 6.10.3
  - `@google-cloud/firestore` version: 3.0.0

#### Steps to reproduce

  1. Create several batch updates and commit them all at once:

```javascript
const batches =[b1, b2, b3, b4, b5];
await Promise.all(batches.map(async (b) => b.commit()));
```

I then get these errors (sometimes but not always)
<img width="1590" alt="Screen Shot 2019-12-18 at 8 52 10 PM" src="https://user-images.githubusercontent.com/15133828/71146136-e5c88e00-21d8-11ea-8b06-8b09ba60b4c0.png">

jakeleventhal:
<img width="1612" alt="Screen Shot 2019-12-18 at 9 22 55 PM" src="https://user-images.githubusercontent.com/15133828/71147149-ae5be080-21dc-11ea-845c-26f2d4fbe755.png">
Interestingly, it starts with this

schmidt-sebastian:
@jakeleventhal Thanks for reporting this. Are you able to share client logs with us? I have been unable to reproduce this locally for now. The logs should offer some insight into the order of operations that causes this.

murgatroid99:
That metadata error is definitely unrelated. That is just a warning that is logged to the console.

The "Channel has been shut down" error comes from grpc-js, and it is only thrown if `Channel#close` or `Client#close` (which calls `Channel#close`) is called by the user (either google-gax or firestore in this case).

jakeleventhal:
@murgatroid99 is the warning due to something on my end?

@schmidt-sebastian the only logs i have are screenshotted in this issue

jakeleventhal:
As this is a fatal error that crashes my server, I have to revert to firestore 2.3.0 (which still has the memory leak issue resolved in 3.0.0 with #768)

murgatroid99:
I think that the issue about it, which you posted in, should explain it. But in short, no, we don't know what causes the header corruption, but at this time we have no reason to believe that it's related to anything the user is doing.

schmidt-sebastian:
I spent another couple hours debugging this. I believe this is an issue in our library, but I am not able to reproduce this. We can re-open this we get a reproduction with Firestore SDK logs (via `setLogFunction()`).

[...]

PR #870, opened 2020-01-10
fix: do not release client before retry

schmidt-sebastian:
Addresses https://github.com/googleapis/nodejs-firestore/issues/829

I think there is a potential problem with stream retry in the existing client. I wasn't able to write a test that hit the scenario, but I can try to put it into words.

In the existing code, requestStream() takes a client object out of the client pool and then calls retry(). There are two Promises that manage all the logic: Result and Lifetime. Since there are separate, it is possible that result.reject() is called (causing a retry) just before lifetime.resolve() is called, which returns the client back to the pool. We then retry with a client that was already returned to the pool, which - if the use is unlucky - is a client that was already garbage collected.

The change in this PR reverses the order: Each retry attempt fetches its own client from the pool, and thus the lifetime is now bound to the length of a single retry attempt.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1173,6 +1173,8 @@ export class Firestore {
    *
    * @private
    * @param backendStream The Node stream to monitor.
+   * @param lifetime A Promise that resolves when the stream receives an 'end',
+   * 'close' or 'finish' message.
    * @param requestTag A unique client-assigned identifier for this request.
    * @param request If specified, the request that should be written to the
    * stream after opening.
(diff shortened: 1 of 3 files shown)


dev/src/index.ts
```diff
@@ -1181,6 +1183,7 @@ export class Firestore {
    */
   private _initializeStream(
     backendStream: Duplex,
+    liftime: Deferred<void>,
```
alexander-fenster: typo, `liftime` vs `lifetime`?

schmidt-sebastian: fixed

PR #875, opened 2020-01-13
refactor: use loop for transactions

schmidt-sebastian:
Our transaction retry is going to get more complicated (since we have to manually RETRY on aborted), so I thought it would be a good exercise to simplify the transaction runner.

This PR:
- moves the runner out of index.ts to allow access to private state (and reduce index.ts file size)
- converts the recursive approach to an iterative loop similar to https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L1134

I was hoping there would be no behavior changes, but since we are now using an `async` function, the `try` block around `updateFunction()` now catches the "No read after write" error, which means that we will now rollback these transactions (which are empty).


dev/src/transaction.ts
```diff
@@ -61,28 +62,17 @@ const READ_AFTER_WRITE_ERROR_MSG =
 export class Transaction {
   private _firestore: Firestore;
   private _writeBatch: WriteBatch;
-  private _requestTag: string;
+  private _requestTag?: string;
```
mikelehen: It's not immediately obvious to me why this can't be passed into the constructor?

And, er, on that note, I can't figure out where this is ever assigned now?  Seems like it'll always be undefined.

schmidt-sebastian: It should have been assigned in `runTransaction()`. This does lead to a bunch of `?` and `!` in the code though, so I happily moved it back to the constructor.

schmidt-sebastian:
@thebrianchen Since this is pretty fragile code, do you mind looking at this as well?

dev/test/write-batch.ts
```diff
@@ -356,6 +356,25 @@ describe('batch support', () => {
     return promise;
   });
 
+  it('cannot reset a committed batch', async () => {
```
thebrianchen: s/cannot/can

Issue #753, opened 2019-09-07
Moment.js does not work for date queries


jakeleventhal:
#### Environment details

  - OS: Linux b8b01391ee38 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux (this is from the node:latest docker image)
  - Node.js version: 12.9.0
  - `@google-cloud/firestore` version: 2.2.8

#### Steps to reproduce

  1. Construct a query on a date field field and pass a moment.js object in as the filter value
```javascript
firestore.collection('Users').where('JoinedDate', '>=', moment().subtract(1, 'day'))
```
  2. You will get an error that looks like this:
```
Error: Value for argument "value" is not a valid Firestore document. Couldn't serialize object of type "Moment". Firestore doesn't support JavaScript objects with custom prototypes (i.e. objects that were created via the "new" operator).
```

Moment.js is such a popular library used for dates that i think an exception should be made for this.

schmidt-sebastian:
@jakeleventhal Sorry to hear about your trouble.

For now, we only special-case JavaScript's Date (as well as our own Timestamp object). You can convert a `moment()` date to a JS Date via `moment().toDate()`. We might consider native support in the future, but would likely need more user feedback/demand before doing so.

jakeleventhal:
Yeah, we currently do that. It just seems like an unnecessary step. Wouldn't the fix be as simple as adding the following to the where clause check? I was going to make the PR myself but i didn't want  to waste my breath if you guys aren't okay with it.

```javascript
if (typeof val === 'object' && val.constructor.name === 'Moment') {
   val = val.toDate();
}
```

schmidt-sebastian:
I talked to some of the folks internally, and we are a little hesitant about adding first party support, especially given existing proposals to add better time management to the Node standard library (such as https://github.com/tc39/proposal-temporal). We can still keep this open as a feature request, which will allow us to gather more feedback.

jakeleventhal:
Right, @bcoe the sample code i provided above is all that you would need. Just a check on the string type to see if the object is of type "Moment"

PR #879, opened 2020-01-14
feat: support serialization of Moment.js

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/753

Note that this only adds a dev dependency on the MomentJs types, but not MomentJs itself.


dev/src/serializer.ts
```diff
@@ -369,6 +381,8 @@ export function validateUserInput(
     // Ok.
   } else if (value instanceof Timestamp || value instanceof Date) {
     // Ok.
+  } else if (isMomentJsType(value)) {
```
thebrianchen: Just a thought -- is there value in switching the logic here to use switch/case instead of if/else blocks?

dev/test/document.ts
```diff
@@ -244,6 +244,37 @@ describe('serialize document', () => {
...
+            }),
+          })
+        );
+        return response(writeResult(1));
+      },
+    };
+
+    return createInstance(overrides).then(firestore => {
+      return firestore.doc('collectionId/documentId').set({
+        moonLanding: new Moment(),
```
thebrianchen: indeed, what a moment that was

dev/src/serializer.ts
schmidt-sebastian: As discussed, this doesn't make sense here since we mix "instanceof" with "typeof" among other things.

dev/test/document.ts
schmidt-sebastian: All the special effects!

willbattel:
Just to be clear: this PR allows us to pass moment objects as field values with set/update operations, but does not allow querying for Timestamp values using moment objects. For the latter we still need to use `toDate` on the moment objects.

Is this correct?

schmidt-sebastian:
@willbattel The serialization logic is the same in all parts of the API. You should be able to use MomentJs types as query constraints.

Issue #827, opened 2019-12-18
Firestore client does not retry on contention errors


mad-it:
#### Current behavior

We are using Google cloud functions in combination with Firestore. When multiple, in this example 2, cloud function instances attempt to process different data on the same documents within a transaction 1 of the instances throws a contention error. This is inline with what is documented [here](https://cloud.google.com/firestore/docs/best-practices).

#### Expected behavior

Instance A & B both read document X inside of a transaction. Instance A updates document X before instance B can commit its changes. The expected behavior would be for the transaction on instance B to be retried because it cannot commit its changes due to the X being a "dirty read".

#### Environment details

  - OS: Ubuntu 19.10
  - Node.js version: 10.18.0
  - npm version:6.12
  - `@google-cloud/firestore` version: 2.6.1

#### Steps to reproduce

Even though we dont have clear reproduction steps, but after some investigation of the code,  the culprit is seems clear.

Looking at [this line](https://github.com/googleapis/nodejs-firestore/blob/22ef0d0229569f0d97ff908b5866264a8de2ca78/dev/src/index.ts#L799) in the code, it looks like contention caused inside of the `updateFunction` for example caused by reading a document, will never trigger a retry.

#### Additional information

Below is a log output of 2 functions "colliding" and no retry being triggered.

The logs have been sanitized by removing all project references, document structures and data being sent over the wire.

<details>
  <summary>Logs </summary>  

  ```
d82a0kxaae6u - Firestore (2.6.1) 2019-12-17T15:41:47.651Z xqgri [Firestore.readStream]: Sending request: {\"database\":\"projects/<OBSCURED>/databases/(default)\",\"transaction\":{\"type\":\"Buffer\",\"data\":[OBSCURED]},\"documents\":[\"projects/<OBSCURED>/databases/(default)/documents/<OBSCURED>\"]}

d82a0kxaae6u - Firestore (2.6.1) 2019-12-17T15:41:47.651Z NA6v7 [Firestore.readStream]: Sending request: {\"parent\":\"projects/<OBSCURED>/databases/(default)/documents/<OBSCURED>\",\"structuredQuery\":{\"from\":[OBSCURED]},\"transaction\":{\"type\":\"Buffer\",\"data\":[OBSCURED]}

[...]

PR #883, opened 2020-01-14
fix: manually retry ABORTED reads in transactions

schmidt-sebastian:
This adds manual retry to our transaction logic similar to what we have in the Web SDK: https://github.com/firebase/firebase-js-sdk/blob/c822e78b00dd3420dcc749beb2f09a947aa4a344/packages/firestore/src/core/transaction_runner.ts#L112

We only retry GRPC exceptions, and only failures with code ABORTED. This is because:
- we don't want to retry if the user throws an error or returns a rejected Promise
- GAPIC retries reads that fail with UNAVAILABLE, INTERNAL and DEADLINE_EXCEEDED under the hood

Fixes https://github.com/googleapis/nodejs-firestore/issues/827 (fingers crossed)

### DIFF
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -565,7 +565,7 @@ function validateReadOptions(
 function isRetryableTransactionError(error: Error): boolean {
   if (error instanceof GoogleError || 'code' in error) {
     // In transactions, the backend returns code ABORTED for reads that fail
-    // due to contention. These errors should be retried for both GoogleError
+    // with contention. These errors should be retried for both GoogleError
     // and GoogleError-alike errors (in case the prototype hierarchy gets
     // stripped somewhere).
     return error.code === Status.ABORTED;
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import {GoogleError, Status} from 'google-gax';
+
 import * as proto from '../protos/firestore_v1_proto_api';
 
 import {DocumentSnapshot, Precondition} from './document';
@@ -434,8 +436,15 @@ export class Transaction {
           'Rolling back transaction after callback error:',
           err
         );
+
         await this.rollback();
-        return Promise.reject(err); // User callback failed
+
+        if (isRetryableTransactionError(err)) {
+          lastError = err;
+          continue; // Retry full transaction
+        } else {
+          return Promise.reject(err); // Callback failed w/ non-retryable error
+        }
       }
 
       try {
@@ -450,7 +459,7 @@ export class Transaction {
     logger(
       'Firestore.runTransaction',
       this._requestTag,
-      'Exhausted transaction retries, returning error: %s',
+      'Transaction not eligible for retry, returning error: %s',
       lastError
     );
     return Promise.reject(lastError);
@@ -551,4 +560,15 @@ function validateReadOptions(
       }
     }
   }
+}
+
+function isRetryableTransactionError(error: Error): boolean {
+  if (error instanceof GoogleError || 'code' in error) {
+    // In transactions, the backend returns code ABORTED for reads that fail
+    // due to contention. These errors should be retried for both GoogleError
+    // and GoogleError-alike errors (in case the prototype hierarchy gets
+    // stripped somewhere).
+    return error.code === Status.ABORTED;
+  }
+  return false;
 }
--- a/dev/test/transaction.ts
+++ b/dev/test/transaction.ts
@@ -106,7 +106,7 @@ function commit(
 
 function rollback(
   transaction?: Uint8Array | string,
-  err?: Error
+  error?: Error
 ): TransactionStep {
   const proto = {
     database: DATABASE_ROOT,
@@ -116,7 +116,7 @@ function rollback(
   return {
     type: 'rollback',
     request: proto,
-    error: err,
+    error,
     response: {},
   };
 }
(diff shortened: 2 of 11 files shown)


dev/test/transaction.ts
```diff
@@ -106,7 +106,7 @@ function commit(
 
 function rollback(
   transaction?: Uint8Array | string,
-  err?: Error
+  error?: Error
```
thebrianchen: optional: change naming for `commit()` as well so they're all in sync ([link](https://github.com/googleapis/nodejs-firestore/blob/8f41b37fd766b18113401c7324e319b52b38cc96/dev/test/transaction.ts#L73))

dev/test/transaction.ts
```diff
@@ -416,14 +424,51 @@ describe('failed transactions', () => {
...
+        await transaction.get(docRef);
+        return 'success';
+      },
+      begin('foo1'),
+      getDocument('foo1', retryableError),
+      rollback('foo1'),
+      begin('foo2', 'foo1'),
+      getDocument('foo2'),
+      commit('foo2')
+    ).then(red => {
```
thebrianchen: s/red/res

dev/test/transaction.ts
```diff
@@ -416,14 +424,51 @@ describe('failed transactions', () => {
...
+
+  it("doesn't retry custom user exceptions in callback", () => {
     return expect(
       runTransaction(
         () => {
           return Promise.reject('request exception');
         },
         begin(),
-        rollback('foo')
+        rollback()
```
thebrianchen: should we make it consistent throughout [elsewhere](https://github.com/googleapis/nodejs-firestore/blob/8f41b37fd766b18113401c7324e319b52b38cc96/dev/test/transaction.ts#L399-L400)

dev/test/transaction.ts
```diff
@@ -416,14 +424,51 @@ describe('failed transactions', () => {
...
+    const nonRetryableError = new GoogleError('Failed Precondition');
+    nonRetryableError.code = Status.FAILED_PRECONDITION;
+
+    return expect(
+      runTransaction(
+        async (transaction, docRef) => {
+          await transaction.get(docRef);
+          return 'failure';
+        },
+        begin('foo'),
```
thebrianchen: If 'foo' is default, do we need it here as well, or can it also be removed?

dev/src/transaction.ts
```diff
@@ -552,3 +561,14 @@ function validateReadOptions(
     }
   }
 }
+
+function isRetryableTransactionError(error: Error): boolean {
+  if (error instanceof GoogleError || 'code' in error) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, that is JavaScript functionality (see https://masteringjs.io/tutorials/fundamentals/hasownproperty for a comparison against `Object.hasOwnProperty`). I am doing it here because I am afraid that some of our dependencies might throw a GoogleError that is not actually a GoogleError by providing just a code and a message field.

dev/test/transaction.ts
schmidt-sebastian: Done. Good catch.

dev/test/transaction.ts
schmidt-sebastian: Fixed (here and the test I copied it from)

dev/test/transaction.ts
schmidt-sebastian: I left it as is here since `getDocument` uses the second argument. I prefer this over only specifying the transaction ID once or using `undefined` in its place.

dev/test/transaction.ts
schmidt-sebastian: Fixed

dev/src/transaction.ts
```diff
@@ -552,3 +561,14 @@ function validateReadOptions(
...
+    // In transactions, the backend returns code ABORTED for reads that fail
+    // with contention. These errors should be retried for both GoogleError
+    // and GoogleError-alike errors (in case the prototype hierarchy gets
```
novemberborn: Could this check for `error.name === 'GoogleError'` instead?

When implementing gRPC servers, it's not unthinkable that an ABORTED error is thrown from inside the transaction, with the expectation that it's returned to the client.

schmidt-sebastian: This should accept any object that has a "code" property. In the environment you are describing, will this property be stripped?

novemberborn: Let me rephrase.

There are many ways of wrongly implementing a transaction function, to do with side-effects, or forgetting to use the transaction itself to perform queries.

This new `ABORTED` handling adds more ways to get it wrong.

There is an assumption that all `GoogleError` instances came from transaction methods. That may not be the case if I use another GCP SDK within my function. I know, I shouldn't be, but somebody still might.

It also won't be the case if I call a Firestore method on a `DocumentReference` which then throws an `ABORTED` error.

These assumptions are then generalized by testing if the thrown error has a `.code = 10` property. I have a `ServiceError` object that is thrown with a gRPC response code. If for some reason I throw that error inside a transaction function, my transaction may be retried even if it succeeded. And again, sure, I suppose I shouldn't be throwing that error, but somebody still might.

A safer approach may be to register errors, thrown as part of the actual transaction, in a `WeakMap`. Then as they propagate through user code, when you catch them again you can check whether they're genuine transaction errors, and then check their code.

schmidt-sebastian: A couple things to note:
- A read outside of a transaction should never result in Code ABORTED. This error should only happen during a transactional read.
- If a user throws an error with `code: 10`, we will retry the transaction. In general, transaction functions should be written so they can be re-run, and re-running them more often than needed should not cause any harm.

I understand that we control the errors that we are catching here. We could tighten this check to make sure that the error is indeed a GoogleError, but for now I decided to err on the side of caution and also retry GoogleError-alike errors. This allows us to catch errors with code properties that were created by our dependencies before we converted everything to use the proper GoogleError type.

PR #884, opened 2020-01-14
refactor: remove usage of 'self'

schmidt-sebastian:


schmidt-sebastian:
> I think there's one more instance of `self = this` [here](https://github.com/googleapis/nodejs-firestore/blob/94ddc897400cafe5a1ee16f3ad0d285411bdd0b2/dev/src/reference.ts#L1914) as well.

Yeah, that is still needed since we use `this.push()` in the through2 context. Going away with https://github.com/googleapis/nodejs-firestore/pull/886

PR #886, opened 2020-01-15
blocked: fix(deps): drop dependency on through2

schmidt-sebastian:
I believe that this will now pass CI since I dropped the dependency on  over the holidays and reworked our error piping. Fingers crossed

Follow up to https://github.com/googleapis/nodejs-firestore/pull/618

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
"I believe that this will now pass CI"

I was wrong. Let's wait till we can drop Node 8.

JustinBeckwith:
I've been trying to slay this dependency for like 2 years man, god speed

PR #897, opened 2020-01-24
fix: retry streaming methods if initial write errored

schmidt-sebastian:
According to https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback, the write() callback may be called with an Error before on('error') is invoked. In this case, we should reject our Stream. In the current implementation, we would accept the Stream as healthy and then forward the error to the user, which prevents retries.

I noticed while looking at this code as we are porting it to Python


schmidt-sebastian:
FYI: There are no tests for now, since mocking this is pretty difficult. Feel free to push back :)

PR #912, opened 2020-02-05
feat: add google-gax status to exports

### LINKED ISSUE #902: Reexport Status from google-gax
I'm trying to handle NOT_FOUND and ALREADY_EXISTS errors with the following code:

```typescript
...

async function updateExample() : Promise<boolean> {
  let col = db.col('examples');
  let doc = col.doc('id');

  try {
    await db.update({
      newData: 'new data',
    });

    return true;
  } catch (err) {
    switch (err.code) {
    case 'not-found':
      return false;
    default:
      throw err;
    }
  }
}
```

But, as it turns out, the error codes come straight from the `google-gax` package and they are defined as numbers in the `Status` enum, instead of strings.

I can solve this with a simple
```typescript
import { Status } from 'google-gax';

...

async function updateExample() : Promise<boolean> {
  let col = db.col('examples');
  let doc = col.doc('id');

  try {
    await db.update({
      newData: 'new data',
    });

    return true;
  } catch (err) {
    switch (err.code) {
    case Status.NOT_FOUND:
      return false;
    default:
      throw err;
    }
  }
}
```
but now I'm importing a transitive dependency that is not defined in my `package.json` and is clearly an implementation detail.

It would be very handy if this package could reexport that enum or provide a thin translation layer between errors.

thebrianchen:
Fixes #902

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -1452,6 +1452,12 @@ declare namespace FirebaseFirestore {
    * API and the underlying Firestore v1 RPCs.
    */
   export const v1: {FirestoreClient: any, FirestoreAdminClient: any};
+
+  /**
+   * Status error code enums used by google-gax that correspond to the error
+   * codes returned by Firestore calls.
+   */
+  export const Status: import('../dev/src/types').Status;
 }
 
 declare module '@google-cloud/firestore' {
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -1457,7 +1457,25 @@ declare namespace FirebaseFirestore {
    * Status error code enums used by google-gax that correspond to the error
    * codes returned by Firestore gRPC calls.
    */
-  export const GrpcStatus: import('google-gax').Status;
+  export enum GrpcStatus {
+    OK = 0,
+    CANCELLED = 1,
+    UNKNOWN = 2,
+    INVALID_ARGUMENT = 3,
+    DEADLINE_EXCEEDED = 4,
+    NOT_FOUND = 5,
+    ALREADY_EXISTS = 6,
+    PERMISSION_DENIED = 7,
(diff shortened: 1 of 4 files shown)


types/firestore.d.ts
```diff
@@ -1452,6 +1452,12 @@ declare namespace FirebaseFirestore {
    * API and the underlying Firestore v1 RPCs.
    */
   export const v1: {FirestoreClient: any, FirestoreAdminClient: any};
+
+  /**
+   * Status error code enums used by google-gax that correspond to the error
+   * codes returned by Firestore calls.
+   */
+  export const Status: import('../dev/src/types').Status;
```
schmidt-sebastian: Two questions:
- Do we need to roundtrip through `src/types`?
- Could we export this type under a different name? `Status` is very generic. It would be better if this was tied to gRPC somehow.

thebrianchen: Removed the roundtrip. I originally thought there was a rule about importing global packages into an export, but that's only inside an import.

Renamed to GrpcStatus.

schmidt-sebastian:
I am also not able to get type completion to work. Look at these two examples:

Screen Shot 2020-02-06 at 1 57 42 PM

Screen Shot 2020-02-06 at 1 58 48 PM

types/firestore.d.ts
```diff
@@ -1452,6 +1452,30 @@ declare namespace FirebaseFirestore {
    * API and the underlying Firestore v1 RPCs.
    */
   export const v1: {FirestoreClient: any, FirestoreAdminClient: any};
+
+  /**
+   * Status error code enums used by google-gax that correspond to the error
```
schmidt-sebastian: Can you drop either "code" or "enums" (and probably google-gax, which is an internal library). My preference would be just "Status codes returned by Firestore's GRPC calls".

PR #915, opened 2020-02-05
fix: wait for operations to complete before deleting clients

### LINKED ISSUE #913: Closing an onSnapshot listener gives UnhandledPromiseRejectionWarnings when followed by client.terminate()
#### Environment details

  - OS: macOS Catalina (`10.15.1`)
  - Node.js version: `10.16.3`
  - npm version: `6.9.0`
  - `@google-cloud/firestore` version: `^3.4.1`

#### Problem

When closing an onSnapshot listener, followed by the termination of the firestore client, an 'UnhandledPromiseRejectionWarning' is shown:

```
onSnapshot { path: 'collection/doc-1', exists: false }
onSnapshot { path: 'collection/doc-1', exists: true }
(node:41140) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: No active request
    at ClientPool.release (/Users/vdvleon/Downloads/reproduction/node_modules/@google-cloud/firestore/build/src/pool.js:95:9)
    at op.catch.then (/Users/vdvleon/Downloads/reproduction/node_modules/@google-cloud/firestore/build/src/pool.js:162:24)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:41140) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:41140) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

#### Steps to reproduce

1. Open an onSnapshot listener (on either documents or queries) and store the 'close' callback.
   `const close = documentReference.onSnapshot(...);`
2. Call the 'close' callback.
   `close();`
3. Terminate the firestore client.
    `firestore.terminate();`

_Relevant code:_
```typescript
const documentReference = firestore.collection("collection").doc("doc-1");

const close = documentReference.onSnapshot((snapshot: DocumentSnapshot) => {
  console.log("onSnapshot", {
    path: snapshot.ref.path,
    exists: snapshot.exists
  });
});

const run = async (): Promise<void> => {
  await clearFirestoreData();

  await documentReference.set({});

  close();

  await firestore.terminate();
};
```

_The full reproduction can be found here:_
[reproduction.zip](https://github.com/googleapis/nodejs-firestore/files/4158731/reproduction.zip)

**To run the code:**

```bash
cd reproduction
npm install
npx ts-node index.ts
```

thebrianchen:
Fixes #913 

Implemented the 2nd suggestion. I think it makes more sense to allow for existing operations to complete before calling terminate. In the user's repro it would be strange to have `close()` throw an error.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -995,22 +995,22 @@ export class Firestore {
   }
 
   /**
-   * Registers the requestTag associated with the listener. This is used to
-   * ensure that all listeners are detached when terminate() is called.
+   * Registers a listener on this client, incrementing the listener count. This
+   * is used to ensure that all listeners are unsubscribed when terminate() is
+   * called.
    *
    * @private
-   * @param requestTag A unique client-assigned identifier
    */
   registerListener(): void {
     this.registeredListenersCount += 1;
   }
 
   /**
-   * Unregisters the requestTag associated with the listener. This is used to
-   * ensure that all listeners are detached when terminate() is called.
+   * Unregisters a listener on this client, decrementing the listener count.
+   * This is used to ensure that all listeners are unsubscribed when terminate()
+   * is called.
    *
    * @private
-   * @param requestTag A unique client-assigned identifier
    */
   unregisterListener(): void {
     this.registeredListenersCount -= 1;
@@ -1024,7 +1024,7 @@ export class Firestore {
   terminate(): Promise<void> {
     if (this.registeredListenersCount > 0) {
       return Promise.reject(
-        'All listeners must be disconnected before terminating the client.'
+        'All onSnapshot() listeners must be unsubscribed before terminating the client.'
       );
     }
     return this._clientPool.terminate();
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -298,6 +298,13 @@ export class Firestore {
    */
   private _projectId: string | undefined = undefined;
 
+  /**
+   * Set of requestTags associated with the listeners that have been registered
+   * on the client.
+   * @private
+   */
+  private registeredListeners = new Set<string>();
+
   // GCF currently tears down idle connections after two minutes. Requests
   // that are issued after this period may fail. On GCF, we therefore issue
   // these requests as part of a transaction so that we can safely retry until
@@ -988,11 +995,38 @@ export class Firestore {
   }
 
   /**
+   * Registers the requestTag associated with the listener. This is used to
+   * ensure that all listeners are detached when terminate() is called.
+   *
+   * @private
+   * @param requestTag A unique client-assigned identifier
+   */
+  registerListener(requestTag: string): void {
+    this.registeredListeners.add(requestTag);
+  }
+
+  /**
+   * Unregisters the requestTag associated with the listener. This is used to
+   * ensure that all listeners are detached when terminate() is called.
+   *
+   * @private
+   * @param requestTag A unique client-assigned identifier
+   */
+  unregisterListener(requestTag: string): void {
+    this.registeredListeners.delete(requestTag);
+  }
+
+  /**
    * Terminates the Firestore client and closes all open streams.
    *
    * @return A Promise that resolves when the client is terminated.
    */
   terminate(): Promise<void> {
+    if (this.registeredListeners.size > 0) {
+      return Promise.reject(
+        'All listeners must be disconnected before terminating the client.'
+      );
+    }
     return this._clientPool.terminate();
   }
 
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -299,8 +299,9 @@ export class Firestore {
   private _projectId: string | undefined = undefined;
 
   /**
-   * Set of requestTags associated with the listeners that have been registered
-   * on the client.
+   * Count of listeners that have been registered on the client.
+   *
+   * The client can only be terminated when there are no registered listeners.
    * @private
    */
   private registeredListenersCount = 0;
@@ -996,7 +997,7 @@ export class Firestore {
 
   /**
    * Registers a listener on this client, incrementing the listener count. This
-   * is used to ensure that all listeners are unsubscribed when terminate() is
+   * is used to verify that all listeners are unsubscribed when terminate() is
    * called.
    *
    * @private
@@ -1007,7 +1008,7 @@ export class Firestore {
 
   /**
    * Unregisters a listener on this client, decrementing the listener count.
-   * This is used to ensure that all listeners are unsubscribed when terminate()
+   * This is used to verify that all listeners are unsubscribed when terminate()
    * is called.
    *
    * @private
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -303,7 +303,7 @@ export class Firestore {
    * on the client.
    * @private
    */
-  private registeredListeners = new Set<string>();
+  private registeredListenersCount = 0;
 
   // GCF currently tears down idle connections after two minutes. Requests
   // that are issued after this period may fail. On GCF, we therefore issue
@@ -1001,8 +1001,8 @@ export class Firestore {
    * @private
    * @param requestTag A unique client-assigned identifier
    */
-  registerListener(requestTag: string): void {
-    this.registeredListeners.add(requestTag);
+  registerListener(): void {
+    this.registeredListenersCount += 1;
   }
 
   /**
@@ -1012,8 +1012,8 @@ export class Firestore {
    * @private
    * @param requestTag A unique client-assigned identifier
    */
-  unregisterListener(requestTag: string): void {
-    this.registeredListeners.delete(requestTag);
+  unregisterListener(): void {
+    this.registeredListenersCount -= 1;
   }
 
   /**
@@ -1022,7 +1022,7 @@ export class Firestore {
    * @return A Promise that resolves when the client is terminated.
    */
   terminate(): Promise<void> {
-    if (this.registeredListeners.size > 0) {
+    if (this.registeredListenersCount > 0) {
       return Promise.reject(
         'All listeners must be disconnected before terminating the client.'
       );
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -44,7 +44,7 @@ export class ClientPool<T> {
 
   /**
    * Deferred promise that is resolved when there are no active operations on
-   * the client pool.
+   * the client pool after terminate() has been called.
    */
   private terminateDeferred = new Deferred<void>();
 
@@ -120,7 +120,7 @@ export class ClientPool<T> {
     const requestCount = this.activeClients.get(client) || 0;
     assert(requestCount > 0, 'No active requests');
     this.activeClients.set(client, requestCount - 1);
-    if (this.opCount === 0) {
+    if (this.terminated && this.opCount === 0) {
       this.terminateDeferred.resolve();
     }
 
@@ -207,7 +207,6 @@ export class ClientPool<T> {
 
     // Wait for all pending operations to complete before terminating.
     if (this.opCount > 0) {
-      this.terminateDeferred = new Deferred<void>();
       logger(
         'ClientPool.terminate',
         /* requestTag= */ null,
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -46,7 +46,7 @@ export class ClientPool<T> {
    * Deferred promise that is resolved when there are no active operations on
    * the client pool.
    */
-  private operationsCompletedDeferred = new Deferred<void>();
+  private terminateDeferred = new Deferred<void>();
 
   /**
    * @param concurrentOperationLimit The number of operations that each client
@@ -121,7 +121,7 @@ export class ClientPool<T> {
     assert(requestCount > 0, 'No active requests');
     this.activeClients.set(client, requestCount - 1);
     if (this.opCount === 0) {
-      this.operationsCompletedDeferred.resolve();
+      this.terminateDeferred.resolve();
     }
 
     if (this.shouldGarbageCollectClient(client)) {
@@ -207,13 +207,14 @@ export class ClientPool<T> {
 
     // Wait for all pending operations to complete before terminating.
     if (this.opCount > 0) {
-      this.operationsCompletedDeferred = new Deferred<void>();
+      this.terminateDeferred = new Deferred<void>();
       logger(
         'ClientPool.terminate',
         /* requestTag= */ null,
-        'Waiting for all pending operations to complete before terminating'
+        'Waiting for %s pending operations to complete before terminating',
+        this.opCount
       );
-      await this.operationsCompletedDeferred.promise;
+      await this.terminateDeferred.promise;
     }
     for (const [client, _requestCount] of this.activeClients) {
       this.activeClients.delete(client);
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -42,6 +42,10 @@ export class ClientPool<T> {
    */
   private terminated = false;
 
+  /**
+   * Deferred promise that is resolved when there are no active operations on
+   * the client pool.
+   */
   private operationsDeferred = new Deferred<void>();
 
   /**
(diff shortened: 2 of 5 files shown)


dev/src/pool.ts
```diff
@@ -112,8 +118,12 @@ export class ClientPool<T> {
    */
   private async release(requestTag: string, client: T): Promise<void> {
     const requestCount = this.activeClients.get(client) || 0;
-    assert(requestCount > 0, 'No active request');
+    assert(
```
schmidt-sebastian: Nit: I like the previous version better cause it is one line instead of 4.

dev/src/pool.ts
```diff
@@ -112,8 +118,12 @@ export class ClientPool<T> {
...
+      requestCount > 0,
+      'Release failed. No active requests are on the client.'
+    );
     this.activeClients.set(client, requestCount - 1);
+    this.operationsDeferred.resolve();
```
schmidt-sebastian: `release()` gets called after every request. It does not mean that no clients have pending operations. If you want to keep `operationsDeferred`, I think you need logic here to determine whether there are any operations left.

dev/src/pool.ts
```diff
@@ -195,6 +205,14 @@ export class ClientPool<T> {
 
   async terminate(): Promise<void> {
     this.terminated = true;
+
+    // If there are any operations still running after terminate() is called,
```
schmidt-sebastian: I think this only thing you have to do here is set `maxIdleClients` to 0 and call `shouldGarbageCollectClient()`.

It will remove the remaining client if it  is currently idle. If not, the next run of `shouldGarbageCollectClient()` will trigger the removal (which should happen after `release()`)

dev/system-test/firestore.ts
```diff
@@ -992,6 +992,19 @@ describe('DocumentReference class', () => {
...
+      const ref = randomCol.doc('doc-1');
+
+      const unsubscribe = ref.onSnapshot(() => {
+        // No-op
+      });
+
+      await ref.set({});
+      unsubscribe();
+      // Terminate should allow unsubscribe to complete.
+      await firestore.terminate();
```
schmidt-sebastian: I think this test won't actually fail in the current client, it will just log a rejected Promise warning. We might be better off adding this as a unit test to the ClientPool unit tests.

schmidt-sebastian:
> Do I read it correctly, that after resolving the Promise returned by `terminate()`, there might be still some actions pending? This is not an expected behavior, as it would again break some of the use cases for which this method was introduced in the first place. A correct behavior would be to wait for all of the pending operations to complete before resolving `terminate()`.

We are debating between letting terminate() "hang" until all listeners are disconnected or terminating all active listeners with errors. Having user feedback here would be valuable.

merlinnot:
In the original report you can see that listeners are actually disconnected before `terminate` is called. It's just that disconnecting the listener is exposed as a synchronous operation, whereas it's actually not.

I'd say that ideally disconnect of a listener should in that case also return a Promise, so you can shut down everything gracefully.

If this API change is not acceptable, what I'd expect is:
- When listeners were disconnected (synchronously) before calling `terminate()`, wait for all of the background operations to complete before resolving the Promise.
- When some of the listeners were not disconnected (synchronously) before calling `terminate()`, return a rejected Promise along the lines of "All listeners must be disconnected before terminating the client.".

Why? Because if you forget to disconnect some listeners and wait for `terminate()`, it will never resolve and debugging this issue might be hard. If you, on the other hand, terminate listeners with errors, you don't give developers a chance to shutdown the client gracefully.

schmidt-sebastian:
@thebrianchen What @merlinnot suggested sounds sane to me. So, on `terminate()` we error if there are any active listeners. If there are no active listens, we block until all other operations complete (including listener disconnects, RunQuery processing and any outstanding writes).

dev/src/index.ts
```diff
@@ -298,6 +298,13 @@ export class Firestore {
    */
   private _projectId: string | undefined = undefined;
 
+  /**
+   * Set of requestTags associated with the listeners that have been registered
+   * on the client.
+   * @private
+   */
+  private registeredListeners = new Set<string>();
```
schmidt-sebastian: I think you only need a counter. Requests tags are also not guaranteed to be unique, since the generator just uses whatever `rand()` implementation the system exposes.

dev/src/index.ts
```diff
@@ -987,12 +994,39 @@ export class Firestore {
...
   }
 
+  /**
+   * Registers the requestTag associated with the listener. This is used to
+   * ensure that all listeners are detached when terminate() is called.
+   *
+   * @private
+   * @param requestTag A unique client-assigned identifier
+   */
+  registerListener(requestTag: string): void {
```
schmidt-sebastian: It would be nice if this was somehow scoped to the class itself, but I can't find a good way to do this. `readStream` would be a natural place, but it would allow stream retries after the client is terminated.

dev/src/pool.ts
```diff
@@ -42,6 +42,12 @@ export class ClientPool<T> {
    */
   private terminated = false;
 
+  /**
+   * Deferred promise that is resolved when there are no active operations on
+   * the client pool.
+   */
+  private operationsDeferred = new Deferred<void>();
```
schmidt-sebastian: Can you come up with a more descriptive name?

dev/src/watch.ts
```diff
@@ -252,7 +252,8 @@ abstract class Watch<T = DocumentData> {
 
     this.initStream();
 
-    return () => {
+    const unsubscribe = () => {
+      this.firestore.unregisterListener(this.requestTag);
```
schmidt-sebastian: I don't think this is the only place that Listens are terminated. What happens if we throw an error and terminate from our end?

dev/src/pool.ts
```diff
@@ -195,6 +204,12 @@ export class ClientPool<T> {
 
   async terminate(): Promise<void> {
     this.terminated = true;
+
+    // Wait for all pending operations to complete before terminating.
+    if (this.opCount > 0) {
+      this.operationsDeferred = new Deferred<void>();
```
schmidt-sebastian: It would help to have a log line here.

dev/system-test/firestore.ts
```diff
@@ -152,6 +152,25 @@ describe('Firestore class', () => {
...
+    await ref.set({});
+    try {
+      await firestore.terminate();
+      return Promise.reject('terminate() should have failed');
+    } catch (err) {
+      expect(err).to.equal(
+        'All listeners must be disconnected before terminating the client.'
+      );
+      unsubscribe();
+      return Promise.resolve();
```
schmidt-sebastian: This line is not needed.

dev/system-test/firestore.ts
```diff
@@ -152,6 +152,25 @@ describe('Firestore class', () => {
...
+  it('throws an error if terminate() is called with active listeners', async () => {
+    const ref = randomCol.doc('doc-1');
+    const unsubscribe = ref.onSnapshot(() => {
+      // No-op
+    });
+
+    await ref.set({});
+    try {
+      await firestore.terminate();
+      return Promise.reject('terminate() should have failed');
```
schmidt-sebastian: Nit: Would prefer if you just threw an exception here. It's better not to mix Promises and async/await when possible.

dev/test/pool.ts
```diff
@@ -327,4 +327,25 @@ describe('Client pool', () => {
...
+
+  it('waits for existing operations to complete before releasing clients', async () => {
+    const clientPool = new ClientPool<{}>(1, 0, () => {
+      return {};
+    });
+    const deferred = new Deferred<void>();
+
+    const op = clientPool.run(REQUEST_TAG, async () => {
+      await deferred.promise;
+      return Promise.resolve('success');
```
schmidt-sebastian: You can just return `deferred.promise`.

dev/test/pool.ts
```diff
@@ -327,4 +327,25 @@ describe('Client pool', () => {
...
+    });
+    const terminate = clientPool
+      .terminate()
+      .then(() => Promise.resolve('success'));
```
schmidt-sebastian: You don't need to return a resolved Promise here, but you could use the `then` callback to modify some global state. That allows you to verify that it hasn't run:

```
let terminated = false;

const terminateOp = clientPool.terminate(() => { terminate = true });

// Give the Promise chain a chance to resolve

setImmediate(() => {
   expect(terminate).to.be.false;
   
   await terminateOp;
   expect(terminated).to.be.true;
   done(); // New done() callback
});

```

dev/test/pool.ts
```diff
@@ -327,4 +327,25 @@ describe('Client pool', () => {
...
+    deferred.resolve();
+
+    // Wait for the terminate promise to resolve after run() completes.
+    await terminate;
+    expect(op).to.become('success');
```
schmidt-sebastian: If you keep this, I think you need to call `await` since `become` is a PromisedAssertion.

dev/test/pool.ts
thebrianchen: This is very clean, thanks!

dev/src/pool.ts
thebrianchen: there are no names I could come up with that sound good: operationsFinishedDeferred, operationsCompletedDeferred, allRequestsCompleted, ready4terminatedeferred

dev/src/watch.ts
thebrianchen: `this.isActive` is only set here and in `closeStream()` (not counting constructor initialization). Added an unregister call in both places iff `isActive` is set to true.

dev/src/index.ts
thebrianchen: I spent a chunk of time thinking about this last week, and I couldn't think of a way to pipe it through another way. 
 - `ClientPool` is charge of terminating the client.
 - It seems like the only place to accurately track listeners w/o having to deal with extra node stream logic is to monitor each listener registration in watch.
- `Firestore` is the only place to store the Watch information.

dev/src/index.ts
```diff
@@ -298,6 +298,13 @@ export class Firestore {
    */
   private _projectId: string | undefined = undefined;
 
+  /**
+   * Set of requestTags associated with the listeners that have been registered
+   * on the client.
+   * @private
```
schmidt-sebastian: Please update JSDoc.

dev/src/index.ts
```diff
@@ -987,12 +994,39 @@ export class Firestore {
...
   }
 
+  /**
+   * Registers the requestTag associated with the listener. This is used to
+   * ensure that all listeners are detached when terminate() is called.
+   *
+   * @private
+   * @param requestTag A unique client-assigned identifier
+   */
+  registerListener(): void {
```
schmidt-sebastian: Please update JSDoc.

dev/src/index.ts
```diff
@@ -987,12 +994,39 @@ export class Firestore {
...
+  }
+
+  /**
+   * Unregisters the requestTag associated with the listener. This is used to
+   * ensure that all listeners are detached when terminate() is called.
+   *
+   * @private
+   * @param requestTag A unique client-assigned identifier
+   */
+  unregisterListener(): void {
```
schmidt-sebastian: Please update JSDoc.

dev/src/index.ts
```diff
@@ -987,12 +994,39 @@ export class Firestore {
...
+
   /**
    * Terminates the Firestore client and closes all open streams.
    *
    * @return A Promise that resolves when the client is terminated.
    */
   terminate(): Promise<void> {
+    if (this.registeredListenersCount > 0) {
+      return Promise.reject(
+        'All listeners must be disconnected before terminating the client.'
```
schmidt-sebastian: Please use the terms we use in our API: "onSnapshot() listeners" and "unsubscribe".

dev/src/pool.ts
```diff
@@ -42,6 +42,12 @@ export class ClientPool<T> {
    */
   private terminated = false;
 
+  /**
+   * Deferred promise that is resolved when there are no active operations on
+   * the client pool.
+   */
+  private operationsCompletedDeferred = new Deferred<void>();
```
schmidt-sebastian: Ideally, we would also block new tasks from submission. We could then use a single boolean ("terminatePending") and a deferred Promise ("terminateDeferred"), which might make this code a bit cleaner. "Operations Completed" is a state that happens all the time, but it is only relevant during shutdown.

dev/src/pool.ts
```diff
@@ -195,6 +204,17 @@ export class ClientPool<T> {
...
   async terminate(): Promise<void> {
     this.terminated = true;
+
+    // Wait for all pending operations to complete before terminating.
+    if (this.opCount > 0) {
+      this.operationsCompletedDeferred = new Deferred<void>();
+      logger(
+        'ClientPool.terminate',
+        /* requestTag= */ null,
+        'Waiting for all pending operations to complete before terminating'
```
schmidt-sebastian: Nit: Instead of "all", you could log the number of operations.

dev/system-test/firestore.ts
```diff
@@ -152,6 +152,24 @@ describe('Firestore class', () => {
...
       });
   });
+
+  it('throws an error if terminate() is called with active listeners', async () => {
+    const ref = randomCol.doc('doc-1');
+    const unsubscribe = ref.onSnapshot(() => {
+      // No-op
+    });
+
+    await ref.set({});
```
schmidt-sebastian: This line can be dropped.

dev/system-test/firestore.ts
```diff
@@ -152,6 +152,24 @@ describe('Firestore class', () => {
...
+    try {
+      await firestore.terminate();
+      throw new Error('terminate() should have failed');
+    } catch (err) {
+      expect(err).to.equal(
+        'All listeners must be disconnected before terminating the client.'
+      );
+      unsubscribe();
+    }
```
schmidt-sebastian: Nit: If you want, you can simplify this as such: https://github.com/googleapis/nodejs-firestore/blob/7562033876dc006e77d00b576b2541a7dfd30c66/dev/test/transaction.ts#L408

dev/test/pool.ts
```diff
@@ -327,4 +327,29 @@ describe('Client pool', () => {
...
+
+    // Run operation that completes after terminate() is called.
+    clientPool.run(REQUEST_TAG, () => {
+      return deferred.promise;
+    });
+    const terminateOp = clientPool.terminate().then(() => {
+      terminated = true;
+    });
+
+    setImmediate(async () => {
```
schmidt-sebastian: Just glancing at this, I don't quite follow why we need `setImmediate`. Can you add a comment (or remove it if it isn't necessary).

dev/src/pool.ts
thebrianchen: The `terminated` boolean already blocks new tasks from submission after terminate() is called. Renamed to `terminateDeferred`

dev/src/index.ts
```diff
@@ -995,22 +995,22 @@ export class Firestore {
   }
 
   /**
-   * Registers the requestTag associated with the listener. This is used to
-   * ensure that all listeners are detached when terminate() is called.
+   * Registers a listener on this client, incrementing the listener count. This
+   * is used to ensure that all listeners are unsubscribed when terminate() is
```
schmidt-sebastian: s/ensure/assert (or verify?). Here and below.

dev/src/pool.ts
```diff
@@ -121,7 +121,7 @@ export class ClientPool<T> {
     assert(requestCount > 0, 'No active requests');
     this.activeClients.set(client, requestCount - 1);
     if (this.opCount === 0) {
```
schmidt-sebastian: Suggestion: `if (terminated && this.opCount == 0)`

dev/src/pool.ts
```diff
@@ -207,13 +207,14 @@ export class ClientPool<T> {
 
     // Wait for all pending operations to complete before terminating.
     if (this.opCount > 0) {
-      this.operationsCompletedDeferred = new Deferred<void>();
+      this.terminateDeferred = new Deferred<void>();
```
schmidt-sebastian: If you apply the previous suggestion, this line can be removed. You may want to update the comment in line 45 as well.

dev/src/index.ts
schmidt-sebastian: I think you missed this comment.

thebrianchen: oops

dev/src/pool.ts
thebrianchen: removed here and added comment above to include that terminate() has to be called.

PR #925, opened 2020-02-11
fix: collectionReference.add() validation

### LINKED ISSUE #924: Add not working using withConverter
#### Environment details

  -  executed on firebase function 
  - `@google-cloud/firestore` version: 3.5 (as a depency of firebase-admin)

#### Steps to reproduce

```
  class Post {
    constructor(readonly title: string, readonly author: string) {}

    toString(): string {
      return this.title + ', by ' + this.author
    }
  }

  const postConverter = {
    toFirestore(p: Post){
      console.log('toFirestore got called')
      return {title: p.title, author: p.author}
    },
    fromFirestore(
      snapshot: firebase.firestore.QueryDocumentSnapshot,
    ): Post {
      const data = snapshot.data()
      return new Post(data.title, data.author)
    }
  }

  await firestore.firestore()
    .collection('posts')
    .withConverter(postConverter)
    .add(new Post("Foo", "123"))
}
```
 
Receiving the following error:

```
Error: Value for argument "data" is not a valid Firestore document. Couldn't serialize object of type "Post". Firestore doesn't support JavaScript objects with custom prototypes (i.e. objects that were created via the "new" operator).
    at Object.validateDocumentData (/srv/node_modules/@google-cloud/firestore/build/src/write-batch.js:617:15)
    at CollectionReference.add (/srv/node_modules/@google-cloud/firestore/build/src/reference.js:1881:23)
    at test (/srv/lib/index.js:54:10)
    at Layer.handle [as handle_request] (/srv/node_modules/express/lib/router/layer.js:95:5)
    at next (/srv/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/srv/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/srv/node_modules/express/lib/router/layer.js:95:5)
    at /srv/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/srv/node_modules/express/lib/router/index.js:335:12)
    at next (/srv/node_modules/express/lib/router/index.js:275:10)
```

So it seems like to me that the `toFirestore` doesnt get called. If i call `toFirebase` hardcoded it works.

Thanks!

thebrianchen:
Fixes #924 

Needed to convert data toFirestore first before performing validation. Also modified system test to use `CollectionReference.add()`.

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -222,11 +222,10 @@ describe('CollectionReference class', () => {
   });
 
   it('supports withConverter()', async () => {
-    const ref = firestore
+    const ref = await firestore
       .collection('col')
       .withConverter(postConverter)
-      .doc('doc');
-    await ref.set(new Post('post', 'author'));
+      .add(new Post('post', 'author'));
     const postData = await ref.get();
     const post = postData.data();
     expect(post).to.not.be.undefined;
--- a/dev/test/collection.ts
+++ b/dev/test/collection.ts
@@ -216,25 +216,47 @@ describe('Collection interface', () => {
     const overrides: ApiOverride = {
       commit: request => {
         docId = request.writes![0].update!.name!;
+        // Remove the auto-generated id and then verify that the rest of the
+        // protobuf matches.
+        delete request.writes![0].update!.name;
+        expect(request).to.deep.equal({
+          database: DATABASE_ROOT,
+          writes: [
+            {
+              update: {
+                fields: {
+                  author: {
+                    stringValue: 'author',
+                  },
+                  title: {
+                    stringValue: 'post',
+                  },
+                },
+              },
+              currentDocument: {
+                exists: false,
+              },
+            },
+          ],
+        });
 
         return response(writeResult(1));
       },
       batchGetDocuments: () => {
+        // Extract the auto-generated document ID.
+        const docIdSplit = docId.split('/');
+        const doc = document(
+          docIdSplit[docIdSplit.length - 1],
+          'author',
+          'author',
+          'title',
+          'post'
+        );
         const stream = through2.obj();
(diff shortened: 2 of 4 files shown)


dev/system-test/firestore.ts
```diff
@@ -222,11 +222,10 @@ describe('CollectionReference class', () => {
   });
 
   it('supports withConverter()', async () => {
-    const ref = firestore
+    const ref = await firestore
       .collection('col')
       .withConverter(postConverter)
-      .doc('doc');
-    await ref.set(new Post('post', 'author'));
+      .add(new Post('post', 'author'));
```
schmidt-sebastian: Please also add a corresponding unit test. Ideally, our entire SDK surface should be tested through unit tests first and integration tests second.

dev/test/collection.ts
```diff
@@ -211,6 +211,68 @@ describe('Collection interface', () => {
     });
   });
 
+  it('for CollectionReference.withConverter().add()', async () => {
+    let docId = '';
+    const overrides: ApiOverride = {
+      commit: request => {
+        docId = request.writes![0].update!.name!;
+        // Remove the auto-generated id and then verify that the rest of the
+        // protobuf matches.
```
schmidt-sebastian: Oh, wow. Good fix for this issue that I definitely didn't see coming.

dev/test/collection.ts
```diff
@@ -211,6 +211,68 @@ describe('Collection interface', () => {
...
+        delete request.writes![0].update!.name;
```
schmidt-sebastian: Instead of deleting the field, can you update the expected request to include the extracted docId? Might be slightly cleaner.

thebrianchen: done.

PR #931, opened 2020-02-18
fix: propagate converter in QueryOptions.with()

thebrianchen:
Reported in this [comment](https://github.com/googleapis/nodejs-firestore/issues/924#issuecomment-5872254330).

We previously were dropping the converter when calling `with()`, which meant that the return type would be typed `T` even though the underlying object was still `DocumentData`

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -1030,14 +1030,13 @@ export class QueryOptions<T> {
    * Returns the union of the current and the provided options.
    * @private
    */
-  with(
-    settings: Partial<Omit<QueryOptions<T>, 'converter'>>,
-    converter = defaultConverter as FirestoreDataConverter<T>
-  ): QueryOptions<T> {
+  with(settings: Partial<Omit<QueryOptions<T>, 'converter'>>): QueryOptions<T> {
     return new QueryOptions(
       coalesce(settings.parentPath, this.parentPath)!,
       coalesce(settings.collectionId, this.collectionId)!,
-      converter,
+      this.converter
+        ? this.converter
+        : (defaultConverter as FirestoreDataConverter<T>),
       coalesce(settings.allDescendants, this.allDescendants)!,
       coalesce(settings.fieldFilters, this.fieldFilters)!,
       coalesce(settings.fieldOrders, this.fieldOrders)!,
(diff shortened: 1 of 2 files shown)


dev/src/reference.ts
```diff
@@ -1030,14 +1030,13 @@ export class QueryOptions<T> {
...
-  with(
-    settings: Partial<Omit<QueryOptions<T>, 'converter'>>,
-    converter = defaultConverter as FirestoreDataConverter<T>
-  ): QueryOptions<T> {
+  with(settings: Partial<Omit<QueryOptions<T>, 'converter'>>): QueryOptions<T> {
     return new QueryOptions(
       coalesce(settings.parentPath, this.parentPath)!,
       coalesce(settings.collectionId, this.collectionId)!,
-      converter,
+      this.converter
```
schmidt-sebastian: Would this always be set?

dev/test/query.ts
```diff
@@ -658,6 +658,38 @@ describe('query interface', () => {
...
+        queryEquals(request, fieldFilters('title', 'EQUAL', 'post'));
+        return stream({document: doc, readTime: {seconds: 5, nanos: 6}});
+      },
+    };
+
+    return createInstance(overrides).then(async firestore => {
+      await firestore
+        .collection('collectionId')
+        .doc('documentId')
+        .set({title: 'post', author: 'author'});
```
schmidt-sebastian: This code snippet (as well as the "commit" override) is not really needed to the test. You are mocking the response, and hence the data doesn't actually need to exist.

dev/test/query.ts
```diff
@@ -658,6 +658,38 @@ describe('query interface', () => {
       expect(posts.docs[0].data().toString()).to.equal('post, by author');
     });
   });
+
+  it('for withConverter() propagation through queryOptions', async () => {
```
schmidt-sebastian: The test name could be improved. The  concatenated name is now "query interface for withConverter() propagation through queryOptions".

What do you think about renaming the test to something like "propagates withConverter() through QueryOptions"?

dev/src/reference.ts
thebrianchen: Yes it should be

PR #937, opened 2020-02-20
fix!: mark v1beta1 client as deprecated

schmidt-sebastian:
This PR marks the old v1beta1 protocol as deprecated, which has not been used in our client since 0.20.


synth.py
```diff
@@ -75,13 +75,26 @@
...
    "/protos/firestore_v1beta1_proto_api'"
- )
+)
 s.replace(
   "dev/test/gapic-firestore-v1beta1.ts",
   "/protos/protos'",
   "/protos/firestore_v1beta1_proto_api'"
 )
 
+# Mark v1beta1 as deprecated
```
JustinBeckwith: @AlanGasperini @alexander-fenster so this is a funny use for synthtool :) For cases where a specific API version itself is deprecated, how do we let customers know?

dev/src/v1beta1/firestore_client.ts
```diff
@@ -52,6 +54,7 @@ const version = require('../../../package.json').version;
  *       committed. Any read with an equal or greater `read_time` is guaranteed
  *       to see the effects of the transaction.
  * @class
+ * @deprecated Use v1/firestore_client instead.
```
JustinBeckwith: Do we know specifically which tools/situations where this will be surfaced?  I think the TypeScript compiler, and eslint will do stuff.  Should we consider emitting runtime warnings for users not using these kinds of tools though?

schmidt-sebastian: It shows up in the generated docs (barely legible) and as a warning in VSCode/IntelliJ/

synth.py
alexander-fenster: @lukesneeringer Is there any annotation that might help mark the whole API version as deprecated?

schmidt-sebastian:
@lukesneeringer @JustinBeckwith We want to launch this as part of the breaking Node 10 release, as deprecation warnings might be considered a breaking change by some of our users. Are there any other strategies worth exploring that don't involve search/replace?

JustinBeckwith:
that sounds fine to me

schmidt-sebastian:
Merging into feature branch.

PR #947, opened 2020-03-02
feat: implement Timestamp.valueOf() (#944)

### LINKED ISSUE #944: FR: Add valueOf() to Timestamp instances
This is a clone of https://github.com/firebase/firebase-js-sdk/issues/2632.

### [REQUIRED] Describe the problem
Unable to compare Timestamp instances with native comparators. Javascript calls `valueOf()` on the object and uses the results of that to do native comparisons.

#### Steps to reproduce:
Attempt to compare Timestamps directly.

#### Relevant Code:
```javascript
firebase = require('firebase-admin')
a = new firebase.firestore.Timestamp(1, 1)
b = new firebase.firestore.Timestamp(2, 2)
console.log(a < b) // false
console.log(b < a) // false
```

#### Workaround
Of course you can simply call `toMillis()` and compare those, but having it auto handled would be better.

dconeybe:
This enables Timestamp objects to be compared for relative ordering using the <, <=, >, and >= operators.

Fixes #944

### DIFF
--- a/dev/src/timestamp.ts
+++ b/dev/src/timestamp.ts
@@ -26,6 +26,15 @@ import api = google.firestore.v1;
  */
 const MS_TO_NANOS = 1000000;
 
+/*!
+ * The minimum legal value for the "seconds" property of a Timestamp object.
+ *
+ * This value corresponds to 0001-01-01T00:00:00Z.
+ *
+ * @type {number}
+ */
+const MIN_SECONDS = -62135596800;
+
 /**
  * A Timestamp represents a point in time independent of any time zone or
  * calendar, represented as seconds and fractions of seconds at nanosecond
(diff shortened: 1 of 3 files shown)


dev/src/timestamp.ts
```diff
@@ -26,6 +26,15 @@ import api = google.firestore.v1;
...
 const MS_TO_NANOS = 1000000;
 
+/*!
+ * The minimum legal value for the "seconds" property of a Timestamp object.
+ *
+ * This value corresponds to 0001-01-01T00:00:00Z.
+ *
+ * @type {number}
+ */
+const MIN_SECONDS = -62135596800;
```
schmidt-sebastian: Can you confirm that the backend rejects all older timestamps, as well as any timestamps past the year 10,000?

If that is the case, then we should add validation to the constructor. We need to make sure that `valueOf()` is never invoked on invalid data. If that is not the case, then we should go back to the drawing board and come up with a strategy.

I am also slightly inclined to ship this as part of the Node 8 deprecation that is around the corner, in which case we can make this a proper breaking change. The PR title would then be "feat!: ...".

dconeybe: I tested it out and an exception is thrown when invalid seconds are specified.  The message looks like this:

```
Error: 3 INVALID_ARGUMENT: Timestamp seconds exceeds limit for field.
```

I wrote a little app to test this out https://github.com/dconeybe/node-timestamp-range-test.

I like the idea of adding the validation to the constructor and shipping this as part of the Node 8 deprecation.  I will update this PR with that logic.

We could even go a step further and make `timestamp.ts` an exact copy of https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/timestamp.ts.  There is no reason why these two files (and their accompanying unit tests) should differ in any way.  Making them identical would help reduce the maintenance burden.  What do you think?

Issue #889, opened 2020-01-17
Firestore client retries a transaction when a document already exists


mad-it:
This issue relates to the fix introduced for https://github.com/googleapis/nodejs-firestore/issues/827

#### Current behavior

When a document already exists, if one tries create a document using the same path with `create()` inside of a transaction, the transaction will keep retrying until it runs out of attempts.

#### Expected behavior

When a document is created with `create()` inside of a transaction, the transaction should intermediately terminate and throw the error.

#### Environment details

  - OS: Ubuntu 19.10
  - Node.js version: 10.18.0
  - npm version:6.12
  - `@google-cloud/firestore` version: 3.4.0

schmidt-sebastian:
Thanks for reporting. We will try to address this shortly.

schmidt-sebastian:
Note: This is next on my list, but I will be out on vacation until Feb 18.

PR #953, opened 2020-03-04
feat: base transaction retries on error codes

schmidt-sebastian:
This PR implements consistent retry behavior all transaction methods:

As per go/transaction-retry-matrix, it implements the following behavior by error code and RPC:

**RPC: BeginTransaction, Rollback**
- Implicit retry for DEADLINE_EXCEEDED, INTERNAL and UNAVAILABLE with backoff (not in this PR since this is done by google-gax)

**RPC: RunQuery, BatchGetDocuments, Commit**
- Restart transaction for ABORTED (no backoff).
- Restart transaction for CANCELLED, UNKNOWN, DEADLINE_EXCEEDED, INTERNAL, UNAVAILABLE, UNAUTHENTICATED,  RESOURCE_EXHAUSTED (with backoff). 
  - Use maximum backoff for RESOURCE_EXHAUSTED.

A restart consists of a rollback and a BeginTransaction RPC with a `retryTransaction`.

I also updated the test cases to allow us to verify the backoff behavior.

Fixes https://github.com/googleapis/nodejs-firestore/issues/889
Fixes https://github.com/googleapis/nodejs-firestore/issues/827

### DIFF
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -471,15 +471,9 @@ export class Transaction {
    * @return A Promise that resolves after the delay expired.
    */
   private async maybeBackoff(error?: GoogleError) {
-    if (error) {
-      if (error.code === Status.RESOURCE_EXHAUSTED) {
-        this._backoff.resetToMax();
-      } else if (error.code === Status.ABORTED) {
-        // We don't backoff for ABORTED to avoid starvation
-        this._backoff.reset();
-      }
+    if (error && error.code === Status.RESOURCE_EXHAUSTED) {
+      this._backoff.resetToMax();
     }
-
     await this._backoff.backoffAndWait();
   }
 }
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -18,6 +18,7 @@ import {GoogleError, Status} from 'google-gax';
 
 import * as proto from '../protos/firestore_v1_proto_api';
 
+import {ExponentialBackoff} from './backoff';
 import {DocumentSnapshot, Precondition} from './document';
 import {Firestore, WriteBatch} from './index';
 import {logger} from './logger';
@@ -64,6 +65,7 @@ const READ_AFTER_WRITE_ERROR_MSG =
 export class Transaction {
   private _firestore: Firestore;
   private _writeBatch: WriteBatch;
+  private _backoff: ExponentialBackoff;
   private _requestTag: string;
   private _transactionId?: Uint8Array;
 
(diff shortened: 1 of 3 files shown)


dev/src/transaction.ts
```diff
@@ -419,6 +422,9 @@ export class Transaction {
         );
       }
 
+      this._writeBatch._reset();
+      await this.maybeBackoff(lastError);
```
wilhuff: `async/await` is amazing. ❤️

dev/src/transaction.ts
```diff
@@ -464,6 +463,25 @@ export class Transaction {
...
+   * @private
+   * @return A Promise that resolves after the delay expired.
+   */
+  private async maybeBackoff(error?: GoogleError) {
+    if (error) {
+      if (error.code === Status.RESOURCE_EXHAUSTED) {
+        this._backoff.resetToMax();
+      } else if (error.code === Status.ABORTED) {
+        // We don't backoff for ABORTED to avoid starvation
+        this._backoff.reset();
```
wilhuff: This contradicts comments in go/transaction-retry-matrix. Do we have experimental data to support that starvation is a problem? Given multiple clients contending on a write, multiple clients failing to back off (with randomness) means they'll never make progress.

dev/src/transaction.ts
```diff
@@ -562,13 +580,22 @@ function validateReadOptions(
...
-    // In transactions, the backend returns code ABORTED for reads that fail
-    // with contention. These errors should be retried for both GoogleError
-    // and GoogleError-alike errors (in case the prototype hierarchy gets
-    // stripped somewhere).
-    return error.code === Status.ABORTED;
+function isRetryableTransactionError(error: GoogleError): boolean {
+  if (error.code !== undefined) {
+    // This list is based on https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/transaction_runner.ts#L112
+    switch (error.code) {
+      case Status.ABORTED:
```
wilhuff: Is it worth distinguishing retry-the-rpc vs retry-the-whole-transaction errors?

dev/src/transaction.ts
schmidt-sebastian: I checked with the Firestore team and removed the special-casing on ABORTED, which simplifies the code a bit since we now always back off. This also allowed me to simplify the test suite.

dev/src/transaction.ts
schmidt-sebastian: The per RPC retry is actually driven by GAX. The one downside of this approach is that if say a RunQuery RPC fails with DEADLINE_EXCEEDED, the RPC will retried by GAX first. If GAX retries don't resolve the issue, then we will restart the transaction.

It's probably possible to turn off GAX retries, but it is a pretty invasive change.

Issue #945, opened 2020-02-28
limitToLast


sgarciac:
Hello, 
Are there plans to support a "limitToLast" parameter to Query? It seems to be available in the firetore javascript SDK.

schmidt-sebastian:
Yes!

PR #954, opened 2020-03-05
feat: add support for Query.limitToLast()

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/945

This adds support for limitToLast(), including order reserving for queries. Fortunately, Watch uses its own comparator, so we don't have to reverse the results in Watch.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -965,7 +965,7 @@ docChangesPropertiesToOverride.forEach(property => {
 
 /** Internal representation of a query cursor before serialization. */
 interface QueryCursor {
-  before?: boolean;
+  before: boolean;
   values: unknown[];
 }
 
@@ -1344,8 +1344,8 @@ export class Query<T = DocumentData> {
   }
 
   /**
-   * Creates and returns a new [Query]{@link Query} that's additionally limited
-   * to only return up to the specified number of documents.
+   * Creates and returns a new [Query]{@link Query} that only returns the
+   * first matching documents.
    *
    * This function returns a new (immutable) instance of the Query (rather than
    * modify the existing instance) to impose the limit.
@@ -1373,8 +1373,8 @@ export class Query<T = DocumentData> {
   }
 
   /**
-   * Creates and returns a new Query that only returns the last matching
-   * documents.
+   * Creates and returns a new [Query]{@link Query} that only returns the
+   * last matching documents.
    *
    * You must specify at least one orderBy clause for limitToLast queries,
    * otherwise an exception will be thrown during execution.
@@ -1525,11 +1525,7 @@ export class Query<T = DocumentData> {
       );
     }
 
-    const options: QueryCursor = {values: []};
-
-    if (before) {
-      options.before = true;
-    }
+    const options: QueryCursor = {values: [], before};
 
     for (let i = 0; i < fieldValues.length; ++i) {
       let fieldValue = fieldValues[i];
@@ -1882,20 +1878,14 @@ export class Query<T = DocumentData> {
    * Converts a QueryCursor to its proto representation.
    *
    * @param cursor The original cursor value
-   * @param reverse Whether to flip the "before" value (e.g. for limitToLast
-   * queries)
    * @private
    */
-  private toCursor(
-    cursor: QueryCursor | undefined,
-    reverse = false
-  ): api.ICursor | undefined {
+  private toCursor(cursor: QueryCursor | undefined): api.ICursor | undefined {
     if (cursor) {
       const values = cursor.values.map(
         val => this._serializer.encodeValue(val) as api.IValue
       );
-      const before = (cursor.before === true) !== reverse ? true : undefined;
-      return {before, values};
+      return cursor.before ? {before: true, values} : {values};
     }
 
     return undefined;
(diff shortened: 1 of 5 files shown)


dev/src/reference.ts
```diff
@@ -969,6 +969,15 @@ interface QueryCursor {
   values: unknown[];
 }
 
+/*!
```
wu-hui: Ultranit: should it be `/**`?

dev/src/reference.ts
```diff
@@ -1349,7 +1365,38 @@ export class Query<T = DocumentData> {
...
-    const options = this._queryOptions.with({limit});
+    const options = this._queryOptions.with({
+      limit,
+      limitType: LimitType.First,
+    });
+    return new Query(this._firestore, options);
+  }
+
+  /**
+   * Creates and returns a new Query that only returns the last matching
```
wu-hui: We should update `limit()`'s comment to mirror the wording there as well.

dev/src/reference.ts
```diff
@@ -1875,21 +1947,44 @@ export class Query<T = DocumentData> {
...
+
+      structuredQuery.orderBy = this._queryOptions.fieldOrders!.map(order => {
+        // Flip the orderBy directions since we want the last results
+        const dir =
+          order.direction === 'DESCENDING' ? 'ASCENDING' : 'DESCENDING';
+        return new FieldOrder(order.field, dir).toProto();
+      });
+
+      // Swap the cursors to match the now-flipped query ordering.
+      structuredQuery.startAt = this.toCursor(
```
wu-hui: I'd suggest keeping the cursor reverse logic here instead of inside `toCursor`, because the order flipping is also done in this block and that makes the intention more explicit IMHO, and that is also how we do it in other SDKs.

dev/src/reference.ts
```diff
@@ -1898,13 +1993,13 @@ export class Query<T = DocumentData> {
   }
 
   /**
-   * Internal streaming method that accepts an optional transaction id.
+   * Internal streaming method that accepts the requets proto.
```
wu-hui: s/requests/request.

types/firestore.d.ts
```diff
@@ -903,6 +903,21 @@ declare namespace FirebaseFirestore {
      * @return The created Query.
      */
     limit(limit: number): Query<T>;
+    
+    /**
+     * Creates and returns a new Query that only returns the last matching
```
wu-hui: Same comment about bringing `limit()`'s comment in the same wording.

dev/system-test/firestore.ts
```diff
@@ -1079,6 +1079,18 @@ describe('Query class', () => {
       });
   };
 
+  function addDocs(...data: DocumentData[]): Promise<WriteResult[]> {
```
wu-hui: Nice refactoring!

dev/system-test/firestore.ts
```diff
@@ -1247,54 +1248,42 @@ describe('Query class', () => {
...
+      .orderBy('doc')
+      .limitToLast(2)
+      .get();
+    expectDocs(res, {doc: 2}, {doc: 3});
   });
 
-  it('has startAt() method', () => {
-    const ref1 = randomCol.doc('doc1');
-    const ref2 = randomCol.doc('doc2');
+  it('limitToLast() supports Query cursors', async () => {
```
wu-hui: Can we add error cases for `limitToLast()`, like cannot be streamed and requires explicit orderBy?

dev/system-test/firestore.ts
```diff
@@ -1820,6 +1794,29 @@ describe('Query class', () => {
           unsubscribe();
         });
     });
+
+    it('orders limitToLast() correctly', async () => {
```
wu-hui: Can this be moved to other limitToLast tests?

dev/test/query.ts
```diff
@@ -1201,6 +1200,115 @@ describe('limit() interface', () => {
...
+      query = query.orderBy('foo').limitToLast(2);
+      const result = await query.get();
+      expect(result.docs[0].id).to.equal('first');
+      expect(result.docs[1].id).to.equal('second');
+    });
+  });
+
+  it('expects number', () => {
+    const query = firestore.collection('collectionId');
+    expect(() => query.limit(Infinity)).to.throw(
```
wu-hui: should this be `limitToLast` instead?

dev/src/reference.ts
schmidt-sebastian: Done.

dev/src/reference.ts
schmidt-sebastian: I am not a huge fan of the code this creates, but it makes sense to keep in aligned with the rest of our SDKs.

dev/src/reference.ts
schmidt-sebastian: Done

dev/system-test/firestore.ts
schmidt-sebastian: This test is an integration test (in system-test/) and thus cannot be moved.

dev/test/query.ts
schmidt-sebastian: Oh yes!

types/firestore.d.ts
schmidt-sebastian: Also done here.

dev/src/reference.ts
wu-hui: I learn something new.

PR #961, opened 2020-03-09
feat: add BatchWriteRequests to WriteBatch.commit()

thebrianchen:


dev/src/write-batch.ts
```diff
@@ -517,12 +560,16 @@ export class WriteBatch {
    * @param commitOptions.transactionId The transaction ID of this commit.
    * @param commitOptions.requestTag A unique client-assigned identifier for
    * this request.
+   * @param useBatch Whether to use BatchWrite API instead of CommitRequest.
```
schmidt-sebastian: A better name would be "atomic" or similar.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
     );
 
     if (explicitTransaction) {
+      assert(!useBatch, 'Cannot use batch mode with transactions');
```
schmidt-sebastian: You also need to make sure that the block above never creates a transaction.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
...
-              )
-          );
-        }
-        return [];
-      });
+    if (useBatch) {
+      return this._firestore
+        .request<api.IBatchWriteRequest, api.BatchWriteResponse>(
+          'batchWrite',
+          request as api.IBatchWriteRequest,
```
schmidt-sebastian: I am leaning towards recommending that you should have two different methods. The new method would be fairly short since we can't support implicit transactions directly.

We should also think about if we need a dummy request for batch writes to make sure the network is initialized.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
...
-                writeResult.updateTime
-                  ? Timestamp.fromProto(writeResult.updateTime)
-                  : commitTime
-              )
-          );
-        }
-        return [];
-      });
+    if (useBatch) {
+      return this._firestore
```
schmidt-sebastian: Please use `await` for new or heavily refactored code to avoid nesting.

dev/src/write-batch.ts
```diff
@@ -103,6 +104,33 @@ export class WriteResult {
   }
 }
 
+/**
+ * A BatchWriteResult wraps the write time and status returned by Firestore
+ * when making BatchWriteRequests.
+ *
+ * @private
+ */
+export class BatchWriteResult {
```
schmidt-sebastian: I think this type could be replaced with a typed array: Array<[WriteResult, Status]>. It would remove the conversion logic you had to add.

What do you think?

dev/src/write-batch.ts
```diff
@@ -506,7 +534,22 @@ export class WriteBatch {
...
   commit(): Promise<WriteResult[]> {
-    return this.commit_();
+    return this.commit_() as Promise<WriteResult[]>;
+  }
+
+  /**
+   * Commits all pending operations to the database and verifies all
+   * preconditions. Fails the entire write if any precondition is not met.
+   *
+   * Though each write is atomic, the writes in the batch are not applied
```
schmidt-sebastian: Would remove "Though each write is atomic".

dev/src/write-batch.ts
```diff
@@ -506,7 +534,22 @@ export class WriteBatch {
    * });
    */
   commit(): Promise<WriteResult[]> {
-    return this.commit_();
+    return this.commit_() as Promise<WriteResult[]>;
+  }
+
+  /**
+   * Commits all pending operations to the database and verifies all
+   * preconditions. Fails the entire write if any precondition is not met.
```
schmidt-sebastian: Is the latter still true? I would expect that the backend would not be able to enforce this any more.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
...
-      });
+    if (useBatch) {
+      return this._firestore
+        .request<api.IBatchWriteRequest, api.BatchWriteResponse>(
+          'batchWrite',
+          request as api.IBatchWriteRequest,
+          tag
+        )
+        .then(resp => {
+          const zipped: Array<[
```
schmidt-sebastian: I don't fully understand the need for the intermediate conversion.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
...
+            return [result, resp.status[i]];
+          });
+          if (request.writes!.length > 0) {
+            return zipped.map(
+              ([writeResult, status]) =>
+                new BatchWriteResult(
+                  writeResult.updateTime
+                    ? Timestamp.fromProto(writeResult.updateTime)
+                    : null,
+                  status.code as Status
```
schmidt-sebastian: This should be the status with its error message.

dev/src/write-batch.ts
```diff
@@ -567,25 +614,55 @@ export class WriteBatch {
...
+            return resp.writeResults.map(
+              writeResult =>
+                new WriteResult(
+                  writeResult.updateTime
+                    ? Timestamp.fromProto(writeResult.updateTime)
+                    : commitTime
+                )
+            );
+          }
+          return [];
```
schmidt-sebastian: This could just be
```
return resp.writeResults.map(write => new WriteResult(Timestamp.fromProto(write.updateTime || resp.commitTime!));
```

dev/src/write-batch.ts
```diff
@@ -517,12 +560,16 @@ export class WriteBatch {
...
-    transactionId?: Uint8Array;
-    requestTag?: string;
-  }): Promise<WriteResult[]> {
+  async commit_(
+    commitOptions?: {
+      transactionId?: Uint8Array;
+      requestTag?: string;
+    },
+    useBatch = false
+  ): Promise<WriteResult[] | BatchWriteResult[]> {
```
schmidt-sebastian: I think this can always return `BatchWriteResult`. `commit()` can drop the unused Status.

dev/src/write-batch.ts
thebrianchen: Separated out logic into `batchCommit()` method

dev/src/write-batch.ts
thebrianchen: N/A if we separate out atomic logic into separate method

dev/src/write-batch.ts
thebrianchen: forgot to refactor 🥇

dev/src/write-batch.ts
thebrianchen: I initially considered that, but opted not to because we would be indexing into the array to access WriteResult and Status, which seemed less readable than being able to directly access properties on an object

dev/src/write-batch.ts
thebrianchen: N/A since I separated out the logic

dev/src/write-batch.ts
thebrianchen: Added an RpcStatus interface to hold the Status, not sure if that's what you were looking for

dev/src/write-batch.ts
thebrianchen: I copied this from the commit() comment and I'm interpreting this to mean that if the the precondition of a single write isn't met the entire batch write fails, which seems to be the case. 
```
await ref.set({foo: 'c'});
await batch.update(ref, {foo: 'a'})
   .delete(ref, { *some invalid update time*});
```
The batch will fail with the error: `Error: 3 INVALID_ARGUMENT: Cannot collapse mutations with different base versions`, and `foo` will still be 'c'.

dev/test/write-batch.ts
```diff
@@ -66,10 +66,6 @@ describe('set() method', () => {
     );
   });
 
-  it('accepts preconditions', () => {
```
thebrianchen: writeBatch.set() does not accept preconditions

dev/src/write-batch.ts
schmidt-sebastian: Yeah, but this shouldn't be true for BulkWriter.

dev/src/write-batch.ts
```diff
@@ -103,6 +104,39 @@ export class WriteResult {
   }
 }
 
+/** Helper type to hold a google.rpc.Status object. */
+interface RpcStatus {
```
schmidt-sebastian: I think the type we actually want to return is `GoogleError` which exists in GoogleGax and matches your interface. Can you use that type instead? This allows us to use the same type in all Promise rejections.

dev/src/write-batch.ts
schmidt-sebastian: Sounds reasonable.

dev/src/write-batch.ts
```diff
@@ -103,6 +104,39 @@ export class WriteResult {
...
+
+/**
+ * A BatchWriteResult wraps the write time and status returned by Firestore
+ * when making BatchWriteRequests.
+ *
+ * @private
+ */
+export class BatchWriteResult {
+  constructor(
+    private readonly _writeTime: Timestamp | null,
```
schmidt-sebastian: Since this is an internal type, there is no need to do the `_` + getter trick, which really only exists to provide JSDoc anyways.

dev/src/write-batch.ts
```diff
@@ -509,6 +543,51 @@ export class WriteBatch {
...
+   * Commits all pending operations to the database and verifies all
+   * preconditions. Fails the entire write if any precondition is not met.
+   *
+   * The writes in the batch are not applied atomically and can be applied out
+   * of order.
+   *
+   * @private
+   */
+  async batchCommit(commitOptions?: {
+    requestTag?: string;
```
schmidt-sebastian: Since this is private and one argument, I think we don't need to have an options object here.

dev/src/write-batch.ts
```diff
@@ -509,6 +543,51 @@ export class WriteBatch {
...
+    this._committed = true;
+    const tag = (commitOptions && commitOptions.requestTag) || requestTag();
+    await this._firestore.initializeIfNeeded(tag);
+
+    const database = this._firestore.formattedName;
+    const request: api.IBatchWriteRequest = {database};
+    const writes = this._ops.map(op => op());
+    request.writes = [];
+
+    for (const req of writes) {
```
schmidt-sebastian: Suggestion:

```
    const request: api.ICommitRequest = {database, writes: []};
    
    const writes = this._ops.map(op => op());
    for( ...
```

You may also fix in the original code.

dev/src/write-batch.ts
```diff
@@ -509,6 +543,51 @@ export class WriteBatch {
...
+      }
+      request.writes.push(req.write);
+    }
+
+    const response = await this._firestore.request<
+      api.IBatchWriteRequest,
+      api.BatchWriteResponse
+    >('batchWrite', request, tag);
+
+    if (request.writes!.length > 0) {
```
schmidt-sebastian: You can drop this if check. map() will do the right thing.

Here and below.

dev/test/write-batch.ts
```diff
@@ -462,3 +460,88 @@ describe('batch support', () => {
     });
   });
 });
+
+describe('batch support', () => {
```
schmidt-sebastian: Nit: Don't re-use "batch", maybe rename to "bulk commit".

dev/src/write-batch.ts
```diff
@@ -509,6 +543,51 @@ export class WriteBatch {
...
+  /**
+   * Commits all pending operations to the database and verifies all
+   * preconditions. Fails the entire write if any precondition is not met.
+   *
+   * The writes in the batch are not applied atomically and can be applied out
+   * of order.
+   *
+   * @private
+   */
+  async batchCommit(commitOptions?: {
```
schmidt-sebastian: What do you think of the name "bulkCommit"? I would like to reduce confusion, and deciding between "commit" and "batchCommit" is going to be tough.

dev/test/write-batch.ts
schmidt-sebastian: But it does accept them as document data :) Thanks for fixing.

thebrianchen: updated the test name

dev/src/write-batch.ts
thebrianchen: good point, dropped

dev/src/write-batch.ts
thebrianchen: thanks!

dev/src/write-batch.ts
thebrianchen: removed usage of `requestTag`

dev/src/write-batch.ts
thebrianchen: I like that more -- the hard part was figuring out where to draw the line since we refer to it externally as bulk writes, but the backend API refers to it as batch writes.

dev/src/write-batch.ts
thebrianchen: done inside bulkCommit()

dev/src/write-batch.ts
thebrianchen: Ah makes sense. Removed the sentence about preconditions

thebrianchen:
@schmidt-sebastian Would you mind taking one last look? made some other changes that I want another look over (using || check, GoogleError usage)

PR #962, opened 2020-03-10
deps: update TypeScript

schmidt-sebastian:
@JustinBeckwith Since we use our own manually generated types, does the reason for not updating all of GCloud to a newer TypeScript version apply to us?


bcoe:
@schmidt-sebastian whoops, we should have probably added a `~` like @JustinBeckwith suggested, might be worth a follow up patch (sorry about that).

PR #965, opened 2020-03-10
feat!: use QueryDocumentSnapshot in FirestoreDataConverter

### LINKED ISSUE #938: FirestoreDataConverter.fromFirestore should use DocumentSnapshot instead of DocumentData

The implementation and therefore the interface interface of the `FirestoreDataConverter` between the nodejs client and the web-js client diverges:

Whereas the web client uses the `DocumentSnapshot` as the input value of the `fromFirestore` function, the nodejs client only uses the `DocumentData` representation of the document:

[nodejs - client](https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L83):
```typescript
interface FirestoreDataConverter<T> {
  fromFirestore(data: DocumentData): T;
}
```

[web - client](https://github.com/firebase/firebase-js-sdk/blob/eb614f0e7bcfee5b4bc1d390f894bea3d8f3867b/packages/firestore/src/api/database.ts#L1384):
```typescript
interface FirestoreDataConverter<T> {
  fromFirestore(snapshot: QueryDocumentSnapshot, options: SnapshotOptions): T;
}
```

The problem with the current implementation is that it is impossible to access document metadata like the documents id.

My preferred solution would be to modify the behavior so that the `DocumentSnapshot` itself is handed to the `fromFirebase` method of the converter:

```typescript
interface FirestoreDataConverter<T> {
  fromFirestore(snapshot: DocumentSnapshot<T>): T;
}
```

⚠️ This would result in a breaking change.

Alternatives cloud be:
* Passing the metadata as a separate parameter to the converter. (would result in a non-breaking change). 😑
* Leaving everything as is. 😞 

 **Additional context**
I would be willing to do the implementation myself providing that there is a positive feedback on the change itself.

thebrianchen:
Fixes #938 

Wait for node8 to be deprecated before releasing.


dev/src/document.ts
```diff
@@ -381,11 +381,26 @@ export class DocumentSnapshot<T = DocumentData> {
       return undefined;
     }
 
-    const obj: DocumentData = {};
-    for (const prop of Object.keys(fields)) {
-      obj[prop] = this._serializer.decodeValue(fields[prop]);
+    // We only want to use the converter and create a new QueryDocumentSnapshot
+    // if a converter has been provided.
+    if (this.ref._converter !== defaultConverter) {
+      const ref = new DocumentReference(this.ref.firestore, this.ref._path);
```
schmidt-sebastian: I don't think you need to use the builder here since we are inside the DocumentSnapshot. Something like this would work:

```
      const untypedReference = new DocumentReference(this.ref.firestore, this.ref._path);
      return this.ref._converter.fromFirestore(
        new QueryDocumentSnapshot<DocumentData>(untypedReference, this._fieldsProto,
          this.readTime, this.createTime, this.updateTime)
      );
```

thebrianchen: done

schmidt-sebastian:
Merging into feature branch.

Issue #969, opened 2020-03-13
Upgrading from 3.6.0 to 3.7.0 throws internal error.


mad-it:
I have an applications running on GKE. When upgrading this application's dependency from 3.6 to 3.7 I get the following error:
```
Error: INTERNAL ERROR: Client is not yet ready to issue requests.
    at Firestore.get projectId [as projectId] (/run/node_modules/@google-cloud/firestore/build/src/index.js:401:19)
    at CollectionReference.toProto (/run/node_modules/@google-cloud/firestore/build/src/reference.js:1556:42)
    at CollectionReference.stream (/run/node_modules/@google-cloud/firestore/build/src/reference.js:1521:30)
```

I tried using `setLogFunction` to print some other useful information, but there is nothing useful in the output. Only line I am getting before the stacktrace is the following:
```
Firestore (3.7.0) 2020-03-13T07:41:13.099Z ##### [Firestore]: Initialized Firestore
```

As you can see from the stacktrace, the only thing we use is the `CollectionReference.stream()` method.

Sorry that I could not add any additional reproduction steps.

TheAmazingPT:
It seems that i have the same problem, when running my node app in **Google Cloud Functions**.
All credentials are provided and no changes have been made in the code.
The deployment is done via **Google Cloud Build**, where the node app is built and bundled.

In the package.json the Firestore package is declared as `"@google-cloud/firestore": "^3.3.2"`.

When forcing the version to `3.6.0`, it seems to work flawlessly.

themikefuller:
Got this same error today when I was changing some settings on a Cloud Function. I thought initially my tinkering with the settings caused the problem. It wasn't fixed by reverting back to the old settings. I can confirm that forcing "@google-cloud/firestore":"3.6.0" in the dependencies corrected the issue.

AndriiTsok:
We experience the same problem and error message.

>> Error: INTERNAL ERROR: Client is not yet ready to issue requests.

schmidt-sebastian:
We are looking at this.

schmidt-sebastian:
We have pinned 3.6.0 as our latest release.

PR #971, opened 2020-03-16
fix: support Query.stream() as first client operation

schmidt-sebastian:
This fixes a regression introduced in 3.7

Fixes https://github.com/googleapis/nodejs-firestore/issues/969

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -1787,14 +1787,12 @@ export class Query<T = DocumentData> {
    * @param {bytes=} transactionId A transaction ID.
    */
   _get(transactionId?: Uint8Array): Promise<QuerySnapshot<T>> {
-    const request = this.toProto(transactionId);
-
     const docs: Array<QueryDocumentSnapshot<T>> = [];
 
     return new Promise((resolve, reject) => {
       let readTime: Timestamp;
 
-      this._stream(request)
+      this._stream(transactionId)
         .on('error', err => {
           reject(err);
         })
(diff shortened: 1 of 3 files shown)


dev/test/query.ts
```diff
@@ -1277,7 +1280,8 @@ describe('limitToLast() interface', () => {
 
   it('requires at least one ordering constraints', () => {
     const query = firestore.collection('collectionId');
-    expect(() => query.limitToLast(1).get()).to.throw(
+    const result = query.limitToLast(1).get();
+    return expect(result).to.eventually.be.rejectedWith(
```
schmidt-sebastian: All the refactor in the original PR was meant to ensure that this is thrown as an exception, rather than as a rejected Promise, but it is probably not worth breaking the client for.

dev/src/reference.ts
```diff
@@ -1989,11 +1986,11 @@ export class Query<T = DocumentData> {
   /**
    * Internal streaming method that accepts the request proto.
```
alexander-fenster: The comment is now outdated.

dev/src/reference.ts
```diff
@@ -2020,26 +2017,30 @@ export class Query<T = DocumentData> {
...
-          backendStream.on('error', err => {
-            logger(
-              'Query._stream',
-              tag,
-              'Query failed with stream error:',
-              err
-            );
+    this.firestore
+      .initializeIfNeeded(tag)
+      .then(async () => {
```
alexander-fenster: `async` is not needed, there is no `await` inside.

schmidt-sebastian: I added a comment. `async` makes sure that any exception from `this.toProto` is converted into a rejected Promise.

dev/src/reference.ts
schmidt-sebastian: Fixed

PR #978, opened 2020-03-20
test: update conformance tests

thebrianchen:
Replacing old binproto test suite with individual json test files from master. All tests are passing except for a new test that uses update_transforms.

Also needed to manually convert Timestamp and Int32Value json representations to objects in order to support protobufjs's lack of support.


dev/conformance/runner.ts
```diff
@@ -514,17 +521,91 @@ function runTest(spec: ConformanceProto) {
   );
 }
 
+/**
+ * Parses through the TestFile and changes all instances of Timestamps encoded
+ * as strings to a proper protobuf type since protobufJS does not support it at
+ * the moment.
+ */
+function updateTimeStamp(obj: {[key: string]: {}}) {
```
schmidt-sebastian: Nit: `updateTimestamp` (in my opinion)

You could also use `normalizeTimestamp`, which is a concept we just introduced to the JS SDK.

dev/conformance/runner.ts
```diff
@@ -514,17 +521,91 @@ function runTest(spec: ConformanceProto) {
...
+  const fieldNames = ['updateTime', 'createTime', 'readTime'];
+  for (const key of Object.keys(obj)) {
+    if (fieldNames.includes(key) && typeof obj[key] === 'string') {
+      obj[key] = timestamp((obj[key] as unknown) as string);
```
schmidt-sebastian: Can you find a way to remove the cast here? You may have to assign obj[key] to a temporary variable.

dev/conformance/runner.ts
```diff
@@ -514,17 +521,91 @@ function runTest(spec: ConformanceProto) {
...
+      updateTimeStamp(obj[key]);
+    }
+  }
+}
+
+/**
+ * Convert a string TimeStamp, e.g. "1970-01-01T00:00:42Z", to its protobuf
+ * type.
+ */
+function timestamp(text: string): {[key: string]: number} {
```
schmidt-sebastian: convertTimestamp()?

Come to think of it, if you don't think this is going to be temporary, can you explore adding it to `convertInput` which already exists in the conformance tests?

thebrianchen: We need to convert the timestamp before creating the `TestFile` with `fromObject`, so adding it into `convertInput` would be too late. However, I could pull both of the functions into a `normalizeInput` instead of having `normalizeTimestamp` and `normalizeInt32Value`. Thoughts?

dev/conformance/pre-runner.ts
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const mkdirp = require('mkdirp');
+
+// We have to manually add the directory here because the docker image that
+// runs CI does not have this directory. We also can't use `mkdir -p` in the
+// package.json file because CI is also run on Windows, which does not support
+// the `-p` flag.
+mkdirp.sync('build/conformance/conformance-tests');
```
schmidt-sebastian: Can we move this to the ./scripts folder? You should also consider making this a `js` file so you can run it directly via Node.

package.json
```diff
@@ -41,7 +41,7 @@
     "lint": "gts check",
     "clean": "gts clean",
     "compile": "tsc -p .",
-    "postcompile": "cp -r dev/protos build && cp dev/src/v1beta1/*.json build/src/v1beta1/ && cp dev/src/v1/*.json build/src/v1/ && cp dev/conformance/test-definition.proto build/conformance && cp dev/conformance/test-suite.binproto build/conformance",
+    "postcompile": "ts-node dev/conformance/pre-runner.ts && cp -r dev/protos build && cp dev/src/v1beta1/*.json build/src/v1beta1/ && cp dev/src/v1/*.json build/src/v1/ && cp dev/conformance/test-definition.proto build/conformance && cp dev/conformance/conformance-tests/*.json build/conformance/conformance-tests",
```
schmidt-sebastian: Should `pre-runner` be renamed to something akin `init-directories`? It is not very obvious what `pre-runner` does.

dev/conformance/runner.ts
schmidt-sebastian: Let's leave it is as is, but rename this function to `convertTimestamp`. I would expect a `timestamp` function to take seconds and nanos and return a Timestamp API obejct.

package.json
thebrianchen: better name, thanks

PR #984, opened 2020-03-24
Use Random Number from `crypto` to generate AutoId.

wu-hui:
Switch to use random number generator from `crypto` to reduce potential conflicts

### DIFF
--- a/dev/src/util.ts
+++ b/dev/src/util.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


dev/src/util.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { randomBytes } from 'crypto';
```
wilhuff: The format of this import appears different than the ones below.

PR #988, opened 2020-03-26
feat: add BulkWriter first draft

thebrianchen:
I wanted to get some first round feedback on the overall design for BulkWriter and see if my functions and data structures make sense. 

@schmidt-sebastian I left some TODO comments in my tests with questions for you.
Still TODO:
- Add more tests (see comments)
- Add converter support

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -49,6 +49,9 @@ class BulkCommitBatch {
   // The set of references present in the WriteBatch.
   readonly refsInBatch = new Set<DocumentReference>();
 
+  // The number of writes in this batch.
+  private _opCount = 0;
+
   // Whether the batch is ready to be sent. Writes cannot be added once a batch
   // is marked as pending.
   private _pending = false;
@@ -75,7 +78,7 @@ class BulkCommitBatch {
   }
 
   get opCount(): number {
-    return this.writeBatch.opCount;
+    return this._opCount;
   }
 
   get pending(): boolean {
@@ -90,13 +93,8 @@ class BulkCommitBatch {
     documentRef: DocumentReference,
     data: DocumentData
   ): Promise<WriteResult> {
-    this.refsInBatch.add(documentRef);
     this.writeBatch.create(documentRef, data);
-    this.resultsMap.set(
-      this.writeBatch.opCount,
-      new Deferred<BatchWriteResult>()
-    );
-    return this.getResult(this.writeBatch.opCount);
+    return this.processOperation(documentRef);
   }
 
   /**
@@ -107,13 +105,8 @@ class BulkCommitBatch {
     documentRef: DocumentReference,
     precondition?: Precondition
   ): Promise<WriteResult> {
-    this.refsInBatch.add(documentRef);
     this.writeBatch.delete(documentRef, precondition);
-    this.resultsMap.set(
-      this.writeBatch.opCount,
-      new Deferred<BatchWriteResult>()
-    );
-    return this.getResult(this.writeBatch.opCount);
+    return this.processOperation(documentRef);
   }
 
   /**
@@ -125,13 +118,8 @@ class BulkCommitBatch {
     data: DocumentData,
     options?: SetOptions
   ): Promise<WriteResult> {
-    this.refsInBatch.add(documentRef);
     this.writeBatch.set(documentRef, data, options);
-    this.resultsMap.set(
-      this.writeBatch.opCount,
-      new Deferred<BatchWriteResult>()
-    );
-    return this.getResult(this.writeBatch.opCount);
+    return this.processOperation(documentRef);
   }
 
   /**
@@ -145,13 +133,21 @@ class BulkCommitBatch {
       {lastUpdateTime?: Timestamp} | unknown | string | FieldPath
     >
   ): Promise<WriteResult> {
-    this.refsInBatch.add(documentRef);
     this.writeBatch.update(documentRef, dataOrField, ...preconditionOrValues);
-    this.resultsMap.set(
-      this.writeBatch.opCount,
-      new Deferred<BatchWriteResult>()
-    );
-    return this.getResult(this.writeBatch.opCount);
+    return this.processOperation(documentRef);
+  }
+
+  /**
+   * Helper to update data structures associated with the operation and
+   * return the result.
+   */
+  private processOperation(
+    documentRef: DocumentReference
+  ): Promise<WriteResult> {
+    this.refsInBatch.add(documentRef);
+    this._opCount++;
+    this.resultsMap.set(this.opCount, new Deferred<BatchWriteResult>());
+    return this.getResult(this.opCount);
   }
 
   /**
@@ -161,7 +157,7 @@ class BulkCommitBatch {
   canAddDoc(documentRef: DocumentReference): boolean {
     return (
       !this.pending &&
-      this.writeBatch.opCount < this.maxBatchSize &&
+      this.opCount < this.maxBatchSize &&
       !this.refsInBatch.has(documentRef)
     );
   }
@@ -251,7 +247,7 @@ export class BulkWriter {
     private readonly firestore: Firestore,
     private readonly options?: BulkWriterOptions
   ) {
-    if (options && !options.disableThrottling) {
+    if (options === undefined || options.throttlingEnabled) {
       this.rampMaxOps();
     }
   }
@@ -304,7 +300,7 @@ export class BulkWriter {
    *
    * @example
    * let bulkWriter = firestore.bulkWriter();
-   * let documentRef = firestore.collection('col').doc();
+   * let documentRef = firestore.doc('col/doc');
    *
    * bulkWriter
    *  .delete(documentRef)
@@ -398,7 +394,7 @@ export class BulkWriter {
    *
    * @example
    * let bulkWriter = firestore.bulkWriter();
-   * let documentRef = firestore.collection('col').doc();
+   * let documentRef = firestore.doc('col/doc');
    *
    * bulkWriter
    *  .update(documentRef, {foo: 'bar'})
@@ -565,15 +561,11 @@ export class BulkWriter {
    * After a batch is complete, try sending batches again.
    */
   private sendBatches(): void {
-    if (this.batchQueue.length === 0) {
-      return;
-    }
-
-    let index = 0;
     const unsentBatches = this.batchQueue.filter(
       batch => batch.sendTime === null
     );
 
+    let index = 0;
     while (
       unsentBatches.length > index &&
       this.isBatchSendable(unsentBatches[index])
@@ -619,7 +611,7 @@ export class BulkWriter {
       }
     }
 
-    if (this.options && this.options.disableThrottling) {
+    if (this.options && !this.options.throttlingEnabled) {
       return true;
     }
 
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -13,10 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as assert from 'assert';
 
-import {FieldPath, Firestore, Timestamp} from '.';
+import {FieldPath, Firestore} from '.';
+import {delayExecution} from './backoff';
 import {DocumentReference} from './reference';
-import {DocumentData, Precondition, SetOptions, UpdateData} from './types';
+import {Timestamp} from './timestamp';
+import {
+  BulkWriterOptions,
+  DocumentData,
+  Precondition,
+  SetOptions,
+  UpdateData,
+} from './types';
 import {Deferred} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
 
@@ -26,20 +35,27 @@ import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
 const DEFAULT_BULK_WRITER_MAX_BATCH_SIZE = 500;
 
 /**
+ * The starting maximum number of operations per second as allowed by the
+ * 5/5/5 rule.
+ */
+const STARTING_MAXIMUM_OPS_PER_SECOND = 500;
+
+/**
  * Used to represent a batch on the BatchQueue.
  *
  * @private
  */
 class BulkCommitBatch {
-  // The maximum number of writes that can be in a single batch.
-  private maxBatchSize = DEFAULT_BULK_WRITER_MAX_BATCH_SIZE;
-
   // The set of references present in the WriteBatch.
   readonly refsInBatch = new Set<DocumentReference>();
 
-  // Whether the batch has been marked to be sent. Writes cannot be added to
-  // pending batches.
-  private pending = false;
+  // Whether the batch is ready to be sent. Writes cannot be added once a batch
+  // is marked as pending.
+  private _pending = false;
+
+  // When the batch was sent. Writes cannot be added to batches that have
+  // already been sent.
+  private _sendTime: Timestamp | null = null;
 
   // A deferred promise that is resolved after the batch has been sent, and a
   // response is received.
@@ -48,7 +64,23 @@ class BulkCommitBatch {
   // A map from each WriteBatch operation to its corresponding result.
   private resultsMap = new Map<number, Deferred<BatchWriteResult>>();
 
-  constructor(readonly id: number, private readonly writeBatch: WriteBatch) {}
+  constructor(
+    readonly id: number,
+    private readonly writeBatch: WriteBatch,
+    private readonly maxBatchSize: number
+  ) {}
+
+  get sendTime(): Timestamp | null {
+    return this._sendTime;
+  }
+
+  get opCount(): number {
+    return this.writeBatch.opCount;
+  }
+
+  get pending(): boolean {
+    return this._pending;
+  }
 
   /**
    * Adds a `create` operation to the WriteBatch. Returns a promise that
@@ -73,7 +105,7 @@ class BulkCommitBatch {
    */
   delete(
     documentRef: DocumentReference,
-    precondition: Precondition
+    precondition?: Precondition
   ): Promise<WriteResult> {
     this.refsInBatch.add(documentRef);
     this.writeBatch.delete(documentRef, precondition);
@@ -128,7 +160,7 @@ class BulkCommitBatch {
    */
   canAddDoc(documentRef: DocumentReference): boolean {
     return (
-      this.pending === false &&
+      !this.pending &&
       this.writeBatch.opCount < this.maxBatchSize &&
       !this.refsInBatch.has(documentRef)
     );
@@ -139,7 +171,8 @@ class BulkCommitBatch {
    * all writes in this batch.
    */
   bulkCommit(): Promise<BatchWriteResult[]> {
-    this.pending = true;
+    assert(this._sendTime === null, 'The batch should not have been sent yet.');
+    this._sendTime = Timestamp.now();
     return this.writeBatch.bulkCommit();
   }
 
@@ -172,12 +205,8 @@ class BulkCommitBatch {
    * response is received.
    */
   awaitBatch(): Promise<void> {
+    this._pending = true;
     return this.completedDeferred.promise;
-  }
-
-  // Visible for testing.
-  setMaxBatchSize(size: number): void {
-    this.maxBatchSize = size;
   }
 }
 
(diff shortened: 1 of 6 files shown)


dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   */
+  private refsInFlight = new Set<DocumentReference>();
+  constructor(
+    private readonly firestore: Firestore,
+    private readonly maxBatchSize = BULK_WRITER_MAX_BATCH_SIZE
+  ) {}
+
+  /**
+   * Create a document with the provided object values. This will fail the batch
+   * if a document exists at its location.
```
schmidt-sebastian: It will only fail this single operation.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * A set of all document references that are currently in flight.
+   */
+  private refsInFlight = new Set<DocumentReference>();
+  constructor(
+    private readonly firestore: Firestore,
+    private readonly maxBatchSize = BULK_WRITER_MAX_BATCH_SIZE
+  ) {}
+
+  /**
+   * Create a document with the provided object values. This will fail the batch
```
schmidt-sebastian: `with the provided data`?

I would try to copy as much as you can from the existing WriteBatch API documentation.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   *
+   * bulkWriter
+   *  .create(documentRef, {foo: 'bar'})
+   *  .then(result => {
+   *    console.log('Successfully executed write at: ', result);
+   *  })
+   *  .catch(err => {
+   *    console.log('Write failed with: ', err);
+   *  });
+   * });
```
schmidt-sebastian: Thanks for adding these examples!

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+
+  /**
+   * Return the first eligible batch that can hold a write to the provided
+   * reference, or creates one if no eligible batches are found.
+   */
+  private getEligibleBatch(ref: DocumentReference): BulkCommitBatch {
+    let batch = null;
+    let logWarning = true;
+    for (const bulkBatch of this.batchQueue) {
+      if (bulkBatch.refsInBatch.has(ref)) {
```
schmidt-sebastian: Should this be inverted?

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+        logger(
+          'BulkWriter',
+          null,
+          'Writing to the same document multiple times will slow down BulkWriter. ' +
+            'Write to unique documents in order to maximize throughput.'
+        );
+      }
+      return this.createNewBatch();
+    } else {
+      return batch;
```
schmidt-sebastian: You could potentially add the document to `refsInBatch` here, which would remove it from all callsites.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * the write. Throws an error if the write fails.
+   *
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   * let documentRef = firestore.collection('col').doc();
+   *
+   * bulkWriter
+   *  .delete(documentRef)
+   *  .then(result => {
+   *    console.log('Successfully executed write at: ', result);
```
schmidt-sebastian: s/executed write/deleted document/

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   *  })
+   *  .catch(err => {
+   *    console.log('Write failed with: ', err);
```
schmidt-sebastian: s/Write/Delete/

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+      bulkCommitBatch.tag,
+      bulkCommitBatch.writeBatch.opCount
+    );
+  }
+
+  /**
+   * Returns a Promise that resolves when there are no more pending writes. It
+   * never fails.
+   *
+   * The promise resolves immediately if there no pending writes. Otherwise, the
```
schmidt-sebastian: Missing "are"

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * pending writes.
+   *
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   *
+   * bulkWriter.create(documentRef, {foo: 'bar'});
+   * bulkWriter.update(documentRef2, {foo: 'bar'});
+   * bulkWriter.delete(documentRef3);
+   * await waitForPendingWrites().then(() => {
+   *   console.log('Successfully executed all writes');
```
schmidt-sebastian: Remove `Successfully`

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * });
+   */
+  async waitForPendingWrites(): Promise<void> {
+    const trackedBatches = this.batchQueue;
+    this.sendBatches();
```
schmidt-sebastian: You should also mention that this will kick off the requests.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+    const writePromises = trackedBatches.map(batch => {
+      batch.pending = true;
+      return this.resultsMap.get(batch.tag)!.promise;
+    });
+    await Promise.all(writePromises);
```
schmidt-sebastian: Nit: If you are using `await` only in the last line, you should just use `return`.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * bulkWriter.delete(documentRef3);
+   * await waitForPendingWrites().then(() => {
+   *   console.log('Successfully executed all writes');
+   * });
+   */
+  async waitForPendingWrites(): Promise<void> {
+    const trackedBatches = this.batchQueue;
+    this.sendBatches();
+    const writePromises = trackedBatches.map(batch => {
+      batch.pending = true;
```
schmidt-sebastian: Doesn't this need to call commit?

dev/src/bulk-writer.ts
schmidt-sebastian: Also: Promise.all() fails if a single Promise gets rejected.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+        bulkBatch.writeBatch.opCount < BULK_WRITER_MAX_BATCH_SIZE &&
+        !bulkBatch.refsInBatch.has(ref)
+      ) {
+        batch = bulkBatch;
+      }
+    }
+
+    if (batch === null) {
+      if (logWarning) {
+        logger(
```
schmidt-sebastian: We could consider logging this to console.warn, since `logger` is off by default.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+  private createNewBatch(): BulkCommitBatch {
+    const tag = requestTag();
+    const deferred = new Deferred<BatchWriteResult[]>();
+    const newBatch = {
+      writeBatch: this.firestore.batch(),
+      refsInBatch: new Set(),
+      tag,
+      pending: false,
+      result: deferred,
+    } as BulkCommitBatch;
```
schmidt-sebastian: I have a feeling that your logic would be a bit cleaner if `BulkCommitBatch` was a class that had methods to modify its internal state. Did you consider that?

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+
+    this.batchQueue.push(newBatch);
+    this.resultsMap.set(tag, deferred);
```
schmidt-sebastian: `tag` is not guaranteed to be unique. Please come up with another way to do this.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+  ): Promise<WriteResult> {
+    const bulkCommitBatch = this.getEligibleBatch(documentRef);
+    bulkCommitBatch.refsInBatch.add(documentRef);
+    bulkCommitBatch.writeBatch.update(
+      documentRef,
+      dataOrField,
+      preconditionOrValues
+    );
+
+    return this.getResult(
```
schmidt-sebastian: What do you think of making a WriteBatch that returns a Promise for each operation? This should simplify your tracking here and remove some of the state from here.

You can take a look at what Java does: https://googleapis.dev/java/google-cloud-clients/latest/com/google/cloud/firestore/UpdateBuilder.html

This is a super class that Transaction and WriteBatch use. You could use an UpdateBuilder that returns Promises, and then ignore this in the WriteBatch class.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * batch cannot be sent.
+   *
+   * After a batch is complete, try sending batches again.
+   */
+  private sendBatches(): void {
+    let counter = 0;
+    for (const batch of this.batchQueue) {
+      // Only send up until the first batch that cannot be sent.
+      if (!this.isBatchSendable(batch)) {
+        break;
```
schmidt-sebastian: Could this be `continue`?

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,428 @@
...
+   * Checks that the provided batch does not write to references that are
+   * currently in flight.
+   */
+  private isBatchSendable(batch: BulkCommitBatch): boolean {
+    for (const ref of batch.refsInBatch) {
+      if (this.refsInFlight.has(ref)) {
+        return false;
+      }
+    }
+    return true;
```
rafikhan: This can be done more idiomatically through the use of `Array.any()`. i.e...
```
return batch.refsInBatch.some(ref => this.refsInFlight.has(ref));
```

dev/src/bulk-writer.ts
rafikhan: When reading this I had the same pause @schmidt-sebastian had. I think there are more idiomatic ways to express this idea (i.e. process the first N elements that meet a certain condition).  This would make more sense as a `while()` loop or perhaps a combination of `Array.prototype.findIndexOf()` and `Array.prototype.slice()`.  Since you ultimately want to return the elements processed I'd suggest with the latter. Plus it lets you get rid of the `counter` variable entirely.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,497 @@
...
+   */
+  private sendBatches(): void {
+    let counter = 0;
+    for (const batch of this.batchQueue) {
+      // Only send up until the first batch that cannot be sent.
+      if (!this.isBatchSendable(batch)) {
+        break;
+      }
+
+      this.refsInFlight = new Set([...this.refsInFlight, ...batch.refsInBatch]);
```
rafikhan, line 466: Declaring a new `Set` and using the spread operator is obscuring the intent of what you're trying to communicate and could be done more efficiently. What you're ultimately trying to do is append `refsInFlight` with the contents in the batch. With `Set`s this is accomplished through the `add()` method. You're also allocating new memory with no clear benefit.

I'd suggest updating this to something more like `batch.refsInBatch.forEach(this.refsInFlight.add);`.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,497 @@
...
+
+      batch.bulkCommit().then(async results => {
+        batch.resolveResults(results);
+
+        // Remove references that are no longer in flight.
+        this.refsInFlight = new Set(
+          [...this.refsInFlight].filter(ref => !batch.refsInBatch.has(ref))
+        );
```
rafikhan, line 474: See my comment from above. This is the same situation but for removal.  As your comment states you're ultimately removing the refs from the batch from those that are in flight. You can express that more literally through something like `batch.refsInBatch.forEach(b => this.refsInFlight.remove(b))`.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,497 @@
...
+      !this.refsInBatch.has(documentRef)
+    );
+  }
+
+  /**
+   * Commits the batch and returns a promise that resolves with the result of
+   * all writes in this batch.
+   */
+  bulkCommit(): Promise<BatchWriteResult[]> {
+    this.pending = true;
```
rafikhan, line 142: Suggestion: Add an assertion that makes sure you're not already pending.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,663 @@
...
+   * delete.
+   * @param {Timestamp=} precondition.lastUpdateTime If set, enforces that the
+   * document was last updated at lastUpdateTime. Fails the batch if the
+   * document doesn't exist or was last updated at a different time.
+   * @returns {Promise<WriteResult>} A promise that resolves with the result of
+   * the write. Throws an error if the write fails.
+   *
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   * let documentRef = firestore.collection('col').doc();
```
dconeybe, line 307: Since this example is for deleting a document, consider adding a document name argument to the `doc()` method call to demonstrate getting an *existing* document to delete.  As it is, calling `doc()` without an argument returns a document with "an automatically-generated unique ID" (i.e. it does not exist).

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,663 @@
...
+   * @param {...(Precondition|*|string|FieldPath)} preconditionOrValues - An
+   * alternating list of field paths and values to update or a Precondition to
+   * restrict this update
+   * @returns {Promise<WriteResult>} A promise that resolves with the result of
+   * the write. Throws an error if the write fails.
+   *
+   *
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   * let documentRef = firestore.collection('col').doc();
```
dconeybe, line 401: Similar comment to my comment above for `delete()`:  Since `update()` requires an existing document, consider adding an argument to the `doc()` method call to demonstrate the typical case of getting an *existing* document to update.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,663 @@
...
+  }
+
+  /**
+   * Attempts to send batches starting from the front of the BatchQueue until a
+   * batch cannot be sent.
+   *
+   * After a batch is complete, try sending batches again.
+   */
+  private sendBatches(): void {
+    if (this.batchQueue.length === 0) {
```
dconeybe, line 568: Does this short-circuit "if" statement provide meaningful savings?  It looks like the algorithm below will gracefully handle an empty batchQueue, nullifying the benefits of this short circuit.  If that is the case, I'd recommend removing this "if" statement because its presence IMO incorrectly implies that it provides a meaningful performance benefit and therefore detracts from readability.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,663 @@
...
+      return;
+    }
+
+    let index = 0;
```
dconeybe, line 572: nit: Consider moving the declaration of "index = 0" to immediately before the "while" loop, closer to where it is initially used.  Its presence here made me look at the code that initializes `unsentBatches` for usages of `index`, which, IMO, detracted from readability.

dev/src/write-batch.ts
```diff
@@ -162,6 +164,13 @@ export class WriteBatch {
     return this._ops.length === 0;
   }
 
+  /**
+   * @private
+   */
+  get opCount(): number {
```
dconeybe, line 170: Is the opCount is used for anything other than an "index" or "key" of a `WriteBatch` in `BulkCommitBatch.resultsMap`?  I was kind of confused when trying to understand its usage in `BulkCommitBatch` because if the opCount is just a convenient monotonically-increasing ID then I'd recommend generating your own ID in BulkCommitBatch to improve readability.

dev/src/types.ts
```diff
@@ -321,6 +321,15 @@ export interface ValidationOptions {
...
 }
 
+/**
+ * An options object that can be used to toggle the default rate limit
+ * throttling implemented by BulkWriter.
+ */
+export interface BulkWriterOptions {
+  /** Whether to disable throttling as specified by the 500/50/5 rule. */
+  readonly disableThrottling: boolean;
+}
```
dconeybe, line 331: nit: Consider renaming this variable to its positive form, namely `enableThrottling` or `throttlingEnabled`.  This generally improves code readability because it reads more naturally than a double-negative.  For example, to determine if throttling is enabled you have to mentally compute `!disableThrottling`, which reads less naturally than just `enableThrottling`.

dev/src/write-batch.ts
```diff
@@ -124,6 +124,8 @@ interface WriteOp {
   precondition?: api.IPrecondition | null;
 }
 
+export class UpdateBuilder {}
```
dconeybe, line 127: I don't see any usages of `UpdateBuilder`.  If this is unused, can it be deleted?

dev/src/write-batch.ts
thebrianchen: Good point! I originally exposed `opCount` in `WriteBatch` to avoid having an extra field on `BulkCommitBatch`, but I didn't consider readability.

dev/src/types.ts
thebrianchen: I was confused on this one because throttling is enabled by default, and I didn't want to confuse developers into thinking that was something they had to turn on. I like the name `throttlingEnabled` since that name eliminates the confusion.

dev/src/bulk-writer.ts

dev/src/bulk-writer.ts

dev/src/bulk-writer.ts
thebrianchen: Ended up using a `while()` loop since I am removing the batch from the BatchQueue after the batch is committed instead of using `slice()` to remove the Batches are the end of the loop.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,655 @@
...
+      batch.refsInBatch.forEach(batch => {
+        this.refsInFlight.add(batch);
+      });
+
+      batch.bulkCommit().then(results => {
+        batch.resolveResults(results);
+
+        const batchIndex = this.batchQueue.indexOf(batch);
+        assert(batchIndex !== -1, 'The batch should be in the BatchQueue');
+        this.batchQueue.splice(batchIndex, 1);
```
schmidt-sebastian, line 583: I would prefer if we simplified this logic to send all batches up until the point where we cannot send the next batch. We are already logging a warning - that should be enough to discourage this usage pattern.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,655 @@
...
+
+    if (batch === null) {
+      if (logWarning) {
+        console.warn(
+          '[BulkWriter]',
+          'Writing to the same document multiple times will slow down BulkWriter. ' +
+            'Write to unique documents in order to maximize throughput.'
+        );
+      }
+      return this.createNewBatch();
```
schmidt-sebastian, line 530: If we create a new batch, we should immediately send the the old batch. "flush()" is only an optimization in case the user wants to send off a smaller batch.

dev/src/bulk-writer.ts
```diff
@@ -0,0 +1,655 @@
...
+      return false;
+    }
+
+    for (const ref of batch.refsInBatch) {
+      if (this.refsInFlight.has(ref)) {
+        return false;
+      }
+    }
+
+    if (this.options && !this.options.throttlingEnabled) {
```
schmidt-sebastian, line 614: You should copy `throttlingEnabled` into local state, otherwise the user can modify  `throttlingEnabled` after the fact.

dev/src/index.ts
```diff
@@ -640,6 +641,30 @@ export class Firestore {
...
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   *
+   * bulkWriter.create(firestore.doc('col/doc1'), {foo: 'bar'})
+   *   .then(res => {
+   *     console.log(`Added document at ${res.writeTime}`);
+   *   });
+   * bulkWriter.update(firestore.doc('col/doc2), {foo: 'bar'});
+   * bulkWriter.delete(firestore.doc('col/doc3'));
+   * await flush().then(() => {
```
schmidt-sebastian, line 660: Nit: Attach a handler to all operations to have more consistency.

This should also call "close()".

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,457 @@
...
+
+    afterEach(() => setTimeoutHandler(setTimeout));
+
+    it('gradually ramps up the maximum operations per second', () => {
+      const nop = () => {};
+      bulkWriter.set(firestore.doc('col/doc'), {});
+
+      // TODO: Should I finish the test and add hooks to monitor that the maxOps
+      // increases by 50% each time? It seems like a quite a bit of boilerplate to test
+      // something pretty simple. On the other hand, it feels bad to not test it at all.
```
schmidt-sebastian, line 454: Yes you should:) I think getting this right will be quite tricky and tests will help.

Issue #990, opened 2020-03-28
.where(admin.firestore.FieldPath.documentId(), 'in', ['id1', 'id2']) fails: '... FieldPath.documentId() must be a string or a DocumentReference'


xaphod:
#### Environment details

  - OS: Firebase Function NodeJS, running on the emulator from Firebase CLI 7.16.1
  - Node.js version: 13.11.0
  - `@google-cloud/firestore` version: 3.7.2
  - `firebase-admin` version: 8.10.0

#### Steps to reproduce

```
    const admin = require('firebase-admin');
    const arr = ['id1', 'id2'];
    const m = await db
      .collection('media')
      .where(admin.firestore.FieldPath.documentId(), 'in', arr)
      .get();
```

Expected: able to query a list of document IDs in a single collection.

Actual: 

```
Error: The corresponding value for FieldPath.documentId() must be a string or a DocumentReference.
    at CollectionReference.validateReference (node_modules/@google-cloud/firestore/build/src/reference.js:1325:19)
    at CollectionReference.where (node_modules/@google-cloud/firestore/build/src/reference.js:1038:26)
```

Does anyone know a workaround? Thanks! (edited several times for clarity)

Padreco-lbr:
Happening to me too, if not an issue, please provide information ou how to make a query based on a list of document Ids. There simply is no documentation examples on that.

schmidt-sebastian:
I will get this fixed today.

xaphod:
Thanks!

pennjlin:
Looks like this issue persists when running with Firebase local emulator. Is there a way to verify the package version used in the emulator? @schmidt-sebastian

schmidt-sebastian:
You can always use `npm list` to verify the installed version. In general, removing the package.lock and `node_modules` followed by a re-install should get you the latest version.

ebeloded:
My `firebase-admin` is up to date, yet I am still running into this exact issue. Any suggestions?

ebeloded:
After looking at the tests, I tried to use `@google-cloud/firestore` instead of `firebase-admin`, and it worked!

This works:
```typescript
import { FieldPath, Firestore } from '@google-cloud/firestore'

const firestore = new Firestore()

const menus = await firestore
  .collection('menus')
  .where(FieldPath.documentId(), 'in', ['cUKi4Kj42LOcLsqBZOBhM'])
  .get()

```

This doesn't:
```typescript
import admin from 'firebase-admin'

[...]

PR #993, opened 2020-03-31
fix: support array of references for IN queries

schmidt-sebastian:
This ports the validation from Web.

Fixes https://github.com/googleapis/nodejs-firestore/issues/990

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -1194,7 +1194,6 @@ describe('Query class', () => {
   });
 
   it('supports "in" with document ID array', async () => {
-    setLogFunction(console.log);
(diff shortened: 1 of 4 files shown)


schmidt-sebastian:
@thebrianchen We may need to port this to Java too.

dev/system-test/firestore.ts
```diff
@@ -1184,6 +1193,15 @@ describe('Query class', () => {
     expectDocs(res, {zip: 98101}, {zip: 98103});
   });
 
+  it('supports "in" with document ID array', async () => {
+    setLogFunction(console.log);
```
thebrianchen: hanging log function?

schmidt-sebastian: Removed

PR #1003, opened 2020-04-08
fix: validate nested arrays in FieldValue

### LINKED ISSUE #999: Misleading error when using empty array with arrayUnion
"firebase-admin": "^8.10.0"

#### Steps to reproduce

In a transaction, if using `firestore.FieldValue.arrayUnion(__)` but with argument as empty array (`[]`), the following error is shown
```
     Error: 3 INVALID_ARGUMENT: Cannot convert an array value in an array value.
```

It is misleading as it sounds like I violated array in array restriction in Firestore. But it is not the case. An expected outcome is that it union with empty, resulting in a no-op.

thebrianchen:
Fixes #999.

### DIFF
--- a/dev/test/field-value.ts
+++ b/dev/test/field-value.ts
@@ -231,7 +231,8 @@ describe('FieldValue.arrayRemove()', () => {
     return createInstance().then(firestore => {
       const docRef = firestore.doc('collectionId/documentId');
       expect(() => docRef.set({foo: FieldValue.arrayRemove([])})).to.throw(
-        `Nested arrays are not supported`
+        'Element at index 0 is not a valid array element. Nested arrays are ' +
+          'not supported.'
       );
     });
   });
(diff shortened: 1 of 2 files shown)


dev/test/field-value.ts
```diff
@@ -121,6 +121,15 @@ describe('FieldValue.arrayUnion()', () => {
     });
   });
 
+  it('must not contain directly nested arrays', () => {
+    return createInstance().then(firestore => {
+      const docRef = firestore.doc('collectionId/documentId');
+      expect(() => docRef.set({foo: FieldValue.arrayUnion([])})).to.throw(
+        `Nested arrays are not supported`
```
schmidt-sebastian, line 128: Can you modify the test to assert on the full error message (including 'invalidArgumentMessage'). I wanted to see what the full error message looks like, but I would rather let you do this research and convey the result via this test.

thebrianchen: done.

PR #1006, opened 2020-04-09
feat!: Drop Node 8 support

schmidt-sebastian:


jrthib:
Does this have an estimated release date?

schmidt-sebastian:
Likely later this month

PR #1010, opened 2020-04-11
fix: remove eslint, update gax, fix generated protos, run the generator

alexander-fenster:
Run the latest version of the generator, update google-gax, update gts, and remove direct dependencies on eslint.


alexander-fenster:
Firestore will need more work.

PR #1013, opened 2020-04-13
deps: update to gts 2.x

schmidt-sebastian:
- Updates gts to 2.0
- Removes custom docChanges() error message that discourages API use from the 0.x API (also just removed from Web)
- Removes `test/typescript.ts`, an "API test" that never caught any issues and was never an actual test
- Removes all auto-generated tests rather than fixing their types. The tests should be regenerated before merging the feature branch.
- Replaces `deep-equal` with `fast-deep-equal`, which has published types and fewer dependencies


package.json
```diff
@@ -41,17 +41,17 @@
     "lint": "gts check",
     "clean": "gts clean",
     "compile": "tsc -p .",
-    "postcompile": "node scripts/init-directories.js && cp -r dev/protos build && cp dev/src/v1beta1/*.json build/src/v1beta1/ && cp dev/src/v1/*.json build/src/v1/ && cp dev/conformance/test-definition.proto build/conformance && cp dev/conformance/conformance-tests/*.json build/conformance/conformance-tests",
+    "postcompile": "mv build/dev/* build/ && rm -r build/dev && node scripts/init-directories.js && cp -r dev/protos build && cp dev/src/v1beta1/*.json build/src/v1beta1/ && cp dev/src/v1/*.json build/src/v1/ && cp dev/conformance/test-definition.proto build/conformance && cp dev/conformance/conformance-tests/*.json build/conformance/conformance-tests",
```
schmidt-sebastian, line 44: For some reason, the new settings now put the source files under `build/dev/src` instead of `build/src`. This mitigates this, but there might be better ways,

JustinBeckwith:
@schmidt-sebastian you are probably going to need to remove `8` from `.github/workflows/ci.yaml`.  It's happening elsewhere in an autosynth PR, but they'd both need to land together, or this one weird trick should help kick CI.  GitHub actions bails by default when the first test suite fails.

schmidt-sebastian:
@JustinBeckwith Thanks for your review. I was planning to run synthtool after merging this PR.

PR #1014, opened 2020-04-13
deps: drop through2

schmidt-sebastian:
Drops dependency on through2 (though dev dependency remains). Also drops readable-stream as Node10 supports async iterator out of the box.


schmidt-sebastian:
IT'S HAPPENING NOW!

Issue #872, opened 2020-01-12
64-bit integers lose precision


MegaByte:
#### Environment details

  - OS: GCF
  - Node.js version: 8.15.0/10.15.3

#### Steps to reproduce

  1. Have data using 64-bit integers with values > 2^53 stored using a non-JS client (e.g. including `5248794116772919364` and larger values)
  2. Read the data.
  3. Use an `orderBy` keyed on that field with a `startAfter` value > 2^53 (e.g. `5248794116772919364`)

#### Actual results:
  1. The reported value is rounded. The document is returned with `5248794116772919000`.
  2. The cursor value is rounded. The document including the `5248794116772919364` value is returned (with `5248794116772919000` as above).

#### Expected results:
  1. The value should not be rounded.
  2. The document with the `5248794116772919364` value should not be returned.

In addition, it is not possible to use an alternate type to avoid the numeric rounding. Using a `string` in `startAfter` results in no data returned at all. Using a `BigInt` (Node 10 only) gives the error: 
> Unexpected value type

See also https://github.com/protocolbuffers/protobuf/issues/3666

schmidt-sebastian:
@MegaByte Thanks for reporting. This is currently working as intended. We rely on the Number semantics of JavaScript. We may evaluate adding support for `BigInt` once we phase out Node 8 support. Support for `BigInt` is however not currently on our roadmap.

cubuspl42:
How is it possible, that the official SDK for accessing Firestore is loosing information, and that's an intended behaviour?

It's not only about precision. Currently, it's absolutely impossible to even learn that the fields stored in database is actually an Integer, or write an Integer to a database.

I believe that it's an officially supported to access Firestore database from multiple platforms. For example, when I'm developing a mobile app in Flutter (Dart) and want to provide some backend logic for it, isn't it an supported, or even encouraged, solution to use Google Cloud Functions with Javascript SDK?

When my application stores an Integer field in the database, how am I supposed to update it / export it? It's beyond inconvenient. It makes a simple task impossible.

At very least, you could box all Integers with a helper `Integer` class, a bit like it works with Timestamps. It wouldn't really solve the precision issue (9 bits of precision would be lost), but it would allow to read/write an officially supported data type using an official SDK. That's a start.

schmidt-sebastian:
We are working on dropping Node 8 support across all GCloud SDKs. Once Firestore has been migrated, we can add support for Node 10's BigInt.

[...]

PR #1016, opened 2020-04-14
 feat: support BigInt

schmidt-sebastian:
This PR adds support for serializing to BigInt and for deserializing to BigInt (optional, via flag). We still need to settle on a flag name.

Fixes https://github.com/googleapis/nodejs-firestore/issues/872

### DIFF
--- a/dev/protos/update.sh
+++ b/dev/protos/update.sh
@@ -83,8 +83,7 @@ PBJS_ARGS=( --proto_path=. \
   --no-decode \
   --no-verify \
   --no-delimited \
-  --force-enum-string \
-  --force-number)
+  --force-enum-string)
       
 "${PBJS}" "${PBJS_ARGS[@]}" -o firestore_v1_proto_api.js \
   -r firestore_v1 \
@@ -92,6 +91,7 @@ PBJS_ARGS=( --proto_path=. \
   "${PROTOS_DIR}/google/protobuf/*.proto" "${PROTOS_DIR}/google/type/*.proto" \
   "${PROTOS_DIR}/google/rpc/*.proto" "${PROTOS_DIR}/google/api/*.proto" \
   "${PROTOS_DIR}/google/longrunning/*.proto"
+sed -i .bak 's:number|Long:number|string:g' firestore_v1_proto_api.js
 "${PBTS}" -o firestore_v1_proto_api.d.ts firestore_v1_proto_api.js
 
 "${PBJS}" "${PBJS_ARGS[@]}" -o firestore_admin_v1_proto_api.js \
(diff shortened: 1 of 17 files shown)


dev/protos/update.sh
```diff
@@ -83,15 +83,15 @@ PBJS_ARGS=( --proto_path=. \
   --no-decode \
   --no-verify \
   --no-delimited \
-  --force-enum-string \
-  --force-number)
+  --force-enum-string)
```
schmidt-sebastian, line 86: I was not able to get the ProtobufJS compiler to use "string|number" in its types. It only offers "Long", "Long|number" and "number". So I chose to use "Long|number" and replace it manually with "string|number". int32 (like nanoseonds) are not affected since they can be represented as number.

dev/src/convert.ts
```diff
@@ -177,16 +177,12 @@ export function valueFromJson(fieldValue: api.IValue): api.IValue {
       return {
         bytesValue: bytesFromJson(fieldValue.bytesValue!),
       };
-    case 'integerValue':
```
schmidt-sebastian, line 180: I dropped this since the client supports "string|number" throughout. Converting to Number results in precision loss.

dev/src/types.ts
```diff
@@ -199,6 +199,14 @@ export interface Settings {
    */
   maxIdleChannels?: number;
 
+  /**
+   * Whether to use `BigInt` for integer types when deserializing Firestore
+   * Documents. Regardless of magnitude, all integer values will be returned as
+   * `BigInt` to match the precision of the Firestore backend. Floating point
+   * numbers continue to use JavaScript's `number` type.
```
thebrianchen, line 206: maybe s/continue to/will for consistent future tense?

dev/src/serializer.ts
```diff
@@ -53,12 +52,15 @@ export interface Serializable {
...
 export class Serializer {
   private createReference: (path: string) => DocumentReference;
+  private createInteger: (n: number | string) => number | BigInt;
 
   constructor(firestore: Firestore) {
     // Instead of storing the `firestore` object, we store just a reference to
     // its `.doc()` method. This avoid a circular reference, which breaks
     // JSON.stringify().
     this.createReference = path => firestore.doc(path);
+    this.createInteger = n =>
```
thebrianchen, line 62: Should we validate the string so that `Number(n)` won't break on bad inputs?

schmidt-sebastian: These values come from the backend and IIRC any issues with it so far (we already used this code below). Opting not do this for now.

dev/src/types.ts
schmidt-sebastian: Opted to change this to present tense for both.

PR #1018, opened 2020-04-14
feat: Add BulkWriter base

thebrianchen:
Moving from the old [PR](https://github.com/googleapis/nodejs-firestore/pull/988) to create a base version without bin packing of writes or the 555 throttling.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -479,11 +479,14 @@ export class BulkWriter {
    */
   private getEligibleBatch(ref: DocumentReference): BulkCommitBatch {
     let eligibleBatch = null;
-    let hasBatches = this.batchQueue.length > 0;
-    if (hasBatches) {
+    let logWarning = false;
+    if (this.batchQueue.length > 0) {
       const lastBatch = this.batchQueue[this.batchQueue.length - 1];
-      if (!lastBatch.docPaths.has(ref.path)) {
-        hasBatches = false;
+      if (
+        lastBatch.docPaths.has(ref.path) &&
+        lastBatch.state === BatchState.OPEN
+      ) {
+        logWarning = true;
       }
       if (lastBatch.canAddDoc(ref)) {
         eligibleBatch = lastBatch;
@@ -491,7 +494,7 @@ export class BulkWriter {
     }
 
     if (eligibleBatch === null) {
-      if (hasBatches) {
+      if (logWarning) {
         console.warn(
           '[BulkWriter]',
           `Duplicate write to document "${ref.path}" detected.`,
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -23,7 +23,7 @@ import {Deferred} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
 
 /**
- * The default maximum number of writes that can be in a single batch.
+ * The maximum number of writes that can be in a single batch.
  */
 const MAX_BATCH_SIZE = 500;
 
@@ -48,9 +48,9 @@ class BulkCommitBatch {
   /**
    * The state of the batch.
    */
-  private _state = BatchState.OPEN;
+  state = BatchState.OPEN;
 
-  // The set of document reference ids present in the WriteBatch.
+  // The set of document reference paths present in the WriteBatch.
   readonly docPaths = new Set<string>();
 
   // A deferred promise that is resolved after the batch has been sent, and a
@@ -72,10 +72,6 @@ class BulkCommitBatch {
    */
   get opCount(): number {
     return this.resultsMap.size;
-  }
-
-  get state(): BatchState {
-    return this._state;
   }
 
   /**
@@ -142,9 +138,14 @@ class BulkCommitBatch {
       'Batch should not contain writes to the same document'
     );
     this.docPaths.add(documentRef.path);
-    this.resultsMap.set(this.opCount, new Deferred<BatchWriteResult>());
+    const deferred = new Deferred<BatchWriteResult>();
+    this.resultsMap.set(this.opCount, deferred);
 
-    return this.resultsMap.get(this.opCount - 1)!.promise.then(result => {
+    if (this.opCount === this.maxBatchSize) {
+      this.state = BatchState.READY_TO_SEND;
+    }
+
+    return deferred.promise.then(result => {
       if (result.writeTime) {
         return new WriteResult(result.writeTime);
       } else {
@@ -154,27 +155,15 @@ class BulkCommitBatch {
   }
 
   /**
-   * Returns whether a write containing the provided document reference can be
-   * added to this batch.
-   */
-  canAddDoc(documentRef: DocumentReference): boolean {
-    return (
-      this._state === BatchState.OPEN &&
-      this.opCount < this.maxBatchSize &&
-      !this.docPaths.has(documentRef.path)
-    );
-  }
-
-  /**
    * Commits the batch and returns a promise that resolves with the result of
    * all writes in this batch.
    */
   bulkCommit(): Promise<BatchWriteResult[]> {
     assert(
-      this._state === BatchState.READY_TO_SEND,
+      this.state === BatchState.READY_TO_SEND,
       'The batch should be marked as READY_TO_SEND before committing'
     );
-    this._state = BatchState.SENT;
+    this.state = BatchState.SENT;
     return this.writeBatch.bulkCommit();
   }
 
@@ -198,8 +187,8 @@ class BulkCommitBatch {
   }
 
   markReadyToSend(): void {
-    if (this._state === BatchState.OPEN) {
-      this._state = BatchState.READY_TO_SEND;
+    if (this.state === BatchState.OPEN) {
+      this.state = BatchState.READY_TO_SEND;
     }
   }
 }
@@ -260,7 +249,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch(documentRef);
     const resultPromise = bulkCommitBatch.create(documentRef, data);
-    this.sendBatchIfFull(bulkCommitBatch);
+    this.sendReadyBatches();
     return resultPromise;
   }
 
@@ -298,7 +287,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch(documentRef);
     const resultPromise = bulkCommitBatch.delete(documentRef, precondition);
-    this.sendBatchIfFull(bulkCommitBatch);
+    this.sendReadyBatches();
     return resultPromise;
   }
 
@@ -344,7 +333,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch(documentRef);
     const resultPromise = bulkCommitBatch.set(documentRef, data, options);
-    this.sendBatchIfFull(bulkCommitBatch);
+    this.sendReadyBatches();
     return resultPromise;
   }
 
@@ -403,7 +392,7 @@ export class BulkWriter {
       dataOrField,
       preconditionOrValues
     );
-    this.sendBatchIfFull(bulkCommitBatch);
+    this.sendReadyBatches();
     return resultPromise;
   }
 
@@ -435,7 +424,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const trackedBatches = this.batchQueue;
     const writePromises = trackedBatches.map(batch => batch.awaitBatch());
-    this.sendBatches();
+    this.sendReadyBatches();
     await Promise.all(writePromises);
   }
 
@@ -482,14 +471,12 @@ export class BulkWriter {
     let logWarning = false;
     if (this.batchQueue.length > 0) {
       const lastBatch = this.batchQueue[this.batchQueue.length - 1];
-      if (
-        lastBatch.docPaths.has(ref.path) &&
-        lastBatch.state === BatchState.OPEN
-      ) {
-        logWarning = true;
-      }
-      if (lastBatch.canAddDoc(ref)) {
-        eligibleBatch = lastBatch;
+      if (lastBatch.state === BatchState.OPEN) {
+        if (!lastBatch.docPaths.has(ref.path)) {
+          eligibleBatch = lastBatch;
+        } else {
+          logWarning = true;
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,535 @@
...
+
+  function incrementOpCount(): void {
+    opCount++;
+  }
+
+  function verifyOpCount(expected: number): void {
+    expect(opCount).to.equal(expected);
+  }
+
+  function setOp(doc: string, value: string): {[key: string]: {}} {
```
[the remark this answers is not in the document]
thebrianchen: I tried rewriting to use the helpers. Creating a request that includes multiple operations is a bit hacky since we have to concat all the individual writes from all the created requests. The `update` helper also fills a precondition if none is provided, so I had to manually delete `currentDocument`. If you prefer this layout, I can update the tests in a separate PR to pass in a precondition instead of having it autofill.

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+      document: document(doc, 'foo', value),
+    }) as BulkWriterRequest;
+  }
+
+  function updateOp(doc: string, value: string): BulkWriterRequest {
+    const updateRequest = update({
+      document: document(doc, 'foo', value),
+      mask: updateMask('foo'),
+    }) as BulkWriterRequest;
+    delete updateRequest.writes[0].currentDocument;
```
schmidt-sebastian, line 91: I don't quite understand why we are dropping the Precondition. Can you add a comment?

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+
+  function verifyOpCount(expected: number): void {
+    expect(opCount).to.equal(expected);
+  }
+
+  function setOp(doc: string, value: string): BulkWriterRequest {
+    return set({
+      document: document(doc, 'foo', value),
+    }) as BulkWriterRequest;
+  }
```
schmidt-sebastian, line 78: Have you thought of doing this:

```
 function setOp(doc: string, value: string): api.IWrite {
    return set({
      document: document(doc, 'foo', value),
    }).writes![0];
  }
```

This allows you to use the Proto types throughout your test suite.

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+const PROJECT_ID = 'test-project';
+
+interface BulkWriterRequest {
+  writes: Array<{[key: string]: {} | undefined}>;
+}
+
+interface BulkWriterResponse {
+  writeResults: Array<{[key: string]: {} | null}>;
+  status: Array<{[key: string]: {}}>;
+}
```
schmidt-sebastian, line 46: The way these two types are defined removes all type safety from the tests. Please try to use the Protobuf types as defined in https://github.com/googleapis/nodejs-firestore/blob/master/dev/protos/firestore_v1_proto_api.d.ts

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+    const status = responses.map(response => ({code: response.code!}));
+    return {
+      writeResults,
+      status,
+    };
+  }
+
+  function instantiateInstance(
+    mock: RequestResponse[],
+    manualFlush = false
```
schmidt-sebastian, line 137: Nit: Use a more descriptive name or add a comment on what this does.

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+  ): Promise<void> {
+    const overrides: ApiOverride = {
+      batchWrite: async request => {
+        expect(request).to.deep.eq({
+          database: `projects/${PROJECT_ID}/databases/(default)`,
+          writes: mock[requestCounter].request.writes,
+        });
+        if (manualFlush) {
+          flushCounter++;
```
schmidt-sebastian, line 146: activeFlushCounter? activeRequestCounter?

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+        const responsePromise = response({
+          writeResults: mock[requestCounter].response.writeResults,
+          status: mock[requestCounter].response.status,
+        });
+        requestCounter++;
+        return responsePromise;
+      },
+    };
+    return createInstance(overrides).then(firestoreClient => {
+      firestore = firestoreClient;
```
schmidt-sebastian, line 163: Have you considered returning `firestore` instead of setting global state? The consuming code would look more natural:

```
  const firestore = await instantiateInstance([
      {
        request: setOp('doc', 'bar'),
        response: createResponse([successResponse(2)]),
      },
    ]);
}

```

or:

```
  const {firestore, bulkWriter}  = await instantiateInstance([
      {
        request: setOp('doc', 'bar'),
        response: createResponse([successResponse(2)]),
      },
    ]);
```

dev/test/bulk-writer.ts
```diff
@@ -0,0 +1,502 @@
...
+        request: updateOp('doc', 'bar1'),
+        response: createResponse([successResponse(1)]),
+      },
+    ]);
+
+    const doc = firestore.doc('collectionId/doc');
+    bulkWriter.set(doc, {foo: 'bar'}).then(incrementOpCount);
+    bulkWriter.update(doc, {foo: 'bar1'}).then(incrementOpCount);
+
+    return bulkWriter.close().then(async () => {
```
schmidt-sebastian, line 340: This test would actually have passed with your original logic that was based on `Set<DocumentReference>`. To test that your are correctly comparing references, you can change this to use with two different document references that point to the same document.

dev/test/bulk-writer.ts
thebrianchen: refactored, so much cleaner.

dev/test/bulk-writer.ts
thebrianchen: done.

dev/test/bulk-writer.ts
```diff
@@ -125,31 +125,40 @@ describe('BulkWriter', () => {
     const writeResults = responses.map(response => ({
```
schmidt-sebastian, line 125: Please use proper types for `successResponse`, `failResponse` and `createResponse` (something like ` writeResult: WriteResult, status: Status }`. You could even change successResponse to return a single valid api.iBatchWriteResponse, which you could then use in all cases where only one response is needed. You can do the same for `failResponse`.  `createResponse` would just merge results.

See diff with proper types: https://gist.github.com/schmidt-sebastian/292dcc2bfdc2ad3309e480ed57251a1e

dev/test/bulk-writer.ts
```diff
@@ -125,31 +125,40 @@ describe('BulkWriter', () => {
...
     return {
       writeResults,
       status,
     };
   }
 
+  /**
+   * Creates an instance with the mocked objects.
+   *
+   * @param checkActiveRequests Whether to check for active requests. If true,
```
schmidt-sebastian, line 140: Should this be `checkActiveRequestCount`?  Or even `enforceSingleConcurrentRequest`?

dev/test/bulk-writer.ts
schmidt-sebastian: Sounds good.

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
            precondition?: Precondition): Transaction;
   }
 
+  /**
```
schmidt-sebastian, line 413: Fix leading spaces.

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
...
+ * A Firestore BulkWriter than can be used to perform large amounts of writes in
```
schmidt-sebastian, line 414: Maybe s/large amounts/a large number/

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
...
+   * Create a document with the provided data. This single operation will fail
+   * if a document exists at its location.
+   *
+   * @param {DocumentReference} documentRef A reference to the document to be
+   * created.
+   * @param {T} data The object to serialize as the document.
+   * @returns {Promise<WriteResult>} A promise that resolves with the result of
+   * the write. Throws an error if the write fails.
+   *
+   * @example
```
schmidt-sebastian, line 432: We don't really have examples in this file.

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
...
+   */
+  flush(): Promise<void>;
+
+  /**
+   * Commits all enqueued writes and marks the BulkWriter instance as closed.
+   *
+   * After calling `close()`, calling any method wil throw an error.
+   *
+   * Returns a Promise that resolves when there are no more pending writes. It
+   * never fails. Calling this method will send all requests. The promise
```
schmidt-sebastian, line 604: It never fails => The Promise will never be rejected.

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
...
+   * });
+   */
+  flush(): Promise<void>;
+
+  /**
+   * Commits all enqueued writes and marks the BulkWriter instance as closed.
+   *
+   * After calling `close()`, calling any method wil throw an error.
+   *
+   * Returns a Promise that resolves when there are no more pending writes. It
```
schmidt-sebastian, line 603: "when all writes have been committed"

types/firestore.d.ts
schmidt-sebastian: Please also update docs for `flush()`.

types/firestore.d.ts
```diff
@@ -411,6 +410,216 @@ declare namespace FirebaseFirestore {
...
+   * never fails. Calling this method will send all requests. The promise
+   * resolves immediately if there are no pending writes.
+   *
+   * @return {Promise<void>} A promise that resolves when there are no more
```
schmidt-sebastian, line 607: "when all writes are committed"

types/firestore.d.ts
schmidt-sebastian: Note that we did put the converter example in here because it is only valid in TypeScript and does not show up in JSDoc.

dev/src/bulk-writer.ts
```diff
@@ -174,10 +174,16 @@ class BulkCommitBatch {
   /**
    * Resolves the individual operations in the batch with the results.
    */
-  processResults(results: BatchWriteResult[]): void {
-    results.map((result, index) => {
-      this.resultsMap.get(index)!.resolve(result);
-    });
+  processResults(results: BatchWriteResult[], error?: Error): void {
+    if (error === undefined) {
+      results.map((result, index) => {
```
schmidt-sebastian, line 179: Can you use a `for` loop here as well? `map` creates a copy of the result that you are then throwing away.

dev/src/bulk-writer.ts
```diff
@@ -403,10 +409,10 @@ export class BulkWriter {
   /**
    * Commits all writes that have been enqueued up to this point in parallel.
    *
-   * Returns a Promise that resolves when there are no more pending writes. It
-   * never fails.
+   * Returns a Promise that resolves when there are no more pending writes.
```
schmidt-sebastian, line 412: Maybe:

"Returns a Promise that resolves when all currently queued operations have been committed. The Promise returned from this method will never be rejected as the success for each individual operation is conveyed via their individual Promises. "

We do need to rewrite the "when there are no more pending writes" part, the rest I am on the fence about.

dev/test/bulk-writer.ts
```diff
@@ -96,62 +87,73 @@ describe('BulkWriter', () => {
...
     return {
-      writeResults,
-      status,
+      writeResults: responses.map(v => v.writeResults![0]),
+      status: responses.map(v => v.status![0]),
     };
   }
 
   /**
+   *  /**
```
schmidt-sebastian, line 131: Please fix.

dev/test/bulk-writer.ts
```diff
@@ -96,62 +87,73 @@ describe('BulkWriter', () => {
...
   ): Promise<BulkWriter> {
     const overrides: ApiOverride = {
       batchWrite: async request => {
         expect(request).to.deep.eq({
           database: `projects/${PROJECT_ID}/databases/(default)`,
           writes: mock[requestCounter].request.writes,
         });
-        if (checkActiveRequests) {
+        if (options && options.failResponse) {
+          throw new Error('batchWrite failed');
```
schmidt-sebastian, line 153: As per Gil: Please use 'test' somewhere in the error message. This might save us some debugging time later.

dev/test/bulk-writer.ts
```diff
@@ -510,4 +513,84 @@ describe('BulkWriter', () => {
     await flush2;
     return bulkWriter.close();
   });
+
+  describe('if bulkCommit() fails', async () => {
+    it('flush() should not fail', async () => {
```
schmidt-sebastian, line 518: Since this entire test suite uses a different behavior for `instantiateInstance`, you can simply overwrite it (or use a diffent function name).

https://gist.github.com/schmidt-sebastian/0724808e4f45f356602085cffe90cea9

types/firestore.d.ts
```diff
@@ -411,6 +410,139 @@ declare namespace FirebaseFirestore {
...
+     *
+     * Returns a Promise that resolves when all writes have been committed.
+     * The Promise will never be rejected.
+     *
+     * The Promise resolves immediately if there are no pending writes. Otherwise,
+     * the Promise waits for all previously issued writes, but it does not wait
+     * for writes that were added after the method is called. If you want to wait
+     * for additional writes, call `flush()` again.
+     *
+     * @return {Promise<void>} A promise that resolves when all writes have
```
schmidt-sebastian, line 525: Please sync these docs with the documentation in the code when completed.

dev/src/bulk-writer.ts

dev/test/bulk-writer.ts
thebrianchen: oops

dev/test/bulk-writer.ts
thebrianchen: changed to `mock batchWrite failed in test`

dev/test/bulk-writer.ts
thebrianchen: thanks for the gist!

types/firestore.d.ts
thebrianchen: synced.

dev/src/bulk-writer.ts
thebrianchen: "Returns a Promise that resolves when all currently queued operations have been committed. The Promise will never be rejected since the results for each individual operation are conveyed via their individual Promises."

Renamed return to: `@return {Promise<void>} A promise that resolves when all enqueued writes up to this point have been committed.`

PR #1019, opened 2020-04-14
chore!: update settings for Node 10

schmidt-sebastian:
This PR updates the synth.py script to work with the new format/tooling brought by the Node 10 migration.

Manual changes are in synth.py and  dev/protos/update.sh to enable namespacing and add `toObject` and `fromObject` (which is needed for the generated unit tests and unfortunately adds a lot of code - maybe we can find a way to remove this code in a follow up).


schmidt-sebastian:
Hm, this is completely busted. I will fix this later.

synth.py
```diff
@@ -98,6 +138,13 @@
   1
 )
 
+os.rename("dev/.gitignore", ".gitignore")
```
JustinBeckwith, line 141: Not sure what all goes here, but this may be missing `.prettierignore`

.eslintignore
```diff
@@ -1,5 +1,6 @@
 **/node_modules
-src/**/doc/*
+**/.coverage
 build/
 docs/
 protos/
+system-test/
```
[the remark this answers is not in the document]
schmidt-sebastian: This should be from the tooling. I didn't make this change manually at least :)

synth.py
schmidt-sebastian: Added

schmidt-sebastian: I take that back: That file doesn't exist:

FileNotFoundError: [Errno 2] No such file or directory: 'dev/.prettierignore' -> '.prettierignore'

I guess we should add this to our templates?

schmidt-sebastian:
@alexander-fenster Can you sanity check the manual changes?

.eslintignore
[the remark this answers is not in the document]
alexander-fenster: Hmmm, we don't do it in other libs. Example: https://github.com/googleapis/nodejs-text-to-speech/blob/master/.eslintignore

.jsdoc.js
```diff
@@ -34,18 +34,19 @@ module.exports = {
...
     ],
     includePattern: '\\.js$'
   },
   templates: {
-    copyright: 'Copyright 2019 Google, LLC.',
+    copyright: 'Copyright 2020 Google LLC',
     includeDate: false,
     sourceFiles: false,
-    systemName: '@google-cloud/firestore',
+    systemName: 'firestore',
```
alexander-fenster, line 46: For this, you need to add `"package-name": "@google-cloud/firestore"` to the generator arguments in `synth.py`.

dev/protos/update.sh
```diff
@@ -77,31 +77,33 @@ cp protobuf/src/google/protobuf/{any,empty,field_mask,struct,timestamp,wrappers}
...
 "${PBJS}" "${PBJS_ARGS[@]}" -o firestore_admin_v1_proto_api.js \
+  -r firestore_admin_v1 \
   "${PROTOS_DIR}/google/firestore/admin/v1/*.proto" \
   "${PROTOS_DIR}/google/protobuf/*.proto" "${PROTOS_DIR}/google/type/*.proto" \
   "${PROTOS_DIR}/google/rpc/*.proto" "${PROTOS_DIR}/google/api/*.proto" \
   "${PROTOS_DIR}/google/longrunning/*.proto"
 "${PBTS}" -o firestore_admin_v1_proto_api.d.ts firestore_admin_v1_proto_api.js
 
 "${PBJS}" "${PBJS_ARGS[@]}" -o firestore_v1beta1_proto_api.js \
+  -r firestore_v1beta1_v1 \
```
alexander-fenster, line 106: Does not really affect anything but the name is weird. Maybe just `firestore_v1beta1`?

dev/src/v1/firestore_admin_client.ts
```diff
@@ -1156,29 +1231,83 @@ export class FirestoreAdminClient {
...
+   * @param {string} request.pageToken
+   *   A page token, returned from a previous call to
+   *   {@link google.firestore.admin.v1.FirestoreAdmin.ListIndexes|FirestoreAdmin.ListIndexes}, that may be used to get the next
+   *   page of results.
+   * @param {object} [options]
+   *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+   * @returns {Object}
+   *   An iterable Object that conforms to @link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols.
+   */
+  listIndexesAsync(
```
alexander-fenster, line 1264: FYI @schmidt-sebastian: we now support this :)  `for await (const index of client.listIndexesAsync(request))` - feel free to use this in your manual layer!  (the same for any list* paginated method)

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "author": "Google Inc.",
   "engines": {
-    "node": "^8.13.0 || >=10.10.0"
+    "node": ">=10.10.0"
```
alexander-fenster, line 8: This is so breaking. The title must be `chore!` then :)

synth.py
```diff
@@ -26,23 +26,43 @@
...
-    "tsconfig.json", "tslint.json", "linkinator.config.json", "webpack.config.js"])
+    "tsconfig.json", "linkinator.config.json", "webpack.config.js"])
 
 # Fix dropping of google-cloud-resource-header
 # See: https://github.com/googleapis/nodejs-firestore/pull/375
 s.replace(
     "dev/src/v1beta1/firestore_client.ts",
-    "return this\._innerApiCalls\.listen\(options\);",
-    "return this._innerApiCalls.listen({}, options);",
+    "return this\.innerApiCalls\.listen\(options\);",
```
alexander-fenster, line 39: Every time I see this replacement I wonder what's wrong with the code here, why we need this replacement?

alexander-fenster:
Manual changes are somewhat scary :) A _lot_ of changes are caused by the fact that the gapic code lives in `dev/src/`. I'm just thinking, is the directory structure fixed in stone here? Other manual packages (spanner, pubsub) have their sources in `src/`, and their `synth.py` files are tiny and simple. Maybe not right now, but let's at least have a plan there?

(an alternative is to teach the generator how to place files into a separate location, but I can't estimate how hard it would be)

.jsdoc.js
schmidt-sebastian: Done (and this fixed the issue here)

package.json
schmidt-sebastian: Changed. Note that the base branch has a `!` already.

.eslintignore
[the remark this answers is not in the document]
schmidt-sebastian: I deleted this file and re-ran synthtool and it was recreated just like this.

dev/src/v1/firestore_admin_client.ts
schmidt-sebastian: Oh, awesome. Thanks for pointing this out. We should probably use expose this in our APIs that currently hide all paging, since we didn't want to use Streams or Callbacks as the primary way of interacting with our APIs.

dev/protos/update.sh
schmidt-sebastian: Fixed

synth.py
schmidt-sebastian: I do think the generated code is busted here. We use `options` to set an additional header, but the way this gets passed down here is as the request (if I remember this correctly). If we don't modify the code, the header is dropped. This only affects bidirectional streaming APIs with additional metadata, so probably just us. I do think we should fix this upstream at some point.

alexander-fenster: OK, this sounds like a bug - filed https://github.com/googleapis/gapic-generator-typescript/issues/425 and one of us will investigate.

PR #1021, opened 2020-04-15
feat: add rate limiter class

thebrianchen:
Adding a rate limiter that implements [token bucket](https://www.linkedin.com/pulse/api-rate-limiting-using-token-bucket-algorithm-siddharth-patnaik/) that calculates if a request containing a number of writes can be made under a rate limit as well as the time required before a request can be made while staying under the limit.

It's a little more verbose than I wanted, but having a separate rate limiter has a few benefits for BulkWriter:
- We would not have to log sendTimes for each batch
- We would not have to manually calculate the number of writes made in the past second each time we want to determine whether a batch can be sent.
- Simplifies calculation for the required delay before the next batch can be sent.

Implementation notes:
- I used a `currentTime` parameter to make testing the limiter easier, since it's based on Timestamps. I was hoping to get some feedback/suggestions on how to test future Timestamps without hurting readability of the class too much.
- It is possible to factor in the next multiplier in calculating the required delay (ex: a request that would currently exceed the maximum capacity would be possible after the next capacity increase), but I don't think it's worth adding.

### DIFF
--- a/dev/src/rate-limiter.ts
+++ b/dev/src/rate-limiter.ts
@@ -36,44 +36,39 @@ export class RateLimiter {
   availableTokens: number;
 
   // When the token bucket was last refilled.
-  lastRefillTime = Timestamp.now();
+  lastRefillTimeMillis: number;
 
   /**
    * @param initialCapacity Initial maximum number of operations per second.
    * @param multiplier Rate by which to increase the capacity.
    * @param multiplierMillis How often the capacity should increase in
    * milliseconds.
-   * @param startTime Used for testing the limiter.
+   * @param startTimeMillis The starting time in epoch milliseconds that the rate limit is based on. Used
+   * for testing the limiter.
    */
   constructor(
     private readonly initialCapacity: number,
     private readonly multiplier: number,
     private readonly multiplierMillis: number,
-    private readonly startTime = Timestamp.now()
+    private readonly startTimeMillis = Date.now()
   ) {
     this.availableTokens = initialCapacity;
-    this.lastRefillTime = startTime;
+    this.lastRefillTimeMillis = startTimeMillis;
   }
 
   /**
-   * Tries to make the number of operations. Returns true if the request succeeded
-   * and false otherwise.
+   * Tries to make the number of operations. Returns true if the request
+   * succeeded and false otherwise.
    *
-   * @param currentTime Used for testing the limiter.
+   * @param currentTimeMillis The date used to calculate the number of available
+   * tokens. Used for testing the limiter.
    * @private
    */
-  tryMakeRequest(numOperations: number) {
-    return this._tryMakeRequest(numOperations, Timestamp.now());
-  }
-
-  /**
-   * Used in testing for different timestamps.
-   *
-   * @private
-   */
-  // Visible for testing.
-  _tryMakeRequest(numOperations: number, currentTime: Timestamp): boolean {
-    this.refillTokens(currentTime);
+  tryMakeRequest(
+    numOperations: number,
+    currentTimeMillis = Date.now()
+  ): boolean {
+    this.refillTokens(currentTimeMillis);
     if (numOperations <= this.availableTokens) {
       this.availableTokens -= numOperations;
       return true;
@@ -87,27 +82,20 @@ export class RateLimiter {
    * capacity. Returns -1 if the request is not possible with the current
    * capacity.
    *
+   * @param currentTimeMillis The date used to calculate the number of available
+   * tokens. Used for testing the limiter.
    * @private
    */
-  getNextRequestDelayMs(numOperations: number) {
-    return this._getNextRequestDelayMs(numOperations, Timestamp.now());
-  }
 
-  /**
-   * Used in testing for different timestamps.
-   *
-   * @private
-   */
-  // Visible for testing.
-  _getNextRequestDelayMs(
+  getNextRequestDelayMs(
     numOperations: number,
-    currentTime: Timestamp
+    currentTimeMillis = Date.now()
   ): number {
     if (numOperations < this.availableTokens) {
       return 0;
     }
 
-    const capacity = this.calculateCapacity(currentTime, this.startTime);
+    const capacity = this.calculateCapacity(currentTimeMillis);
     if (capacity < numOperations) {
       return -1;
     }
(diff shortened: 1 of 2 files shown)


dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,161 @@
...
+ * RateLimiter determines doing so stays under the provided rate limits. It can
+ * also determine how much time is required before a request can be made.
+ *
+ * RateLimiter can also implement a gradually increasing rate limit. This is
+ * used to enforce the 500/50/5 rule
+ * (https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic).
+ *
+ * @private
+ */
+export class RateLimiter {
```
schmidt-sebastian, line 34: Do we have to add some code to make tokens expire even if they are not used?  I would suspect that if you issue 500 writes per second, then sit idle for a second, you can only issue 500 writes and not 1000.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  /**
+   * Calculates the maximum capacity based on the two provided timestamps.
+   *
+   * @private
+   */
+  // Visible for testing.
+  calculateCapacity(currentTime: Timestamp, startTime: Timestamp): number {
```
schmidt-sebastian, line 147: Is it possible to drop the `startTime` argument here? It makes this code easier to reason about. In the tests, you would simply have to create a new instance.

dev/test/rate-limiter.ts
```diff
@@ -0,0 +1,95 @@
...
+import {expect} from 'chai';
+
+import {Timestamp} from '../src';
+import {RateLimiter} from '../src/rate-limiter';
+
+describe('RateLimiter', () => {
+  let limiter: RateLimiter;
+
+  beforeEach(() => {
+    limiter = new RateLimiter(500, 1.5, 5 * 60 * 1000, new Timestamp(0, 0));
```
schmidt-sebastian, line 24: Please add an inline comment for all non-descriptive arguments.

dev/test/rate-limiter.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Once tokens have been used, further requests should fail.
+    expect(limiter._tryMakeRequest(1, new Timestamp(0, 0))).to.be.false;
+
+    // Tokens will only refill up to max capacity.
+    expect(limiter._tryMakeRequest(501, new Timestamp(1, 0))).to.be.false;
+    expect(limiter._tryMakeRequest(500, new Timestamp(1, 0))).to.be.true;
+
+    // Tokens will refill incrementally based on the number of ms elapsed.
+    expect(limiter._tryMakeRequest(251, new Timestamp(1, 500 * 1e6))).to.be
+      .false;
```
schmidt-sebastian, line 40: This might make more sense as:

```
  expect(limiter._tryMakeRequest(250, new Timestamp(1, 499 * 1e6))).to.be
      .false;
```

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // When the token bucket was last refilled.
+  lastRefillTime = Timestamp.now();
+
+  /**
+   * @param initialCapacity Initial maximum number of operations per second.
+   * @param multiplier Rate by which to increase the capacity.
+   * @param multiplierMillis How often the capacity should increase in
+   * milliseconds.
+   * @param startTime Used for testing the limiter.
```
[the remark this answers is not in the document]
thebrianchen: done.

dev/test/rate-limiter.ts
thebrianchen: done.

dev/src/rate-limiter.ts
thebrianchen: Although we are not making tokens expire, we are already limiting the maximum number of tokens that can be accrued. When we refill the tokens, we call `Math.min()` on the capacity and the number of added tokens to prevent the scenario you mentioned above from happening.

I also added another test case to cover this as well.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * Refills the number of available tokens based on how much time has elapsed
+   * since the last time the tokens were refilled.
+   *
+   * @private
+   */
+  private refillTokens(currentTime = Timestamp.now()): void {
+    if (currentTime.toMillis() > this.lastRefillTime.toMillis()) {
```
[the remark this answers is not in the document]
thebrianchen: Switched to using `Date` and `getTime()`.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Refills the number of available tokens based on how much time has elapsed
+   * since the last time the tokens were refilled.
+   *
+   * @param currentTimeMillis The date used to calculate the number of available
+   * tokens. Used for testing the limiter.
+   * @private
+   */
+  private refillTokens(currentTimeMillis = Date.now()): void {
+    if (currentTimeMillis > this.lastRefillTimeMillis) {
```
schmidt-sebastian, line 116: We might want to make the behavior of this class more predictable and reject timestamps that go back in time.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // When the token bucket was last refilled.
+  lastRefillTimeMillis: number;
+
+  /**
+   * @param initialCapacity Initial maximum number of operations per second.
+   * @param multiplier Rate by which to increase the capacity.
+   * @param multiplierMillis How often the capacity should increase in
+   * milliseconds.
+   * @param startTimeMillis The starting time in epoch milliseconds that the rate limit is based on. Used
```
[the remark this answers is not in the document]
thebrianchen: done.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,150 @@
...
+   * Returns the number of ms needed to make a request with the provided number
+   * of operations. Returns 0 if the request can be made with the existing
+   * capacity. Returns -1 if the request is not possible with the current
+   * capacity.
+   *
+   * @param currentTimeMillis The date used to calculate the number of available
+   * tokens. Used for testing the limiter.
+   * @private
+   */
+
```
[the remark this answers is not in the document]
thebrianchen: done.

dev/src/rate-limiter.ts
thebrianchen: Added behavior to reject timestamps that are before the last refill time. Also added test for this.

dev/src/rate-limiter.ts
```diff
@@ -0,0 +1,161 @@
...
+    return this._tryMakeRequest(numOperations, Timestamp.now());
+  }
+
+  /**
+   * Used in testing for different timestamps.
+   *
+   * @private
+   */
+  // Visible for testing.
+  _tryMakeRequest(numOperations: number, currentTime: Timestamp): boolean {
```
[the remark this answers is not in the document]
thebrianchen: Renamed to `requestTimeMillis`.

PR #1032, opened 2020-04-20
feat: add bundle.proto and generated js code

wu-hui:
Note this is committing to a remote branch, not `master`.

Also, there are other changes picked up by `update.sh`, they are in a separate commit: 049ea619e2b3cc405b44429aca8c2a66de8ea341


wu-hui:
Done, thanks!

dev/protos/google/firestore/proto/bundle.proto
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 Google LLC.
```
schmidt-sebastian, line 1: We need to find a more descriptive (or at least less repetitive) path for this file :) What about simply `dev/protos/firestore/bundle.proto`? If you want to keep this under `google` (which I am not all that excited about), then I would make it `dev/protos/google/firestore/local/bundle.proto`.

dev/protos/update.sh
```diff
@@ -38,8 +38,8 @@ PBTS="$(npm bin)/pbts"
 pushd "$WORK_DIR"
 
 # Clone necessary git repos.
-git clone https://github.com/googleapis/googleapis.git
-git clone https://github.com/google/protobuf.git
+git clone --depth 1 https://github.com/googleapis/googleapis.git
+git clone --depth 1 https://github.com/google/protobuf.git
```
schmidt-sebastian, line 42: Do these changes make this script fast? That would so exciting :)

schmidt-sebastian, line 42: Do these changes make this script fast? That would so exciting :)

schmidt-sebastian, line 42: Do these changes make this script fast? That would so exciting :)

dev/protos/update.sh
```diff
@@ -89,6 +89,7 @@ PBJS_ARGS=( --proto_path=. \
       
 "${PBJS}" "${PBJS_ARGS[@]}" -o firestore_v1_proto_api.js \
   "${PROTOS_DIR}/google/firestore/v1/*.proto" \
+  "${PROTOS_DIR}/google/firestore/proto/*.proto" \
```
schmidt-sebastian, line 92: Is it possible to export this to a separate JS file as well? I have never tried this option, but I wonder if `--sparse` will allow you to reference external types (https://www.npmjs.com/package/protobufjs#pbjs-for-javascript) and not include the `v1` types twice.

dev/protos/google/firestore/proto/bundle.proto
wu-hui: Done.

dev/protos/update.sh
wu-hui: Yep, way faster.

dev/protos/update.sh
wu-hui: I tried `--sparse` and it simply throws error.

dev/protos/firestore/bundle.proto
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+//
+
+// This file defines the format of Firestore bundle file/stream. It is not a part of the
+// Firestore API, only a specification used by Server and Client SDK to write and read
+// bundles.
+
+syntax = "proto3";
+
+package firestore.proto;
```
schmidt-sebastian, line 22: This is now just `firestore` if my understanding is correct.

wu-hui: Done.

PR #1033, opened 2020-04-20
feat: add 555 rule support

thebrianchen:
Adding throttling. 

I tried to write the tests as cleanly as I could without overriding too much. Since `RateLimiter` already has tests, I only tested that the timeout handler in BulkWriter was being called. The test will fail with an assertion if throttling is not in place. However, I fear that this test can be brittle if another part of the SDK calls a timeout. Any ideas/tips on how to improve this test? Thanks!

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -16,9 +16,17 @@
 import * as assert from 'assert';
 
 import {FieldPath, Firestore} from '.';
+import {delayExecution} from './backoff';
+import {RateLimiter} from './rate-limiter';
 import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
-import {DocumentData, Precondition, SetOptions, UpdateData} from './types';
+import {
+  BulkWriterOptions,
+  DocumentData,
+  Precondition,
+  SetOptions,
+  UpdateData,
+} from './types';
 import {Deferred} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
 
@@ -26,6 +34,29 @@ import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
  * The maximum number of writes that can be in a single batch.
  */
 const MAX_BATCH_SIZE = 500;
+
+/**
+ * The starting maximum number of operations per second as allowed by the
+ * 500/50/5 rule.
+ *
+ * https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic.
+ */
+const STARTING_MAXIMUM_OPS_PER_SECOND = 500;
+
+/**
+ * The rate by which to increase the capacity as specified by the 500/50/5 rule.
+ *
+ * https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic.
+ */
+const RATE_LIMITER_MULTIPLIER = 1.5;
+
+/**
+ * How often the operations per second capacity should increase in milliseconds
+ * as specified by the 500/50/5 rule.
+ *
+ * https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic.
+ */
+const RATE_LIMITER_MULTIPLIER_MILLIS = 5 * 60 * 1000;
 
 /**
  * Used to represent the state of batch.
@@ -226,7 +257,23 @@ export class BulkWriter {
    */
   private closed = false;
 
-  constructor(private readonly firestore: Firestore) {}
+  /**
+   * Rate limiter used to throttle requests as per the 500/50/5 rule.
+   */
+  private rateLimiter = new RateLimiter(
+    STARTING_MAXIMUM_OPS_PER_SECOND,
+    RATE_LIMITER_MULTIPLIER,
+    RATE_LIMITER_MULTIPLIER_MILLIS
+  );
+
+  private readonly options: BulkWriterOptions = {};
+
+  constructor(
+    private readonly firestore: Firestore,
+    options?: BulkWriterOptions
+  ) {
+    this.options = {...options};
+  }
 
   /**
    * Create a document with the provided data. This single operation will fail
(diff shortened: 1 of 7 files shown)


dev/src/types.ts
```diff
@@ -321,6 +321,15 @@ export interface ValidationOptions {
   allowTransforms: boolean;
 }
 
+/**
+ * An options object that can be used to toggle the default rate limit
+ * throttling implemented by BulkWriter.
+ */
+export interface BulkWriterOptions {
```
schmidt-sebastian, line 328: This needs to be added to the public d.ts file as well.

dev/src/bulk-writer.ts
```diff
@@ -226,7 +257,23 @@ export class BulkWriter {
...
+    STARTING_MAXIMUM_OPS_PER_SECOND,
+    RATE_LIMITER_MULTIPLIER,
+    RATE_LIMITER_MULTIPLIER_MILLIS
+  );
+
+  private readonly options: BulkWriterOptions = {};
+
+  constructor(
+    private readonly firestore: Firestore,
+    options?: BulkWriterOptions
```
schmidt-sebastian, line 273: This will likely be an internal constructor, won't it? If so, I prefer if you just spelled out the option here:

```
interface Firestore() {
  bulkWriter(options?: BulkWriterOptions): BulkWriter {
    return new BulkWriter(this, !options?.disableThrottling);
}

class Firestore() {
  // Note that this is no longer the public API, so we can use Denver's suggestion
  constructor(firestore, readonly enableThrottling) {} 
}
```


If you want to go even fancier you could also pass in a no-throttling implementation of RateLimiter. This could be as simple as `new RateLimiter(Math.Infinity)` if I read this correctly.

dev/src/rate-limiter.ts
```diff
@@ -90,6 +90,7 @@ export class RateLimiter {
     numOperations: number,
     requestTimeMillis = Date.now()
   ): number {
+    this.refillTokens(requestTimeMillis);
```
schmidt-sebastian, line 93: Hm. This method read to me as "let me check when I can make the next request, but there won't be any side effects". Can you refactor this to not have side effects?

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
     return bulkWriter.close();
   });
 
+  describe.only('500/50/5 support', () => {
```
schmidt-sebastian, line 509: s/.only//

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
...
+    afterEach(() => setTimeoutHandler(setTimeout));
+
+    it('does not send batches if doing so exceeds the rate limit', done => {
+      // The test is considered a success if BulkWriter tries to send the second
+      // batch again after a timeout.
+      setTimeoutHandler((_, timeout) => {
```
schmidt-sebastian, line 515: I have some suggestions for this below.

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
...
+            response: mergeResponses(responses1),
+          },
+          {
+            request: createRequest(requests2),
+            response: mergeResponses(responses2),
+          },
+        ],
+        /* enforceSingleConcurrentRequest= */ true
+      ).then(async bulkWriter => {
+        for (let i = 0; i < 500; i++) {
```
schmidt-sebastian, line 543:
```suggestion
        setTimeoutHandler(() => done(new Error()));
        for (let i = 0; i < 500; i++) {
```

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
...
+          bulkWriter
+            .set(firestore.doc('collectionId/doc' + i), {foo: 'bar'})
+            .then(incrementOpCount);
+        }
+        const flush1 = bulkWriter.flush();
+
+        // Sending this next batch would go over the 500/50/5 capacity, so
+        // check that BulkWriter doesn't send this batch until the first batch
+        // is resolved.
+        for (let i = 500; i < 505; i++) {
```
schmidt-sebastian, line 553:
```suggestion
        setTimeoutHandler(t => (expect(t)... ; done());
        for (let i = 500; i < 505; i++) {
```

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
...
+          {
+            request: createRequest(requests1),
+            response: mergeResponses(responses1),
+          },
+          {
+            request: createRequest(requests2),
+            response: mergeResponses(responses2),
+          },
+        ],
+        /* enforceSingleConcurrentRequest= */ true
```
schmidt-sebastian, line 541: IMHO, you don't need to set this here, at which point you can also drop the call to `activeRequestDeferred`.

dev/test/bulk-writer.ts
```diff
@@ -505,6 +506,63 @@ describe('BulkWriter', () => {
...
+        // is resolved.
+        for (let i = 500; i < 505; i++) {
+          bulkWriter
+            .set(firestore.doc('collectionId/doc' + i), {foo: 'bar'})
+            .then(incrementOpCount);
+        }
+        const flush2 = bulkWriter.flush();
+        activeRequestDeferred.resolve();
+        await flush1;
+        await flush2;
```
schmidt-sebastian, line 561: Since you are using `done()`, these `awaits` are not needed, are they?

dev/src/bulk-writer.ts
thebrianchen: Wow that is clean! Changed to take in `enableThrottling` and use RateLimiter constructor to enable/disable throttling.

dev/src/rate-limiter.ts
thebrianchen: To give an accurate estimate of how many ms are required to make the next request, we have to refill the tokens when this method is called. Otherwise, `getNextRequestDelayMs()` can return an incorrect value.
 
Without `refillTokens()` in `getNextRequestDelayMs()`:
```
tryMakeRequest(500, time(0 seconds)) // success
getNextRequestDelay(500, time(1 second)) // there are no availableTokens without refill, incorrectly returns 1000ms.
```

Since the token refill logic takes into account the last refill time, calling `refillTokens()` again as part of `tryMakeRequest()` won't have any unintended side effects.

dev/src/types.ts
thebrianchen: added to firestore.d.ts, and also added option to pass in options to index.ts.

dev/test/bulk-writer.ts
thebrianchen: done.

dev/test/bulk-writer.ts
thebrianchen: done.

dev/test/bulk-writer.ts
thebrianchen: done.

dev/test/bulk-writer.ts
thebrianchen: Initially I added them to ensure the test failed if we didn't throttle, but with the timeout handlers + assert, this isn't necessary anymore.

dev/test/bulk-writer.ts
thebrianchen: thanks for the rec -- it really simplified the test.

dev/src/index.ts
```diff
@@ -644,8 +646,14 @@ export class Firestore {
 
   /**
    * Creates a [BulkWriter]{@link BulkWriter}, used for performing
-   * multiple writes in parallel.
+   * multiple writes in parallel. Automatically throttles writes as specified
```
schmidt-sebastian, line 649: Nit: Maybe use `gradually ramps ups writes`.

dev/src/rate-limiter.ts
schmidt-sebastian: That's fair. It seems like this could potentially used by simplified to not store the request time at all and to always refill tokens by default.

dev/src/types.ts
```diff
@@ -321,6 +321,15 @@ export interface ValidationOptions {
...
+  /** Whether to disable throttling as specified by the 500/50/5 rule. */
```
schmidt-sebastian, line 329: I would drop the mention of 500/50/5 here. Details are not relevant here since we explicitly don't want to throttle.

types/firestore.d.ts
```diff
@@ -545,6 +545,15 @@ declare namespace FirebaseFirestore {
     close(): Promise<void>;
   }
 
+  /**
+   * An options object that can be used to toggle the default rate limit
```
schmidt-sebastian, line 549: Maybe: `An options object that can be used to disable request throttling in BulkWriter.`

dev/src/index.ts
thebrianchen: done.

dev/src/types.ts
thebrianchen: done.

types/firestore.d.ts
thebrianchen: much better, done.

PR #1034, opened 2020-04-21
feat: Add support to convert a Query to a BundledQuery.

wu-hui:
Note the tricky thing here is: we introduced limitToLast, which means we need to decide if the `structuredQuery` saved in the bundle should be translated or not.

This implementation saves it without translation, which means the client SDK will use the `structureQuery` and `limitType` saved here to construct a new client `Query`, and leaves the translation to the existing logic on the client side.


dev/src/reference.ts
```diff
@@ -1977,26 +1950,89 @@ export class Query<T = DocumentData> {
...
-          o.toProto()
-        );
+    }
+
+    const reqOpts: api.IRunQueryRequest = {
+      parent: parentPath.formattedName,
+      structuredQuery,
+    };
+
+    reqOpts.transaction = transactionId;
```
schmidt-sebastian, line 1960: Nit: Merge this into the assignment above.

dev/src/reference.ts
```diff
@@ -1977,26 +1950,89 @@ export class Query<T = DocumentData> {
...
+
+    return reqOpts;
```
schmidt-sebastian, line 1962: Possible inline?

dev/src/reference.ts
```diff
@@ -1977,26 +1950,89 @@ export class Query<T = DocumentData> {
...
+  }
+
+  /**
+   * Internal method for serializing a query to its RunQuery proto
+   * representation with an optional transaction id.
```
schmidt-sebastian, line 1967: This documentation seems not applicable.

dev/src/reference.ts
```diff
@@ -1977,26 +1950,89 @@ export class Query<T = DocumentData> {
...
+
+  private toStructuredQuery(): api.IStructuredQuery {
+    const structuredQuery: api.IStructuredQuery = {
+      from: [
+        {
+          collectionId: this._queryOptions.collectionId,
+        },
+      ],
+    };
+    if (this._queryOptions.allDescendants) {
```
schmidt-sebastian, line 2001: Nit: Empty line above

dev/src/reference.ts
```diff
@@ -1977,26 +1950,89 @@ export class Query<T = DocumentData> {
...
+        },
+      };
     }
 
+    if (this._queryOptions.hasFieldOrders()) {
+      structuredQuery.orderBy = this._queryOptions.fieldOrders.map(o =>
+        o.toProto()
+      );
+    }
+    structuredQuery.startAt = this.toCursor(this._queryOptions.startAt);
```
schmidt-sebastian, line 2025: Nit: Empty line above

dev/src/reference.ts
wu-hui: Done.

dev/src/reference.ts
wu-hui: Done.

dev/src/reference.ts
wu-hui: Fixed.

dev/src/reference.ts
wu-hui: Done.

dev/src/reference.ts
wu-hui: Done.

PR #1035, opened 2020-04-22
Update conformance tests to use update_transform

thebrianchen:
Edited the json files and added the skipped test back into the conformance tests in `runner.ts`.

Tests are passing according to CI. Docs don't compile in this PR but that's being fixed in #1036.

### DIFF
--- a/dev/conformance/conformance-tests/set-mergeall-empty.json
+++ b/dev/conformance/conformance-tests/set-mergeall-empty.json
@@ -16,6 +16,9 @@
               "update": {
                 "name": "projects/projectID/databases/(default)/documents/C/d",
                 "fields": {}
+              },
+              "updateMask": {
+                "fieldPaths": []
               }
             }
           ]
--- a/dev/conformance/conformance-tests/set-mergeall-empty.json
+++ b/dev/conformance/conformance-tests/set-mergeall-empty.json
@@ -16,9 +16,6 @@
               "update": {
                 "name": "projects/projectID/databases/(default)/documents/C/d",
                 "fields": {}
-              },
-              "updateMask": {
-                "fieldPaths": []
               }
             }
           ]
--- a/dev/conformance/conformance-tests/set-st-alone-mergeall.json
+++ b/dev/conformance/conformance-tests/set-st-alone-mergeall.json
@@ -18,9 +18,7 @@
                 "name": "projects/projectID/databases/(default)/documents/C/d"
               },
               "updateMask": {
-                "fieldPaths": [
-                  "a"
-                ]
+                "fieldPaths": []
               },
               "updateTransforms": [
                 {
--- a/dev/conformance/conformance-tests/set-st-alone-mergeall.json
+++ b/dev/conformance/conformance-tests/set-st-alone-mergeall.json
@@ -20,7 +20,7 @@
               "updateTransforms": [
                 {
                   "fieldPath": "a",
-                  "setToServerValue": 1
+                  "setToServerValue": "REQUEST_TIME"
                 }
               ]
             }
--- a/dev/conformance/conformance-tests/set-st-alone-mergeall.json
+++ b/dev/conformance/conformance-tests/set-st-alone-mergeall.json
@@ -13,15 +13,16 @@
           "database": "projects/projectID/databases/(default)",
           "writes": [
             {
-              "transform": {
-                "document": "projects/projectID/databases/(default)/documents/C/d",
-                "fieldTransforms": [
-                  {
-                    "fieldPath": "a",
-                    "setToServerValue": "REQUEST_TIME"
-                  }
-                ]
-              }
+              "update": {
+                "fields": {},
+                "name": "projects/projectID/databases/(default)/documents/C/d"
+              },
+              "updateTransforms": [
+                {
+                  "fieldPath": "a",
+                  "setToServerValue": 1
+                }
+              ]
             }
           ]
         }
--- a/dev/conformance/conformance-tests/set-st-alone-mergeall.json
+++ b/dev/conformance/conformance-tests/set-st-alone-mergeall.json
@@ -17,6 +17,11 @@
                 "fields": {},
                 "name": "projects/projectID/databases/(default)/documents/C/d"
               },
+              "updateMask": {
+                "fieldPaths": [
+                  "a"
+                ]
+              },
               "updateTransforms": [
                 {
                   "fieldPath": "a",
--- a/dev/conformance/conformance-tests/set-st-merge-nowrite.json
+++ b/dev/conformance/conformance-tests/set-st-merge-nowrite.json
@@ -19,15 +19,16 @@
           "database": "projects/projectID/databases/(default)",
           "writes": [
             {
-              "transform": {
-                "document": "projects/projectID/databases/(default)/documents/C/d",
-                "fieldTransforms": [
-                  {
-                    "fieldPath": "b",
-                    "setToServerValue": "REQUEST_TIME"
-                  }
-                ]
-              }
+              "update": {
+                "fields": {},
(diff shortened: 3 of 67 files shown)


dev/conformance/conformance-tests/set-mergeall-empty.json
```diff
@@ -16,9 +16,6 @@
               "update": {
                 "name": "projects/projectID/databases/(default)/documents/C/d",
                 "fields": {}
-              },
```
schmidt-sebastian, line 19: Do you know why this had to change? It feels like what you have is correct, but I am a bit concerned that it changed  in your branch. 

If this change is correct, we need to update the description.

dev/conformance/conformance-tests/update-arrayunion-alone.json
```diff
@@ -10,30 +10,46 @@
...
+              },
+              "updateMask": {
+                "fieldPaths": []
+              },
+              "updateTransforms": [
+                {
+                  "appendMissingElements": {
+                    "values": [
+                      {
+                        "integerValue": {
```
schmidt-sebastian, line 28: This should use the JSON representation rather than the Long representation. Take a look at the original file to see what it should look like.

dev/conformance/conformance-tests/update-paths-arrayremove-alone.json
```diff
@@ -19,30 +19,46 @@
...
+              "updateMask": {
+                "fieldPaths": []
+              },
+              "updateTransforms": [
+                {
+                  "fieldPath": "a",
+                  "removeAllFromArray": {
+                    "values": [
+                      {
+                        "integerValue": {
```
schmidt-sebastian, line 38: Same comment as above regarding integer versus Long representation.

dev/conformance/conformance-tests/update-paths-arrayunion-alone.json
```diff
@@ -19,30 +19,46 @@
...
+              },
+              "updateMask": {
+                "fieldPaths": []
+              },
+              "updateTransforms": [
+                {
+                  "appendMissingElements": {
+                    "values": [
+                      {
+                        "integerValue": {
```
schmidt-sebastian, line 37: Same.

dev/conformance/conformance-tests/update-paths-st-with-empty-map.json
```diff
@@ -41,25 +41,14 @@
                 ]
               },
               "updateTransforms": [
-              {
-               "fieldPath": "a.c",
-                "setToServerValue": 1
-              }
+                {
```
schmidt-sebastian, line 44: Nit: Remove spaces.

dev/conformance/conformance-tests/update-paths-st-with-empty-map.json
```diff
@@ -41,25 +41,14 @@
                 ]
               },
               "updateTransforms": [
-              {
```
schmidt-sebastian, line 44: Why is there an "updateTransforms"? Was this edited earlier?

dev/conformance/conformance-tests/update-paths-st-with-empty-map.json
```diff
@@ -41,25 +41,14 @@
...
-               "fieldPath": "a.c",
-                "setToServerValue": 1
-              }
+                {
+                  "fieldPath": "a.c",
+                  "setToServerValue": 1
```
schmidt-sebastian, line 46: All of "setToServerValue: 1" should be "setToServerValue: "REQUEST_TIME""

dev/conformance/conformance-tests/set-mergeall-empty.json
thebrianchen: You can see the diff responsible for this [here](https://github.com/googleapis/nodejs-firestore/compare/master...bc/bulk-master?expand=1#diff-9251327f0da3d89ce43887319402b935R347).

In the original logic: if the `merge` option is specified and the `transform` field is empty, add the `updateMask` field. This meant that we could have an empty `updateMask` field as long as the `transform` field was empty.

With the changes to `updateTransform`, I changed the code to only add the updateMask if it wasn't empty. To preserve the original behavior, we can change the line of code I mentioned above to:
```
if (hasMerge && (!documentMask.isEmpty || transform.isEmpty))
```

Does including an empty `updateMask` change behavior? If not, I think it'd be better to leave the field out. The description is still valid, since we are still checking that a merge with empty data results in a valid request being created.

schmidt-sebastian: Actually, come to think of this, this is now broken. Without the update mask, set({}, {merge:true}) is now the same as set({}, {merge:true}) and replaces all data on the backend. Instead, it should preserve the document on the backend and be a no op if the document already exists.

Can you fix this? I will look at the rest of this diff once this is fixed.

dev/test/document.ts
```diff
@@ -1150,6 +1150,7 @@ describe('set document', () => {
           request,
           set({
             document: document('documentId'),
+            mask: updateMask(),
```
thebrianchen, line 1153: turns out we had the test all this time, but I deleted it from the unit tests back when I first added the proto :(

dev/conformance/conformance-tests/set-st-alone-mergeall.json
```diff
@@ -13,15 +13,16 @@
...
-                    "fieldPath": "a",
-                    "setToServerValue": "REQUEST_TIME"
-                  }
-                ]
-              }
+              "update": {
+                "fields": {},
+                "name": "projects/projectID/databases/(default)/documents/C/d"
+              },
+              "updateTransforms": [
```
schmidt-sebastian, line 20: How much work would it be to preserve the initial Proto for the case where we only have transforms and no updates? That produces a cleaner and more minimalist Proto.

dev/conformance/conformance-tests/set-st-alone-mergeall.json
```diff
@@ -13,15 +13,16 @@
...
-              "transform": {
-                "document": "projects/projectID/databases/(default)/documents/C/d",
-                "fieldTransforms": [
-                  {
-                    "fieldPath": "a",
-                    "setToServerValue": "REQUEST_TIME"
-                  }
-                ]
-              }
+              "update": {
```
schmidt-sebastian, line 16: I think this is not correct and needs to provide an update mask. Otherwise, it overwrites all fields on the backend.

dev/conformance/conformance-tests/set-st-merge-nowrite.json
```diff
@@ -19,15 +19,16 @@
...
-                  {
-                    "fieldPath": "b",
-                    "setToServerValue": "REQUEST_TIME"
-                  }
-                ]
-              }
+              "update": {
+                "fields": {},
+                "name": "projects/projectID/databases/(default)/documents/C/d"
+              },
```
schmidt-sebastian, line 25: This is missing an updateMask - but I would prefer if we could keep the original Protobuf format here.

dev/conformance/conformance-tests/set-st-alone-mergeall.json
```diff
@@ -17,6 +17,11 @@
                 "fields": {},
                 "name": "projects/projectID/databases/(default)/documents/C/d"
               },
+              "updateMask": {
```
schmidt-sebastian, line 20: This needs to be an empty mask as you are otherwise deleting field "a" before applying the transform, which invalidates increment() and arrayUnion().

thebrianchen: Resolved. I think it's in a happy state now.

thebrianchen: Resolved. Also added a system-test for this.

PR #1036, opened 2020-04-22
Make BulkWriter private until protos are released

thebrianchen:
- Remove BulkWriter mentions from docs.
- Remove export for `BulkWriter and `BulkWriterOptions` in `firestore.d.ts`.
- Prefix `bulkWriter()` with `_` in `index.ts`


types/firestore.d.ts
```diff
@@ -416,7 +416,7 @@ declare namespace FirebaseFirestore {
    *
    * @class
    */
-  export class BulkWriter {
+  class BulkWriter {
```
schmidt-sebastian, line 419: All types in a `d.ts` file as exported by default. Unfortunately, we can only hide them by removing the entries completely.

dev/src/index.ts
```diff
@@ -677,7 +678,7 @@ export class Firestore {
    * });
    * bulkWriter.close();
    */
-  bulkWriter(options?: BulkWriterOptions): BulkWriter {
+  _bulkWriter(options?: BulkWriterOptions): BulkWriter {
```
schmidt-sebastian, line 681: Make sure to also remove the `export { BulkWriter }` line.

types/firestore.d.ts
thebrianchen: done.

dev/src/index.ts
thebrianchen: done.

PR #1041, opened 2020-04-23
feat: Implement BundleBuilder

wu-hui:
Implements `BundleBuilder` and added system tests around bundles.


dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  stream(): NodeJS.ReadableStream {
+    assert(
+      this.docRefs.length +
+        this.docSnaps.length +
+        this.queries.size +
+        this.querySnaps.size >
+        0,
+      'Nothing is added to the bundle.'
+    );
```
schmidt-sebastian, line 132: If you do want throw an error here, it should be validation error rather than an assert. This is caused by the user and not by us and our SDK should only trigger assertions for problems that the SDK created.

I do however think that it would be more user friendly to just return an empty bundle, as this behavior is less surprising than an exception that might only occur under certain usage scenarios that otherwise are valid.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+    for (const snap of this.docSnaps) {
+      this.addBundledDocument(snap);
+    }
+
+    for (const [name, snap] of Array.from(this.querySnaps)) {
+      if (this.namedQueries.has(name)) {
```
schmidt-sebastian, line 138: This should be part of the `add` call.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+        throw new Error(`Query name conflict: ${name}`);
+      }
+
+      this.addNamedQuery(name, snap);
+    }
+
+    const promises: Array<Promise<void>> = [];
+    for (const doc of this.docRefs) {
+      promises.push(
```
schmidt-sebastian, line 147: It is surprising to me that the `stream()` API fetches the documents from the backend. I think we should maybe revisit the API here and change add() to only take DocumentSnapshots or QuerySnapshots (and take the opportunity to split the methods into two instead of overloading them). 

If we want to keep the ability to add DocumentReferences, then `add()` should be async and do perform the document retrieval.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+        doc.get().then(snap => {
+          this.addBundledDocument(snap);
+        })
+      );
+    }
+
+    for (const [name, query] of Array.from(this.queries)) {
+      if (this.namedQueries.has(name)) {
```
schmidt-sebastian, line 155: This should be part of the add call.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+        query.get().then(snap => {
+          const querySnap = snap as QuerySnapshot;
+          this.addNamedQuery(name, querySnap);
+        })
+      );
+    }
+
+    const readable = new Readable({
+      objectMode: false,
+      read: async size => {
```
schmidt-sebastian, line 169: Read cannot be `async` as far as I can tell. It needs to return the data that is being read.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+        await Promise.all(promises);
+        const metadata: firestore.IBundleMetadata = {
+          id: this.id,
+          createTime: this.latestReadTime.toProto().timestampValue,
+        };
+        this.pushLengthPrefixedString(
+          readable,
+          JSON.stringify({metadata} as firestore.IBundleElement)
```
schmidt-sebastian, line 177: The cast is not valid (since `metadata` is `IBundleMetadata`, but not `{metadata}`). You can just drop the cast.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  private pushLengthPrefixedString(readable: Readable, payload: string) {
+    const buffer = Buffer.from(payload, 'utf-8');
+    const lengthBuffer = Buffer.from(buffer.length.toString());
+
+    readable.push(Buffer.concat([lengthBuffer, buffer]));
+  }
+
+  stream(): NodeJS.ReadableStream {
```
schmidt-sebastian, line 124: If you rebase this on top of the Node 10 branch you can use `async iterators` which are a user friendly version of  streams. While we didn't discuss this during the design, do you want to explore what this option would look like?

https://javascript.info/async-iterators-generators

Example: https://github.com/googleapis/nodejs-firestore/blob/node10/dev/src/v1/firestore_client.ts#L1335

dev/src/index.ts
```diff
@@ -749,6 +750,18 @@ export class Firestore {
...
 
+  /**
+   * Creates a new `BundleBuilder` instance to package selected Firestore data into
+   * a bundle.
+   *
+   * @param bundleId. The id of the bundle. When loaded on clients, client SDKs use this id
+   * and the timestamp associated with the built bundle to tell if it has been loaded already.
+   * If not specified, a random string will be used.
+   */
+  bundle(name?: string): BundleBuilder {
```
schmidt-sebastian, line 761: Should this be `bundleBuilder()`?

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+import {firestore, google} from '../protos/firestore_v1_proto_api';
+
+import {DocumentSnapshot} from './document';
+import {DocumentReference, Query, QuerySnapshot} from './reference';
+import {Timestamp} from './timestamp';
+
+import api = google.firestore.v1;
+
+/**
+ * Builds a Firestore data bundle with results from given queries and specified
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+ * documents.
+ *
+ * For documents in scope for multiple queries, the latest read version will
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+ * be included in the bundle.
+ */
+export class BundleBuilder {
+  // DocumentReference and Query added to the bundle.
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+  private docRefs: DocumentReference[] = [];
+  private queries: Map<string, Query> = new Map<string, Query>();
+
+  // DocumentSnapshot and QuerySnapshot added to the bundle.
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+    QuerySnapshot
+  >();
+
+  // Resulting documents for the bundle.
+  private documents: Map<string, BundledDocument> = new Map();
+  // Named queries saved in the bundle.
+  private namedQueries: Map<string, firestore.INamedQuery> = new Map();
+
+  // The latest read time among all bundled documents and queries.
+  private latestReadTime: Timestamp = Timestamp.fromProto({
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+ * Builds a Firestore data bundle with results from given queries and specified
+ * documents.
+ *
+ * For documents in scope for multiple queries, the latest read version will
+ * be included in the bundle.
+ */
+export class BundleBuilder {
+  // DocumentReference and Query added to the bundle.
+  private docRefs: DocumentReference[] = [];
+  private queries: Map<string, Query> = new Map<string, Query>();
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+  // Named queries saved in the bundle.
+  private namedQueries: Map<string, firestore.INamedQuery> = new Map();
+
+  // The latest read time among all bundled documents and queries.
+  private latestReadTime: Timestamp = Timestamp.fromProto({
+    seconds: 0,
+    nanos: 0,
+  });
+
+  constructor(private id: string) {}
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  add(doc: DocumentReference): BundleBuilder;
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+  add(docSnap: DocumentSnapshot): BundleBuilder;
+  add(queryName: string, querySnap: QuerySnapshot): BundleBuilder;
+  add(queryName: string, query: Query): BundleBuilder;
+  add(
+    docOrName: DocumentReference | DocumentSnapshot | string,
+    query?: QuerySnapshot | Query
+  ) {
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+    for (const snap of querySnap.docs) {
+      this.addBundledDocument(snap);
+    }
+
+    if (querySnap.readTime > this.latestReadTime) {
+      this.latestReadTime = querySnap.readTime;
+    }
+  }
+
+  private pushLengthPrefixedString(readable: Readable, payload: string) {
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
wu-hui: Note this is the `read` from `ReadableOptions`, and its return type is a `void`.

However I will switch to asyncIterator anyways, so it does not matter any more.

dev/src/bundle.ts
wu-hui: I think it is valid but unnecessary, because it is casting to `IBundleElement`.

I slightly prefer to keep these though, because it can serve as a reminder that it is a `IBundleElement` that gets saved in the bundle.

dev/src/bundle.ts
wu-hui: As discussed, we will use this stream impl for now. Async iterators are nice, but bunldes are usually already serialized when user have to deal with them, so make it part of interface does not bring a lot benefit usually come in other cases.

dev/src/bundle.ts
wu-hui: Done.

dev/src/bundle.ts
wu-hui: Done.

dev/src/bundle.ts
wu-hui: Done, no more references or queries, only snapshots.

dev/src/bundle.ts
```diff
@@ -0,0 +1,225 @@
...
+    nanos: 0,
+  });
+
+  constructor(private id: string) {}
+
+  add(doc: DocumentReference): BundleBuilder;
+  add(docSnap: DocumentSnapshot): BundleBuilder;
+  add(queryName: string, querySnap: QuerySnapshot): BundleBuilder;
+  add(queryName: string, query: Query): BundleBuilder;
+  add(
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/index.ts
wu-hui: This is an outcome of the API review, the argument here is in Node `Builder` is seldom used.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+    for (const snap of querySnap.docs) {
+      this.addBundledDocument(snap);
+    }
+
+    if (querySnap.readTime > this.latestReadTime) {
+      this.latestReadTime = querySnap.readTime;
+    }
+  }
+
+  private lengthPrefixedBuffer(payload: string): Buffer {
```
schmidt-sebastian, line 116: Please add some JSDoc here, as the method is not 100% self explanatory.

This is also missing a return type.

dev/src/bundle.ts
schmidt-sebastian: Sounds good. FWIW, Async Iterators can easily consume Streams and customers can build their own Async Iterator API.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+            this.lengthPrefixedBuffer(
+              JSON.stringify({document} as firestore.IBundleElement)
+            )
+          );
+        }
+        readable.push(null);
+      },
+    });
+
+    return readable;
```
schmidt-sebastian, line 164: You are building the stream contents before returning them, which takes away some of the advantage of using a stream to begin with. Take a look at https://nodesource.com/blog/understanding-streams-in-nodejs/ (search for "async iterator").

Otherwise, I think you might need to move your `push` calls outside of `read()`.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+    const readable = new Readable({
+      objectMode: false,
+      read: () => {
+        const metadata: firestore.IBundleMetadata = {
+          id: this.bundleId,
+          createTime: this.latestReadTime.toProto().timestampValue,
+        };
+        readable.push(
+          this.lengthPrefixedBuffer(
+            JSON.stringify({metadata} as firestore.IBundleElement)
```
schmidt-sebastian, line 132: I stumbled over this again and it took me a while to figure out why your code is correct.

It might be easier to parse if `lengthPrefixedBuffer` took a `firestore.IBundleElement` and invoked JSON.stringify directly.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+          );
+        }
+        readable.push(null);
+      },
+    });
+
+    return readable;
+  }
+
+  build(): Promise<Buffer> {
```
schmidt-sebastian, line 167: This could be simplified:

```
 async build(): Promise<Buffer> {
    const stream = this.stream();
    let data = Buffer.alloc(0);
    for await (const chunk of stream) {
      data = Buffer.concat([data, chunk as Uint8Array]);
    }
    return data;
  }
```

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+      let data = Buffer.alloc(0);
+
+      stream.on('data', chunk => (data = Buffer.concat([data, chunk])));
+      stream.on('end', () => resolve(data));
+      stream.on('error', error => reject(error));
+    });
+  }
+}
+
+class BundledDocument {
```
schmidt-sebastian, line 179: Please add JSDoc.

dev/system-test/firestore.ts
```diff
@@ -2321,3 +2324,271 @@ describe('Client initialization', () => {
     });
   }
 });
+
+describe('Bundle building', () => {
+  // Helper function for testing: converts "13{..}43{..}51{..}" to [[13,43,51],[obj,obj,obj]]
+  function bundleToLengthsAndObject(bundle: string): [number[], object[]] {
```
schmidt-sebastian, line 2330: To make this easier to digest, consider returning an array with objects: `Array<{length: number, content: IBundleElement}>`

dev/system-test/firestore.ts
```diff
@@ -2321,3 +2324,271 @@ describe('Client initialization', () => {
...
+    const isNum = (c: string) => !isNaN(Number(c));
+    let jsons = '';
+    let bracketDepth = 0;
+    let insertComma = false;
+
+    // Turn "13{..}43{..}51{..}" into "{..},{..},{..}"
+    for (const c of bundle) {
```
schmidt-sebastian, line 2337: This is a surprising way to read this data. You should take advantage of the data format you build and read this in two phases:

- Extract the length by reading until you hit "{".
- Reading the next "length" bytes

dev/system-test/firestore.ts
```diff
@@ -2321,3 +2324,271 @@ describe('Client initialization', () => {
...
+      ref1.set({name: '1', sort: 1, value: 'string value'}),
+      ref2.set({name: '2', sort: 2, value: 42}),
+      ref3.set({name: '3', sort: 3, value: {nested: 'nested value'}}),
+      ref4.set({name: '4', sort: 4, value: FieldValue.serverTimestamp()}),
+    ]);
+  });
+
+  afterEach(() => verifyInstance(firestore));
+
+  it('succeeds to read length prefixed json with testing function', () => {
```
schmidt-sebastian, line 2401: Should we make this a unit test and move `bundleToLengthsAndObject` to `helpers.ts`?

dev/system-test/firestore.ts
```diff
@@ -2321,3 +2324,271 @@ describe('Client initialization', () => {
...
+      id: 'test-bundle',
+      createTime: snap.readTime.toProto().timestampValue
+    });
+
+    const namedQuery = (elements[1] as IBundleElement).namedQuery;
+    // Verify saved query.
+    expect(namedQuery).to.deep.equal({
+      name: 'query',
+      readTime: snap.readTime.toProto().timestampValue,
+      bundledQuery: extend(
```
schmidt-sebastian, line 2507: Nit: Replace all usages of `extend` with `{ ...oldData }`.

dev/system-test/firestore.ts
```diff
@@ -2321,3 +2324,271 @@ describe('Client initialization', () => {
...
+    const [lengths, objects] = bundleToLengthsAndObject(bundleString);
+    expect(lengths).to.deep.equal([47, 53, 454]);
+    expect(objects).to.deep.equal([
+      {a: 'string value'},
+      {b: 123},
+      {c: {d: 'nested value'}},
+    ]);
+  });
+
+  it('succeeds with document snapshots', async () => {
```
schmidt-sebastian, line 2413: All of these tests should be unit tests in their own file (bundle.ts). We can cherry-pick some tests and duplicate them as system tests.

We also need at least one test that consumes `stream()` directly.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+import {DocumentSnapshot} from './document';
+import {QuerySnapshot} from './reference';
+import {Timestamp} from './timestamp';
+
+import api = google.firestore.v1;
+
+/**
+ * Builds a Firestore data bundle with results from the given queries and specified
+ * documents.
+ *
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+
+import {DocumentSnapshot} from './document';
+import {QuerySnapshot} from './reference';
+import {Timestamp} from './timestamp';
+
+import api = google.firestore.v1;
+
+/**
+ * Builds a Firestore data bundle with results from the given queries and specified
+ * documents.
```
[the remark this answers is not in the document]
wu-hui: Agree.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+  // Named queries saved in the bundle, keyed by query name.
+  private namedQueries: Map<string, firestore.INamedQuery> = new Map();
+
+  // The latest read time among all bundled documents and queries.
+  private latestReadTime = new Timestamp(0, 0);
+
+  constructor(private bundleId: string) {}
+
+  add(documentSnapshot: DocumentSnapshot): BundleBuilder;
+  add(queryName: string, querySnap: QuerySnapshot): BundleBuilder;
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * Adds a Firestore document snapshot or query snapshot to the bundle.
+   * Both the documents data and the query read time will be included in the bundle.
+   *
+   * @param {DocumentSnapshot | string=} documentOrName A document snapshot to add or a name of a query.
+   * @param {Query?=} querySnapshot A query snapshot to add to the bundle, if provided.
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -0,0 +1,184 @@
...
+   * @returns {BundleBuilder} This instance.
+   *
+   * @example
+   * let bundle = firestore.bundle('data-bundle');
+   * const bundleStream =
+   *    await bundle.add(await firestore.doc('abc/123').get()) // Add a document
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/system-test/firestore.ts
schmidt-sebastian: System tests are usually easier to write, but they:

- are harder to debug
- are not deterministic
- make it impossible to test some edge cases
- cannot be run without Production credentials
- don't count against code coverage

Please add unit tests :)

dev/src/bundle.ts
```diff
@@ -19,14 +19,15 @@ import {firestore, google} from '../protos/firestore_v1_proto_api';
...
+import {validateOptional, validateString} from './validate';
 
 import api = google.firestore.v1;
 
 /**
  * Builds a Firestore data bundle with results from the given queries and specified
  * documents.
  *
- * For documents in scope for multiple queries, only the latest version will
+ * For documents that are included multiple times, only the latest version will
```
schmidt-sebastian, line 30: Independent of how we decide to implement this, should we relax this documentation to allow us to send multiple document version if the are processed "oldest to newest" first?

dev/src/bundle.ts
```diff
@@ -41,31 +42,36 @@ export class BundleBuilder {
...
    * let bundle = firestore.bundle('data-bundle');
-   * const bundleStream =
-   *    await bundle.add(await firestore.doc('abc/123').get()) // Add a document
-   *                .add('coll-query', await firestore.collection('coll').get()) // Add a named query.
-   *                .stream();
+   * const docSnapshot = await firestore.doc('abc/123').get();
+   * const querySnapshot = await firestore.collection('coll').get();
+   * const bundleStream = bundle.add(docSnapshot) // Add a document
+   *                            .add('coll-query', querySnapshot) // Add a named query.
+   *                            .stream();
```
schmidt-sebastian, line 60:
```suggestion
   *   bundle.add(docSnapshot) // Add a document
   *   bundle.add('coll-query', querySnapshot) // Add a named query.
   *   const  bundleStream.stream();
```

dev/src/bundle.ts
```diff
@@ -41,31 +42,36 @@ export class BundleBuilder {
...
    * // Stream `bundleStream` to clients.
    */
   add(
     documentOrName: DocumentSnapshot | string,
     querySnapshot?: QuerySnapshot
   ): BundleBuilder {
-    if (documentOrName instanceof DocumentSnapshot) {
+    if (
+      documentOrName instanceof DocumentSnapshot &&
+      validateOptional(querySnapshot, {optional: true})
```
schmidt-sebastian, line 69: `validateOptional` only validates that `querySnapshot` can be `undefined` not that it needs to be `undefined` when `documentOrName` is a `DocumentSnapshot`.

This validation should also be moved into the if branch.

dev/src/bundle.ts
```diff
@@ -41,31 +42,36 @@ export class BundleBuilder {
...
+    ) {
       this.addBundledDocument(documentOrName);
     } else if (querySnapshot instanceof QuerySnapshot) {
-      this.addNamedQuery(documentOrName, querySnapshot);
+      validateString('documentOrName', documentOrName);
```
schmidt-sebastian, line 73: You need to make sure that there is no way for the user to exit out of this function without either adding something to the bundle or receiving a validation error.

dev/src/bundle.ts
```diff
@@ -113,8 +118,12 @@ export class BundleBuilder {
     }
   }
 
-  private lengthPrefixedBuffer(payload: string): Buffer {
-    const buffer = Buffer.from(payload, 'utf-8');
+  // Converts a IBundleElement to a Buffer whose content is the length prefixed JSON representation
```
schmidt-sebastian, line 121: For method comments, prefer JSDoc:

```
/**
 * Converts a IBundleElement...
 * @private
 */
```

Note that you have to add `@private` to make sure this doesn't show up in our public docs. JSDoc doesn't know that this method is TypeScript private.

dev/src/bundle.ts
```diff
@@ -176,6 +169,9 @@ export class BundleBuilder {
   }
 }
 
+/**
+ * Convenient class to hold both the metadata and the actual content of a document to be bundled.
+ */
```
schmidt-sebastian, line 174: Add `@private` to hide this from the public API docs.

dev/system-test/firestore.ts
schmidt-sebastian: The `helpers.ts` is actually a unit test-only file: https://github.com/googleapis/nodejs-firestore/blob/master/dev/test/util/helpers.ts

dev/system-test/firestore.ts
schmidt-sebastian: I haven't yet convinced myself that the logic here adds more test coverage, but I am convinced that it is much harder to parse than the code that we expect users to write would be. In fact, if we don't think that our users will parse bundles like this, then we shouldn't either. We need to validate that the format we are using doesn't just match our specification, but can actually be consumed as intended by our users.

You could use this function in one (or if you want to maybe even most) unit tests, but you should at least have some coverage that uses the "straightforward" way to parse this data. FWIW, there are even packages that you can use: https://www.npmjs.com/package/length-prefixed-json-stream

dev/src/bundle.ts
schmidt-sebastian: `async` is not part of the API, only `Promise<Buffer>` is. There are some subtle differences for `async` functions but I don't think the differences apply here.

```
let i = 0;
function inc() : Promise.resolve() {
  ++i
  return Promise.resolve();
}
inc() // `i` is now 1
```
Versus:
```
let i = 0;
async function inc() : Promise.resolve() {
  ++i
  return Promise.resolve();
}
inc() // `i` is now 0, since the entire function is executed via the Promise chain
await Promise.resolve(); // `i` is now 1
```

dev/src/bundle.ts
schmidt-sebastian: There is no code in here that enforces that enforces that no documents are added after Stream is called. We should either add this or support the following use pattern:

```
const stream = bundle.stream();
bundle.add()
bundle.add()
```

Since we can support this API usage with no additional complexity, I think we should evaluate offering it. We can even directly instantiate a stream when we create a BundleBuilder and push directly to it instead of storing the documents in temporary variables. While we still need to keep the read times, we don't need to keep the contents once the user has read the data which reduces our memory consumption.

If you want rather want to add enforcement that all data has been added before invoking `stream`, I would still like to encourage you to re-arrange the code here to push data outside of the `read` call. You could also simply use a PassThrough stream: 

```
const readable = new PassThrough();
readable.push('...');
readable.push('...');
readable.push('...');
return readable;
```

dev/system-test/firestore.ts
wu-hui: Moved some to unit tests, thanks.

dev/test/bundle.ts
```diff
@@ -0,0 +1,114 @@
...
+      (elements[1] as IBundleElement).documentMetadata,
+      (elements[2] as IBundleElement).document,
+    ];
+    expect(result1).to.deep.equal([
+      {
+        documentKey: snap1.toDocumentProto().name,
+        readTime: snap1.readTime.toProto().timestampValue,
+      },
+      snap1.toDocumentProto(),
+    ]);
```
schmidt-sebastian, line 98: This seems like it should be two different `expect` calls.

dev/test/bundle.ts
```diff
@@ -0,0 +1,114 @@
...
+  });
+
```
schmidt-sebastian, line 100: Should we add a test that a second call to build() contains the first round of documents as well as the ones added after the first call?

dev/test/util/helpers.ts
```diff
@@ -353,3 +355,16 @@ export const postConverter = {
     return new Post(data.title, data.author);
   },
 };
+
+export async function bundleToElementArray(
+  bundle: Buffer
+): Promise<Array<firestore.IBundleElement>> {
+  const result: Array<firestore.IBundleElement> = [];
+  const readable = new PassThrough();
```
schmidt-sebastian, line 363:
```suggestion
  const readable = Readable.from(bundle);
```

dev/src/external-modules.d.ts
```diff
@@ -16,3 +16,4 @@
 
 // TODO(mrschmidt): Come up with actual definitions for these modules.
 declare module 'functional-red-black-tree';
+declare module 'length-prefixed-json-stream';
```
schmidt-sebastian, line 19: I see what you did there, adding work to my TODO :)

dev/system-test/firestore.ts
wu-hui: There are other fields: `select` and `offset`, besides cursors. I'd prefer to fixing this in a separate PR.

dev/src/bundle.ts
```diff
@@ -0,0 +1,213 @@
...
+      if (!this.documents.has(snap.id) || savedReadTime < snap.readTime) {
+        this.documents.set(snap.id, {
+          document: docProto,
+          metadata: {
+            documentKey: docProto.name,
+            readTime: snap.readTime.toProto().timestampValue,
+          },
+        });
+      }
+      if (snap.readTime > this.latestReadTime) {
```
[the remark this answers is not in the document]
wu-hui: Note `NoDocument` does not exist in Node SDK, it's a local proto. 

For now, I simply store the a document proto with only document name, with `fields`, `updateTime` and `createTime` missing.

dev/src/external-modules.d.ts
wu-hui: Damn, I should have picked someone else.

dev/src/document.ts
```diff
@@ -498,12 +498,17 @@ export class DocumentSnapshot<T = DocumentData> {
    * @private
    */
   toDocumentProto(): api.IDocument {
-    return {
-      createTime: this.createTime,
-      updateTime: this.updateTime,
-      name: this._ref.formattedName,
-      fields: this._fieldsProto!,
-    };
+    const result: api.IDocument = {name: this._ref.formattedName};
```
schmidt-sebastian, line 501: I think you should do this conversion in the caller. The backend has a specified way to return "NoDocuments": https://source.corp.google.com/piper///depot/google3/google/firestore/v1beta1/firestore.proto;l=496

In the calling code, you could do something like 

```
if (doc.exists) ? { document: doc.toProto()} : {name: this._ref.formattedName}
```

In the Mobile SDKs, you want to use as much of the existing deserialization logic as possible. For that, you want to match your Proto format. From looking at your Proto, it seems that you could also add an `exists` to `IBundledDocumentMetadata` and omit the document?

Drive by comment on Proto: We may also want to use "name" for document keys since this is what the backend uses.

dev/system-test/firestore.ts
schmidt-sebastian: Works for me.

dev/test/bundle.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from './util/helpers';
 import IBundleElement = firestore.IBundleElement;
 
-describe('Bundle Buidler', () => {
+describe.only('Bundle Buidler', () => {
```
schmidt-sebastian, line 27: If you merge @thebrianchen's changes form the Node10 branch into your PR you will get a lint check that will catch this :)

dev/src/bundle.ts
```diff
@@ -49,10 +46,7 @@ export function validateDocumentSnapshot(
  * @param arg The argument name or argument index (for varargs methods).
  * @param value The input to validate.
  */
-export function validateQuerySnapshot(
-  arg: string | number,
-  value: unknown
-): void {
+function validateQuerySnapshot(arg: string | number, value: unknown): void {
```
schmidt-sebastian, line 49: Please do move to bottom of the module to match the code layout in the rest of the repo.

dev/test/bundle.ts
wu-hui: There is a lot changes from bulk writer too. I will do the merge later..

wu-hui:
> Also added a version to metadata.

One more thing: the change to make it non-public will be separate, such that it can be easily reverted.

dev/protos/firestore/bundle.proto
```diff
@@ -85,6 +88,9 @@ message BundleMetadata {
 
   // Time at which the documents snapshot is taken for this bundle.
   google.protobuf.Timestamp create_time = 2;
+
+  // The version of bundle.
```
schmidt-sebastian, line 92: Nit: "The _schema_ version of _the_ bundle."

dev/src/bundle.ts
```diff
@@ -109,10 +85,11 @@ export class BundleBuilder {
   private addBundledDocument(snap: DocumentSnapshot) {
     const docProto = snap.toDocumentProto();
     this.documents.set(snap.id, {
-      document: docProto,
+      document: snap.exists ? docProto : undefined,
```
schmidt-sebastian, line 88: Nit: I don't know that snap.toDocumentProto() does if the document doesn't exist. I would not need to check if you called it here inline.

dev/src/bundle.ts
```diff
@@ -174,16 +152,19 @@ export class BundleBuilder {
     for (const bundledDocument of this.documents.values()) {
       const documentMetadata: firestore.IBundledDocumentMetadata =
         bundledDocument.metadata;
-      const document: api.IDocument = bundledDocument.document;
+      const document = bundledDocument.document;
```
schmidt-sebastian, line 155: Nit: Move to line 162.

dev/protos/firestore/bundle.proto
wu-hui: Done.

dev/src/bundle.ts
wu-hui: there is another `docProto.name` below, so I prefer to keeping it this way.

dev/src/bundle.ts
wu-hui: Done.

PR #1056, opened 2020-04-29
chore: add vscode files

thebrianchen:


.gitignore
```diff
@@ -9,6 +9,5 @@ system-test/*key.json
 *.lock
 build/
 package-lock.json
-.vscode
```
schmidt-sebastian, line 12: This will be overwritten again. IMHO it is fine to add the files and still have them in .gitignore, it just makes the task of modifying the files more explicit.

PR #1059, opened 2020-04-30
docs: Add documentation for FirestoreDataConverter

thebrianchen:
Looking at the docs, I realized that `FirestoreDataConverter` was nowhere to be found since I didn't include it in `index.ts`. I added links to it in each the `withConverter` `@param` definitions as well. However, I'm not sure how to properly include the method docs for to/fromFirestore in JSdoc without actually writing out code for the interface. I included the methods as `@property`, which isn't ideal but is better that nothing.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -535,11 +535,10 @@ export class DocumentReference<T = DocumentData> implements Serializable {
   }
 
   /**
-   * Applies a custom data converter to this DocumentReference, allowing you
-   * to use your own custom model objects with Firestore. When you call
-   * set(), get(), etc. on the returned DocumentReference instance, the
-   * provided converter will convert between Firestore data and your custom
-   * type U.
+   * Applies a custom data converter to this DocumentReference, allowing you to
+   * use your own custom model objects with Firestore. When you call set(),
+   * get(), etc. on the returned DocumentReference instance, the provided
+   * converter will convert between Firestore data and your custom type U.
    *
    * Using the converter allows you to specify generic type arguments when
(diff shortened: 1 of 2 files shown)


dev/src/reference.ts
```diff
@@ -576,7 +575,8 @@ export class DocumentReference<T = DocumentData> implements Serializable {
    *   post.someNonExistentProperty; // TS error
    * }
    *
-   * @param converter Converts objects to and from Firestore.
+   * @param {FirestoreDataConverter=} converter Converts objects to and from
```
schmidt-sebastian, line 578: Drop `=` since it is not optional.

dev/src/reference.ts
```diff
@@ -2176,7 +2176,8 @@ export class Query<T = DocumentData> {
...
```
schmidt-sebastian, line 2179: Drop `=` since it is not optional.

dev/src/reference.ts
```diff
@@ -2450,7 +2451,8 @@ export class CollectionReference<T = DocumentData> extends Query<T> {
...
```
schmidt-sebastian, line 2454: Drop `=` since it is not optional.

dev/src/reference.ts
thebrianchen: done.

dev/src/reference.ts
thebrianchen: done.

dev/src/reference.ts
thebrianchen: done.

Issue #1000, opened 2020-04-05
WriteBatch.set should have a signature override


sk-:
`WriteBatch.set` changes its semantics when passing options, specially when providing `{merge: true}`

In that case an overload to the signature should be added, as in that case there's no need to provide a full document.

The [original definition](c89546f5ae83da3845076aeeffcda75f9b208f5c) is:

```ts
set<T>(documentRef: DocumentReference<T>, data: T, options?: SetOptions): WriteBatch;
``` 

The proposed overload would be something like:
```ts
set<T>(documentRef: DocumentReference<T>, data: Partial<T>, {merge: true}): WriteBatch;
```

schmidt-sebastian:
@sk Oh, that makes a lot of sense, and it doesn't seem like a breaking change. We will get this done shortly.

schmidt-sebastian:
@thebrianchen Brian, can you fix this here and in the Web SDK?

sk-:
This applies to `DocumentReference.set`, `Transaction.set`, and `WriteBatch.set`

thebrianchen:
@sk- After looking into this further, we realized that accepting an overload for `Partial<T>` in `set()` would require modifying the signature of `FirestoreDataConverter.toFirestore()` to also take `Partial<T>`, since Firestore needs to first convert user data into `DocumentData` ([code](https://github.com/googleapis/nodejs-firestore/blob/19c2c75d86c3aab967d21da16598016185ae360b/dev/src/write-batch.ts#L299)).

Expanding the signature of `toFirestore()` to `Partial<T>` would be a breaking change. We're going to look into other options, as well as what this means for the Web SDK. If you have any suggestions, please join in :)

sk-:
@thebrianchen actually the types for `set` and `toFirestore` are currently not correct, as they do not allow to specify [`FieldValue`](https://firebase.google.com/docs/reference/js/firebase.firestore.FieldValue)s.

Also note that currently if you are using a converter, you cannot specify partial objects with either `{merge: true}` or `{mergeFields: [...]}`, and you can't use `FieldValue`s.

Using something like https://pirix-gh.github.io/ts-toolbelt/modules/_object_update_.html, would allow to update the fields.

```ts
import {Any, Union} from 'ts-toolbelt';
import {FieldValue} from '@google-cloud/firestore'; 

export type UpdateAll<O extends object, A extends any> = {
  [P in keyof O]: Union.Replace<A, Any.x, O[P]>;
[...]

PR #1060, opened 2020-05-01
fix: add toFirestoreFromMerge()

thebrianchen:
Original issue: #1000


dev/src/reference.ts
```diff
@@ -375,13 +376,16 @@ export class DocumentReference<T = DocumentData> implements Serializable {
...
 
+  set(data: Partial<T>, options: SetOptions): Promise<WriteResult>;
+  set(data: T, options?: SetOptions): Promise<WriteResult>;
   /**
    * Writes to the document referred to by this DocumentReference. If the
    * document does not yet exist, it will be created. If you pass
    * [SetOptions]{@link SetOptions}, the provided data can be merged into an
-   * existing document.
+   * existing document. Using Partial objects requires
+   * [SetOptions]{@link SetOptions} to be passed in as well.
```
schmidt-sebastian, line 386: I feel this detail is confusion. We should expand on this in an additional paragraph or omit it.

dev/src/reference.ts
```diff
@@ -398,12 +402,18 @@ export class DocumentReference<T = DocumentData> implements Serializable {
...
+  set(data: T | Partial<T>, options?: SetOptions) {
+    let writeBatch = new WriteBatch(this._firestore);
+    if (
+      options &&
+      (options.merge || options.mergeFields) &&
+      !isDefaultConverter(this._converter)
+    ) {
+      writeBatch = writeBatch.set(this, data as Partial<T>, options);
+    } else {
+      writeBatch = writeBatch.set(this, data as T, options);
```
schmidt-sebastian, line 414: If you add "the catchall" signature as a overload to WriteBatch you won't have to do this dance:

dev/src/transaction.ts
```diff
@@ -256,10 +268,18 @@ export class Transaction {
...
   ): Transaction {
-    this._writeBatch.set(documentRef, data, options);
+    if (
+      options &&
+      (options.merge || options.mergeFields) &&
+      !isDefaultConverter(documentRef._converter)
+    ) {
+      this._writeBatch.set(documentRef, data as Partial<T>, options);
+    } else {
+      this._writeBatch.set(documentRef, data as T, options);
```
schmidt-sebastian, line 281: Same as above - you should be able to remove this.

dev/src/types.ts
```diff
@@ -144,6 +144,16 @@ export interface FirestoreDataConverter<T> {
    */
   toFirestore(modelObject: T): DocumentData;
 
+  /**
+   * Called by the Firestore SDK to convert a custom model object of Partial<T>
+   * into a plain Javascript object (suitable for writing directly to the
+   * Firestore database).
+   *
+   * This method must be defined in order to use set() with objects of type
+   * Partial<T>.
```
schmidt-sebastian, line 153: Please rewrite this part to refer to "merge". First paragraph is fine and dandy.

dev/src/types.ts
```diff
@@ -164,6 +174,23 @@ export const defaultConverter: FirestoreDataConverter<DocumentData> = {
...
 
+/**
+ * Checks if the provided converter is the same as the default converter.
+ * @private
+ */
+export function isDefaultConverter<U>(
+  converter: FirestoreDataConverter<U>
+): boolean {
+  return (
+    converter.toFirestore.toString() ===
```
schmidt-sebastian, line 185: Isn't this just `converter === defaultConverter`?

dev/src/write-batch.ts
```diff
@@ -268,16 +269,27 @@ export class WriteBatch {
...
   /**
    * Write to the document referred to by the provided
-   * [DocumentReference]{@link DocumentReference}.
-   * If the document does not exist yet, it will be created. If you pass
-   * [SetOptions]{@link SetOptions}., the provided data can be merged
-   * into the existing document.
+   * [DocumentReference]{@link DocumentReference}. If the document does not
+   * exist yet, it will be created. If you pass [SetOptions]{@link SetOptions},
+   * the provided data can be merged into the existing document. Using Partial
+   * objects requires [SetOptions]{@link SetOptions} to be passed in as well.
```
schmidt-sebastian, line 288: Same comment regarding refocusing this on "merge".

dev/src/write-batch.ts
```diff
@@ -300,14 +312,29 @@ export class WriteBatch {
...
     validateDocumentReference('documentRef', documentRef);
-    let firestoreData = documentRef._converter.toFirestore(data);
+    let firestoreData: DocumentData;
+    if (
+      (mergeLeaves || mergePaths) &&
+      !isDefaultConverter(documentRef._converter)
+    ) {
+      if (documentRef._converter.toFirestoreFromPartial === undefined) {
+        throw new Error(
+          'toFirestoreFromPartial() must be defined to use merge with Partials.'
```
schmidt-sebastian, line 329: Nit: "to use set() with `merge` or `mergeFields`"

dev/system-test/firestore.ts
```diff
@@ -2137,6 +2137,22 @@ describe('WriteBatch class', () => {
       });
   });
 
+  it('set supports partials', async () => {
+    const ref = randomCol.doc('doc').withConverter(postConverter);
+    await ref.set(new Post('walnut', 'author'));
+    const batch = firestore.batch();
+    batch.set(ref, {title: 'olive'} as Partial<Post>, {merge: true});
```
schmidt-sebastian, line 2144: We need to write our API to ensure that users don't have to do this cast:

```
set<T>(
    documentRef: DocumentReference<T>,
    data: Partial<T>,
    options: { merge: .. } | { mergePaths : ... }
  ): Transaction;
  set<T>(
    documentRef: DocumentReference<T>,
    data: T
  ): Transaction;
```

types/firestore.d.ts
```diff
@@ -91,6 +91,16 @@ declare namespace FirebaseFirestore {
...
 
+    /**
+     * Called by the Firestore SDK to convert a custom model object of
+     * Partial<T> into a plain Javascript object (suitable for writing directly
+     * to the Firestore database).
+     *
+     * This method must be defined in order to use set() with objects of type
+     * Partial<T>.
+     */
+    toFirestoreFromPartial?(modelObject: Partial<T>): DocumentData;
```
schmidt-sebastian, line 102: Should this be `toFirestoreFromMerge`?

Even though you made this optional, this PR still smells like a breaking change to me since you added the validation that rejects calls if this is not set. I wonder if "merge" is a prominent enough part of our API that warrants that we should require that this is defined. What do you think? I am leaning 50% in favor of dropping the question mark.

dev/src/reference.ts
thebrianchen: wow that's clean! thanks

dev/src/transaction.ts
thebrianchen: done.

dev/system-test/firestore.ts
thebrianchen: Can't specify all the options since the `SetOptions` type allows both `merge` and `mergeFields` to be passed in, but I made `SetOption` required, which still fixed the need for cast.

dev/src/reference.ts
thebrianchen: Omitted since we already explain it in `toFirestoreFromMerge`

dev/src/types.ts
thebrianchen: done.

dev/src/write-batch.ts
thebrianchen: done.

types/firestore.d.ts
thebrianchen: This would be a breaking change only for users using merge with converters. I don't see why anyone would be passing in `merge: true` with `<T>` objects, though I could see this breaking for existing developers using this with `mergeFields`.

However, wouldn't requiring this on the interface be a breaking change as well? Requiring `toFirestoreFromMerge()` would break developers using the converter without merge, and force developers who don't need to merge (ex: typical BulkWriter use case) to define a dummy method that they're not going to use. If I understand this correctly, the number of users that would be broken if we required this would be greater. Also the fact that we haven't received any feature requests for merge on Web could mean that there aren't that many developers using converters with merge. 

Considering that we're already making a breaking change with the change from `DocumentData` to `QueryDocumentSnapshot` in `fromFirestore()`, I'm leaning towards adding this as optional. Thoughts?

dev/src/write-batch.ts
```diff
@@ -302,14 +313,30 @@ export class WriteBatch {
...
     const mergeLeaves = options && options.merge === true;
     const mergePaths = options && options.mergeFields;
     validateDocumentReference('documentRef', documentRef);
-    let firestoreData = documentRef._converter.toFirestore(data);
+    let firestoreData: DocumentData;
+    if (
+      (mergeLeaves || mergePaths) &&
+      documentRef._converter !== defaultConverter
+    ) {
+      if (documentRef._converter.toFirestoreFromMerge === undefined) {
```
schmidt-sebastian, line 328: Might as well make this stricter:

```
typeof documentRef._converter.toFirestoreFromMerge !== 'function'
```

dev/test/util/helpers.ts
```diff
@@ -334,6 +334,9 @@ export const postConverter = {
   toFirestore(post: Post): DocumentData {
     return {title: post.title, author: post.author};
   },
+  toFirestoreFromMerge(post: Partial<Post>): DocumentData {
+    return Object.assign({}, post);
```
schmidt-sebastian, line 338: This is so 2019. All the cool kids use `{... post}`. :)

dev/test/write-batch.ts
```diff
@@ -76,6 +78,24 @@ describe('set() method', () => {
     nullObject.bar = 'ack';
     writeBatch.set(firestore.doc('sub/doc'), nullObject);
   });
+
+  it('requires toFirestoreFromMerge() for Partial usage', () => {
+    const converter = postConverter;
+    delete converter.toFirestoreFromMerge;
```
schmidt-sebastian, line 84: This likely breaks the tests if they are run out of order since you are modifying an instance that is used in other tests.

dev/test/write-batch.ts
```diff
@@ -76,6 +78,24 @@ describe('set() method', () => {
...
+    ).to.throw(
+      'toFirestoreFromMerge() must be defined to use set() with `merge` or `mergeFields`.'
+    );
+  });
+
+  it('requires SetOptions to use partials', () => {
+    const ref = firestore.doc('sub/doc').withConverter(postConverter);
+    expect(() =>
+      (writeBatch as InvalidApiUsage).set(ref, {title: 'foo'} as Partial<Post>)
+    ).to.throw('Value for argument "data" is not a valid Firestore document.');
```
schmidt-sebastian, line 97: How does you code know that  `{title: 'foo'}` is not a `Post`?

types/firestore.d.ts
schmidt-sebastian: I think either way this is a breaking change and we probably need to wait until later to roll this out to the Web SDK. What you have proposed is probably less breaking than requiring the interface.

dev/src/write-batch.ts
thebrianchen: Done. good tip!

dev/test/util/helpers.ts
thebrianchen: However futile it might be, I appreciate you trying to up my cool factor

dev/test/write-batch.ts
thebrianchen: Tried to be a cool 2020 kid and used the spread operator.

dev/test/write-batch.ts
thebrianchen: Oops forgot to remove.

schmidt-sebastian: Oh, does this throw because author is now `undefined`? I feel like that it is likely more a test of the converter. I suggest to rework is as a converter test or to remove.

thebrianchen: Removed the test, since we already have other tests that check for undefined values.

thebrianchen:
Continued in #1087.

Issue #1031, opened 2020-04-19
Ignore undefined values rather than throw exceptions


nojvek:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/nodejs-firestore/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS: Mac
  - Node.js version: 12
  - npm version: 6.13.6
  - `@google-cloud/firestore` version: 3.7.4"

#### Steps to reproduce

  1. Insert using `fstore.collection(`blah`).doc(`foo`).set({"hello": undefined})`
  
### Expected

Like JSON.stringify, undefined keys are ignored.

### Actual

Error: Value for argument "data" is not a valid Firestore document. Cannot use "undefined" as a Firestore value (found in field "hello").

### Explanations

undefines are a routine part of js. Not gracefully handling undefines means firestore becomes a brittle library that barfs every now and then. The developer experience could be much better.

A quick dirty workaround is `doc.set(JSON.parse(JSON.stringify(val))`, which is yuck. The library can handle this elegantly and preserve JSON.stringify like behaviour.

schmidt-sebastian:
@nojvek Thanks for this request! This has come up a couple of times. I will talk with the team to see if this is a behavior we can support.

nojvek:
If you're open to a PR, I can submit one. I did a log of debugging yest to see where this was failing.

schmidt-sebastian:
This is more of a political challenge rather than a technical one, as this might also lead to unexpected behavior. Stripping `undefined` breaks the assumption that data the read from Firestore is the same as the data written to, and as such I am expecting some pushback. If we do decide to change the behavior, then I would be more than happy to accept a PR.

merlinnot:
That's definitely a two edged sword - I personally like the current behavior and I saw a couple of instances where it saved people from writing invalid data to the DB. We could also expose it as an option, so that people can decide for themselves.

nojvek:
[...]

PR #1062, opened 2020-05-01
feat: add ignoreUndefinedProperties option

schmidt-sebastian:
This adds an optional setting to ignore nested undefined values in all API calls that accept user data.

Fixes https://github.com/googleapis/nodejs-firestore/issues/1031

### DIFF
--- a/dev/test/ignore-undefined.ts
+++ b/dev/test/ignore-undefined.ts
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 import {expect} from 'chai';
 import {fieldFilters, orderBy, queryEquals, startAt} from './query';
 import {
--- a/dev/test/ignore-undefined.ts
+++ b/dev/test/ignore-undefined.ts
@@ -1,0 +1,283 @@
+import {expect} from 'chai';
+import {fieldFilters, orderBy, queryEquals, startAt} from './query';
+import {
+  ApiOverride,
+  create,
+  createInstance,
+  document,
+  InvalidApiUsage,
+  requestEquals,
+  response,
+  set,
+  stream,
+  update,
+  updateMask,
+  writeResult,
+} from './util/helpers';
+
+const FOO_MAP = {
+  mapValue: {
+    fields: {
+      bar: {
+        stringValue: 'bar',
+      },
+    },
+  },
+};
+
+describe('ignores undefined values', () => {
+  it('in set()', () => {
+    const overrides: ApiOverride = {
+      commit: request => {
+        requestEquals(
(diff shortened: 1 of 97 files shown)


nojvek:
This is great. Thanks. 

Pretty cool how battle tested such a minor feature is. 

To be honest I’m far in love with the nodejs sdk compared to the browser sdk. The browser sdk is quite shallow and a pain to use. Doesn’t have listCollectionIds and a bunch of other commands that are here. I hope someday this sdk will be usable in the browser too with the same api. The api is well thought out.

schmidt-sebastian:
Just FYI: `listCollectionIds()` and `listDocuments()` require Service Account credentials and are not callable with the Firebase user accounts that the Web SDK uses for authentication. For this reason, listCollectionIds() is not available in the Web SDK. 

Thank you for your kind words.

dev/test/ignore-undefined.ts
```diff
@@ -0,0 +1,297 @@
...
+              .startAt(undefined);
+          }).to.throw(
+            '"undefined" values are only ignored in object properties.'
+          );
+        }
+      );
+    });
+  });
+
+  describe('when setting is disabled', () => {
```
wu-hui, line 232: `Reject undefined fields when setting is disabled`?

types/firestore.d.ts
```diff
@@ -143,6 +143,13 @@ declare namespace FirebaseFirestore {
      * when the client becomes idle. Defaults to 1.
      */
     maxIdleChannels?: number;
+
+    /**
+     * Whether to skip nested properties that are set to `undefined` during object
+     * serialization. If set to `false` or omitted, the SDK will throw an
+     * exception when it encounters values of type `undefined`.
```
wu-hui, line 150: Do you think we should also document the effects (or non-effect) during deserialization?

wu-hui:
Sorry for the delayed review!

dev/test/ignore-undefined.ts
schmidt-sebastian: The full test names are: 
"rejects undefined values when setting is disabled in set()"
"rejects undefined values when setting is disabled in create()"
...

The name starts with the first "define" in line 158.

types/firestore.d.ts
schmidt-sebastian: Sounds good. Changed to:

```
  /**
   * Whether to skip nested properties that are set to `undefined` during object
   * serialization. If set to `true`, these properties will be skipped and are 
   * not be written to Firestore. If the setting is set `false` or omitted, 
   * the SDK will throw an exception when it encounters properties of type 
   * `undefined`.
   */
```

kladnik:
@schmidt-sebastian Might be a silly question, but how do I enable this option? I am using `firebase-admin` and I cannot find a way to enable it, but I am getting: 
```
Error: Value for argument "data" is not a valid Firestore document. Cannot use "undefined" as a Firestore value (found in field "progress.test"). If you want to ignore undefined values, enable `ignoreUndefinedProperties`.
```
Which means that this feature should already be working.

schmidt-sebastian:
@kladnik See the following:

```
const firestore  = admin.firestore();
firestore.settings({ignoreUndefinedProperties:true});
```

Make sure that you have installed the latest version of `@google-cloud/firestore`. You may have to remove your `package-lock.json` and your `node_modules` folder.

kladnik:
@schmidt-sebastian Thanks! Might be good to add this example in the docs somewhere, so it will help others aswell.

guness:
@schmidt-sebastian , totally agree with @kladnik .

I was really happy that you solved this after spending two hours how to fix the problem until realizing it is not working somehow.

then took one more hour to find out that solution actually requires deletion of those two.

Issue #1058, opened 2020-04-30
Typeguard for exists


sk-:
Would you consider adding an alternate way of calling `exists` in `DocumentSnapshot`s which support a Typeguard?

At the moment if you do:

```ts
const snapshot = ref.get()
if (snapshot.exists) {
  const data = snapshot.data();
  // here data can still be undefined to the TS compiler
}
```

The problem is that `exists` does not have a type guard, and in fact they are not supported in getters. See https://github.com/Microsoft/TypeScript/issues/7150.

However, if you could add a new method for `exists` you could solve the problem, with something like the following:

```ts
class DocumentSnapshopt<T> {
    private _fieldsProto: ApiMapValue | undefined;
    newExists(): this is QueryDocumentSnapshot<T> {
        return this.raw !== undefined;
    }
    data() {
        if (this._fieldsProto === undefined) {
            return undefined;
        }
        return ...
    }
}

const rawResult = new RawResult();

if (rawResult.exists()) {
    const data = rawResult.decode()
}
```

See https://stackoverflow.com/questions/61510649/typeguard-for-property/61520437#61520437 and this [TS Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=2&pc=1#code/C4TwDgpgBAShDOBXANsKBeKBvAvgKDwEsA7YCAJwDMBDAY2hmoHc4lUBRAD0PmHigicyxACb9GLBCjRY8UeVBERaAeyUAKAJQAuWFNR58eWsmrxxzVtOxyFYcoQBu1MlHLNdvB8QDmUAD5QiKIQlCQQIgDctvKCPHxausAAFjxQaRJWHNy8-LIKBW4QwIjkxFApPAB07kxQAITomMFKYcQR0QX4BUqqGjp6bGiBLaHhIjaFCoSUUOqV8DXMGE1BIW0RmpNTheTFpeWjG1Exhd07eyVlFamLtVBmg9Kn+EaqxLxulvpomO11mR+WmixhUHzQIhc1AAstQQAAhCAAVXW4wwX0kQyqvTUEGBRFm6lqWWAVTiuS0W3yCgA9DSoAAJCjQSHAahQWjUYjEFRoABG0CO41O70+rPZmGJP2xylxWkMeCAA)

schmidt-sebastian:
@sk- Thanks for filing this. We have been trying to do this for a long long time. I will play around with it to see if we can introduce this as a non-breaking change.

[...]

PR #1064, opened 2020-05-01
RFC feat: type-guards for DocumentSnapshot

schmidt-sebastian:
Proposal only. Not for submission.

Addresses https://github.com/googleapis/nodejs-firestore/issues/1058

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -216,7 +216,7 @@ declare namespace FirebaseFirestore {
      * snapshots.
      */
     getAll(...documentRefsOrReadOptions: Array<DocumentReference<DocumentData> | ReadOptions>):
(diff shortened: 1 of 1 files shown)


types/firestore.d.ts
```diff
@@ -952,7 +954,7 @@ declare namespace FirebaseFirestore {
      * @param snapshot The snapshot of the document to start after.
      * @return The created Query.
      */
-    startAt(snapshot: DocumentSnapshot<any>): Query<T>;
+    startAt(snapshot: (DocumentSnapshot<any> | QueryDocumentSnapshot<any>)): Query<T>;
```
schmidt-sebastian, line 957: Note that this is non-sensical but avoids a breaking change.

PR #1067, opened 2020-05-01
chore: remove internal WriteOp

schmidt-sebastian:


dev/src/write-batch.ts
```diff
@@ -337,18 +326,14 @@ export class WriteBatch {
         documentMask = DocumentMask.fromObject(firestoreData);
       }
 
-      const write = document.toProto();
+      const update = document.toProto();
```
thebrianchen, line 329: Is there a reason why this variable is called `update`, but the other variables are `write`? They all have type `api.IWrite`. Consider using one or the other?

schmidt-sebastian: No, no reason. Updated to `write`.

PR #1070, opened 2020-05-05
fix: handle terminate in BulkWriter

thebrianchen:


dev/src/index.ts
```diff
@@ -1117,9 +1147,23 @@ export class Firestore {
         'All onSnapshot() listeners must be unsubscribed before terminating the client.'
       );
     }
+    if (this.pendingOperationsCount > 0) {
+      return Promise.reject(
+        'All BulkWriter operations must be completed before terminating the client.'
```
schmidt-sebastian, line 1152: Mention that the user should call and await `close()`.

dev/src/index.ts
```diff
@@ -349,11 +349,21 @@ export class Firestore {
...
   private registeredListenersCount = 0;
 
+  /**
+   * Number of pending operations on the client.
+   *
+   * The client can only be terminated when there are no pending writes or
+   * registered listeners.
+   * @private
+   */
+  private pendingOperationsCount = 0;
```
schmidt-sebastian, line 365: Make this name bulk writer specific, since we are not sharing it with other logic. There could be many other pending operations (such as normal set()s).

dev/system-test/firestore.ts
```diff
@@ -153,7 +154,7 @@ describe('Firestore class', () => {
       })
       .then(() => Promise.reject('set() should have failed'))
       .catch(err => {
-        expect(err).to.equal('The client has already been terminated');
+        expect(err).to.equal(CLIENT_TERMINATED_ERROR_MSG);
```
schmidt-sebastian, line 157: FWIW Having the actual message here makes typos much more apparent.

Optional.

dev/system-test/firestore.ts
```diff
@@ -2265,6 +2280,40 @@ describe('QuerySnapshot class', () => {
...
+  let randomCol: CollectionReference;
+
+  beforeEach(() => {
+    firestore = new Firestore({});
+    randomCol = getTestRoot(firestore);
+  });
+
+  afterEach(() => verifyInstance(firestore));
+
+  it.skip('increments and decrements the pendingOperationsCount', async () => {
```
schmidt-sebastian, line 2294: What about something like "can terminate() when BulkWrite is closed"?

dev/system-test/firestore.ts
```diff
@@ -2265,6 +2280,40 @@ describe('QuerySnapshot class', () => {
...
+    writer.set(ref, {foo: 'bar'});
+    writer.set(ref, {foo: 'bar2'});
+    await writer.close();
+    return firestore.terminate();
+  });
+
+  it('subsequent operations fail after terminate() is called', () => {
+    const ref = randomCol.doc('doc1');
+    const writer = firestore._bulkWriter();
+    return firestore.terminate().then(() => {
```
schmidt-sebastian, line 2306: Can you rewrite to follow this pattern: https://github.com/googleapis/nodejs-firestore/blob/master/dev/system-test/firestore.ts#L2060 ?

dev/src/index.ts
thebrianchen: Updated message.

dev/system-test/firestore.ts
thebrianchen: done.

dev/system-test/firestore.ts
thebrianchen: done.

dev/system-test/firestore.ts
thebrianchen: done.

dev/src/index.ts
thebrianchen: Renamed to `bulkWritersCount`

dev/src/bulk-writer.ts
```diff
@@ -502,12 +504,16 @@ export class BulkWriter {
    * });
    */
   close(): Promise<void> {
+    if (!this.closed) {
```
schmidt-sebastian, line 507:
```suggestion
    this.verifyNotClosed();
    this.firestore.decrementBulkWritersCount();
```

This makes it more obvious since this.flush() calls it as well, but in a non-obvious way.

dev/system-test/firestore.ts
schmidt-sebastian: I believe you forgot to push this, but maybe I am not supposed to review this yet :)

dev/src/bulk-writer.ts
thebrianchen: this makes more sense, thanks

dev/system-test/firestore.ts
thebrianchen: I think you were viewing the outdated one?

dev/system-test/firestore.ts
```diff
@@ -173,6 +173,20 @@ describe('Firestore class', () => {
       unsubscribe();
     }
   });
+
+  it('throws an error if terminate() is called with pending BulkWriter operations', async () => {
+    const writer = firestore._bulkWriter();
+    const ref = randomCol.doc('doc-1');
+    writer.set(ref, {foo: 'bar'});
+    try {
```
schmidt-sebastian, line 181: I still see the try/catch code here. Did you update this test to use `eventually.be.rejectedWith` as well?

thebrianchen: Confused which test you were referring to, but it has been now :)

thebrianchen:
Realized that we don't need `verifyNotTerminated()` since `terminate()` cannot be called unless BulkWriter has been closed first, which eliminates the need for the check.

dev/src/index.ts
```diff
@@ -1117,6 +1147,12 @@ export class Firestore {
         'All onSnapshot() listeners must be unsubscribed before terminating the client.'
       );
     }
+    if (this.bulkWritersCount > 0) {
+      return Promise.reject(
+        'All BulkWriter instances must be closed by calling and awaiting ' +
+          '`BulkWriter.close()` before terminating the client.'
+      );
+    }
```
schmidt-sebastian, line 1155: Optional suggestion: Could we combine the error message to something like:

"All onSnapshot() listeners must be unsubscribed and BulkWriter instances must be closed  before terminating the client. There are ${this.registeredListenersCount} active listeners and ${this. bulkWritersCount} that are still performing writes."

This gives more information and avoids the disappointment when the users have unregistered all listeners but still cannot terminate the client.

dev/system-test/firestore.ts
```diff
@@ -173,6 +173,16 @@ describe('Firestore class', () => {
       unsubscribe();
     }
   });
+
+  it('throws an error if terminate() is called with pending BulkWriter operations', async () => {
+    const writer = firestore._bulkWriter();
+    const ref = randomCol.doc('doc-1');
+    writer.set(ref, {foo: 'bar'});
+    expect(firestore.terminate()).to.eventually.be.rejectedWith(
```
schmidt-sebastian, line 181: This expectation is now a Promise that has to be awaited.

dev/system-test/firestore.ts
```diff
@@ -2265,6 +2275,27 @@ describe('QuerySnapshot class', () => {
...
+  let randomCol: CollectionReference;
+
+  beforeEach(() => {
+    firestore = new Firestore({});
+    randomCol = getTestRoot(firestore);
+  });
+
+  afterEach(() => verifyInstance(firestore));
+
+  it.skip('can terminate once BulkWriter is closed', async () => {
```
schmidt-sebastian, line 2289: Add TODO(bulkwriter) or something enable this test>

thebrianchen: done.

dev/system-test/firestore.ts
thebrianchen: done.

dev/src/index.ts
thebrianchen: combined.

PR #1073, opened 2020-05-08
fix: Hide bundle builder from public API

wu-hui:
Hides bundle builder before Web SDK is able to load bundles.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -758,7 +758,7 @@ export class Firestore {
    *
    * @private
    */
-  bundle_(name?: string): BundleBuilder {
+  bundle(name?: string): BundleBuilder {
     return new BundleBuilder(name || autoId());
   }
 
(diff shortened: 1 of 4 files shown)


dev/src/index.ts
```diff
@@ -755,8 +755,10 @@ export class Firestore {
    * @param bundleId. The id of the bundle. When loaded on clients, client SDKs use this id
    * and the timestamp associated with the built bundle to tell if it has been loaded already.
    * If not specified, a random identifier will be used.
+   *
+   * @private
    */
-  bundle(name?: string): BundleBuilder {
+  bundle_(name?: string): BundleBuilder {
```
schmidt-sebastian, line 761: s/_bundle/bundle_

This is the much more common way to hide private APIs.

`snapshot_` is named `snapshot_` because it is actually part of the public interface since it is consumed by Firebase Functions.

wu-hui: Done.

PR #1078, opened 2020-05-12
feat: Firestore Bundles implementation

wu-hui:


dev/src/bundle.ts
```diff
@@ -0,0 +1,208 @@
...
+  validateString,
+} from './validate';
+
+import api = google.firestore.v1;
+
+const BUNDLE_VERSION = 1;
+
+/**
+ * Builds a Firestore data bundle with results from the given document and query snapshots.
+ */
```
schmidt-sebastian, line 33: Nit: Add `@private` for now to hide from JSDoc.

wu-hui: Done.

PR #1085, opened 2020-05-19
feat: Add totalDocuments and totalBytes to bundle metadata.

wu-hui:


dev/src/bundle.ts
```diff
@@ -168,6 +159,18 @@ export class BundleBuilder {
...
+      id: this.bundleId,
+      createTime: this.latestReadTime.toProto().timestampValue,
+      version: BUNDLE_VERSION,
+      totalDocuments: this.documents.size,
+      totalBytes: bundleBuffer.length,
+    };
+    bundleBuffer = Buffer.concat([
+      this.elementToLengthPrefixedBuffer({metadata}),
+      bundleBuffer,
+    ]);
```
schmidt-sebastian, line 173: I am not sure I understand how this will work. I assume this was added to allow the progress handler to display their status. For that to work, would this not need to be the first message?

wu-hui: Note the order of the arguments to `Buffer.concat` is different here. Also added a comment.

dev/src/bundle.ts
```diff
@@ -168,6 +159,19 @@ export class BundleBuilder {
...
     }
+
+    const metadata: firestore.IBundleMetadata = {
+      id: this.bundleId,
+      createTime: this.latestReadTime.toProto().timestampValue,
+      version: BUNDLE_VERSION,
+      totalDocuments: this.documents.size,
+      totalBytes: bundleBuffer.length,
+    };
+    // Prepends the metadata element to the bundleBuffer, note `bundleBuffer` is the second argument to `Buffer.concat`.
```
schmidt-sebastian, line 170: s/, note/:/

```
 // Prepends the metadata element to the bundleBuffer: `bundleBuffer` is the second argument to `Buffer.concat`.
```

dev/test/bundle.ts
```diff
@@ -80,11 +80,14 @@ describe('Bundle Buidler', () => {
     expect(elements.length).to.equal(3);
 
     const meta = (elements[0] as IBundleElement).metadata;
+    expect(meta!.totalBytes).greaterThan(0);
```
schmidt-sebastian, line 83: Is the size not fixed here in the unit tests? Or will that make the tests too fragile?

dev/test/bundle.ts
```diff
@@ -128,11 +131,14 @@ describe('Bundle Buidler', () => {
     expect(elements.length).to.equal(4);
 
     const meta = (elements[0] as IBundleElement).metadata;
```
schmidt-sebastian, line 133: Do you mind adding a helper for the metadata comparison?

dev/src/bundle.ts
wu-hui: Done.

dev/test/bundle.ts
wu-hui: Yes, I started with checking for exact size, but it became too fragile.

dev/test/bundle.ts
wu-hui: Done.

dev/test/bundle.ts
```diff
@@ -24,6 +24,16 @@ import {
   verifyInstance,
 } from './util/helpers';
 import IBundleElement = firestore.IBundleElement;
+import IBundleMetadata = firestore.IBundleMetadata;
+
+export function verifyMetadata(
+  meta: IBundleMetadata,
+  expected: IBundleMetadata
+): void {
+  expect(meta!.totalBytes).greaterThan(0);
```
schmidt-sebastian, line 33: Drop `!`

dev/test/bundle.ts
```diff
@@ -24,6 +24,16 @@ import {
   verifyInstance,
 } from './util/helpers';
 import IBundleElement = firestore.IBundleElement;
+import IBundleMetadata = firestore.IBundleMetadata;
+
+export function verifyMetadata(
+  meta: IBundleMetadata,
+  expected: IBundleMetadata
```
schmidt-sebastian, line 31: I was hoping you would pass in the expected state one at a time:

```
export function verifyMetadata(
  meta: IBundleMetadata,
  id: string,
  createTime: ReadTime,
  version: number,
  totalDocuments: 0
)
```

From that, you can simplify further. Extract a constant for the bundle ID, and use it everywhere. The version is also always a constant. So in the end you have:

```
export function verifyMetadata(
  meta: IBundleMetadata,
  createTime: ReadTime,
  totalDocuments: 0
)
```

You can then also compare each field directly:


```
  expect(meta.totalBytes).greaterThan(0);
  expect(meta.id).to.equal(BUNDLE_ID);
  expect(meta.version).to.equal(1);
  expect(meta.createTime).to.deep.equal(readTime.toProto());
  expect(meta.totalDocuments).to.equal(totalDocuments);
```

This removes the need to delete the field and ruins nasty surprises in the future.

wu-hui: Done.

dev/test/bundle.ts
wu-hui: Done.

PR #1087, opened 2020-05-21
feat: add support for set() with SetOptions when using FirestoreDataConverter

### LINKED ISSUE #1000: WriteBatch.set should have a signature override

thebrianchen:
Fixes #1000.


dev/src/write-batch.ts
```diff
@@ -296,14 +306,23 @@ export class WriteBatch {
...
   ): WriteBatch {
     validateSetOptions('options', options, {optional: true});
     const mergeLeaves = options && options.merge === true;
     const mergePaths = options && options.mergeFields;
     validateDocumentReference('documentRef', documentRef);
-    let firestoreData = documentRef._converter.toFirestore(data);
+    let firestoreData: DocumentData;
+    if (mergeLeaves || mergePaths) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      firestoreData = (documentRef._converter as any).toFirestore(
```
schmidt-sebastian, line 319: Please add a short comment on why we need this cast to `any`.

dev/test/util/helpers.ts
```diff
@@ -342,6 +342,19 @@ export const postConverter = {
   },
 };
 
+export const postConverterMerge = {
+  toFirestore(post: Partial<Post>): DocumentData {
```
schmidt-sebastian, line 346: Could we add SetOptions here and additional asserts? Something like that:

```
if (merge) {
expect(post).to.be.an.instanceOf(Post);
} else {
expect(post).to.not.be.an.instanceOf(Post);
}
```

dev/src/document.ts
```diff
@@ -382,12 +388,14 @@ export class DocumentSnapshot<T = DocumentData> {
...
     // We only want to use the converter and create a new QueryDocumentSnapshot
     // if a converter has been provided.
-    if (this.ref._converter !== defaultConverter) {
+    if (this.ref._converter !== defaultConverter()) {
       const untypedReference = new DocumentReference(
         this.ref.firestore,
         this.ref._path
       );
-      return this.ref._converter.fromFirestore(
+      return ((this.ref._converter as unknown) as FirestoreDataConverter<
```
schmidt-sebastian, line 396: This compiles for me as just "this.ref._converter.fromFirestore".

thebrianchen: done.

dev/test/util/helpers.ts
thebrianchen: Done, but flipped the expect statements.

dev/src/write-batch.ts
thebrianchen: done.

schmidt-sebastian:
Do you want to make this a "feat" before merging? This deserves more recognition :)

PR #1092, opened 2020-05-27
fix: incomplete expect in rate-limiter test

thebrianchen:
Oops. Ran into this while porting rate limiter.


dev/src/rate-limiter.ts
```diff
@@ -110,7 +110,7 @@ export class RateLimiter {
    * tokens. Used for testing the limiter.
```
schmidt-sebastian, line 110: Update comment above? s/date/time/

thebrianchen: done.

PR #1095, opened 2020-05-29
fix: authenticate as admin user when ssl:false is set

yuchenshi:
This fixes https://github.com/firebase/firebase-js-sdk/issues/3105#issuecomment-635541894.

Before this change, `db.settings({..., ssl: false})` will connect but without any credentials (similar to a client device that is not logged in), which causes Firestore Emulator to evaluate security rules for requests. This is surprising for an Admin SDK and it is extremely hard to troubleshoot.

After this change, the same code should correctly behave like admin. This also provides one additional way to get admin privileges in the Firestore Emulator if for some reasons the environment variables is not viable / desirable.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -378,13 +378,6 @@ export class Firestore {
       delete emulatorSettings.servicePath;
       delete emulatorSettings.apiEndpoint;
 
-      // Manually merge the Authorization header to preserve user-provided headers
-      emulatorSettings.customHeaders = Object.assign(
-        {},
-        emulatorSettings.customHeaders,
-        {Authorization: 'Bearer owner'}
-      );
-
       this.validateAndApplySettings(emulatorSettings);
     } else {
       this.validateAndApplySettings({...settings, ...libraryHeader});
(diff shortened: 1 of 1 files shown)


dev/src/index.ts
```diff
@@ -424,7 +417,21 @@ export class Firestore {
         if (this._settings.ssl === false) {
           const grpc = require('@grpc/grpc-js');
           const sslCreds = grpc.credentials.createInsecure();
-          client = new module.exports.v1({sslCreds, ...this._settings});
+
+          // Use user-provided headers, but provide an Authorization header by default
+          // so that connection is recognized as admin in Firestore Emulator. (If for
+          // some reason we're not connecting to the emulator, then this will result in
+          // denials with invalid token, rather than behave like clients not logged in.)
+          const customHeaders = Object.assign(
```
schmidt-sebastian, line 425: Nit: Replace with modern syntax:

```
const customHeaders = { Authorization: 'Bearer owner', ... this._settings.customHeaders };
```

yuchenshi: Done.

yuchenshi:
> Oh, one more thing: Before merging, you need to update your PR title as it will be used to generate release notes. Something like:
> 
> ```
> fix: authenticate as admin user when ssl:false is set
> ```

Done.

yuchenshi:
> Please verify that it is already not possible to connect to a Firestore backend with `grpc.credentials.createInsecure()` before merging.

Verified that https://grpc.github.io/grpc/node/grpc.credentials.html#.createInsecure also drops all credentials and cannot be combined with other credentials.

samtstern:
This change somehow broke the `FIRESTORE_EMULATOR_HOST` flow, see:
https://github.com/firebase/firebase-tools/issues/2310

PR #1110, opened 2020-06-03
feat: add Partition and BatchWrite protos

schmidt-sebastian:


synth.py
```diff
@@ -146,7 +146,6 @@
 
 os.rename("dev/.gitignore", ".gitignore")
 os.rename("dev/.eslintignore", ".eslintignore")
-os.rename("dev/.eslintrc.json", ".eslintrc.json")
```
thebrianchen, line 149: why are we removing this line again?

schmidt-sebastian: We probably want to keep our override that disallows describe.only.

PR #1222, opened 2020-06-17
fix: retry ABORTED writes in bulkCommit

thebrianchen:
With throttling enabled in the current state (no retries on backend), this implementation successfully gets all ABORTED writes to eventually succeed.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -24,7 +24,6 @@ import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
 import {Deferred, wrapError} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
-import {Status} from 'google-gax';
 
 /*!
  * The maximum number of writes that can be in a single batch.
@@ -193,9 +192,6 @@ class BulkCommitBatch {
       if (result.writeTime) {
         return new WriteResult(result.writeTime);
       } else {
-        if (result.status.code === Status.ABORTED) {
-          console.error('STATUS' + result.status.note);
-        }
         throw result.status;
       }
     });
@@ -628,6 +624,7 @@ export class BulkWriter {
       if (delayMs === 0) {
         this.sendBatch(batch);
       } else {
+        console.warn('throttling');
         delayExecution(() => this.sendReadyBatches(), delayMs);
         break;
       }
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -28,7 +28,7 @@ import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
 import {Deferred, getRetryCodes, wrapError} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
-import {GoogleError, Status} from 'google-gax';
+import {Status} from 'google-gax';
 import {logger} from './logger';
 
 /*!
@@ -96,7 +96,7 @@ class BulkCommitBatch {
 
   constructor(
     private readonly firestore: Firestore,
-    private readonly writeBatch: WriteBatch,
+    private writeBatch: WriteBatch,
     private readonly maxBatchSize: number
   ) {
     this.backoff = new ExponentialBackoff();
@@ -231,8 +231,8 @@ class BulkCommitBatch {
     let retryAttempts = 0;
     while (retryAttempts < MAX_RETRY_ATTEMPTS) {
       let retryIndexes: number[] = [];
+      await this.backoff.backoffAndWait();
       try {
-        await this.backoff.backoffAndWait();
         const results = await this.writeBatch.bulkCommit();
         retryIndexes = this.processResults(originalIndexMap, results);
       } catch (err) {
@@ -259,8 +259,7 @@ class BulkCommitBatch {
             `${retryIndexes.length}.`
         );
 
-        this.writeBatch._sliceIndexes(retryIndexes);
-        this.writeBatch._markNotCommitted();
+        this.writeBatch = this.writeBatch._sliceIndexes(retryIndexes);
       } else {
         break;
       }
@@ -280,32 +279,28 @@ class BulkCommitBatch {
     results: BatchWriteResult[],
     error?: Error
   ): number[] {
-    let indexesToRetry: number[] = [];
-    if (error === undefined) {
-      for (let i = 0; i < results.length; i++) {
-        const writeResult = results[i];
-        if (this.shouldRetry(writeResult.status.code)) {
-          indexesToRetry.push(i);
-        } else {
-          const originalIndex = originalIndexMap.get(i)!;
-          this.resultsMap.get(originalIndex)!.resolve(results[i]);
-        }
-      }
-    } else {
-      // If the batchWrite RPC fails with a retryable error, retry all writes
-      // in the batch.
-      if (error instanceof GoogleError && this.shouldRetry(error.code)) {
-        indexesToRetry = Array.from(
-          new Array(originalIndexMap.size),
-          (_, i) => i
-        );
+    const indexesToRetry: number[] = [];
+    if (error) {
+      results = Array.from({length: originalIndexMap.size}, () => {
+        return {
+          writeTime: null,
+          status: error,
+        };
+      });
+    }
+
+    for (let i = 0; i < results.length; i++) {
+      const writeResult = results[i];
+      const originalIndex = originalIndexMap.get(i)!;
+      if (writeResult.status.code === Status.OK) {
+        this.resultsMap.get(originalIndex)!.resolve(results[i]);
+      } else if (this.shouldRetry(writeResult.status.code)) {
+        indexesToRetry.push(i);
       } else {
-        for (let i = 0; i < this.opCount; i++) {
-          const originalIndex = originalIndexMap.get(i)!;
-          this.resultsMap.get(originalIndex)!.reject(error);
-        }
+        this.resultsMap.get(originalIndex)!.reject(writeResult.status);
       }
     }
+
     if (indexesToRetry.length === 0) {
       this.completedDeferred.resolve();
     }
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -24,6 +24,7 @@ import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
 import {Deferred, wrapError} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
+import {Status} from 'google-gax';
 
 /*!
  * The maximum number of writes that can be in a single batch.
@@ -192,6 +193,9 @@ class BulkCommitBatch {
       if (result.writeTime) {
         return new WriteResult(result.writeTime);
       } else {
+        if (result.status.code === Status.ABORTED) {
+          console.error('STATUS' + result.status.note);
+        }
         throw result.status;
       }
     });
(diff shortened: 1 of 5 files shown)


dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
       api.BatchWriteResponse
     >('batchWrite', request, tag, retryCodes);
 
-    return (response.writeResults || []).map((result, i) => {
+    let retryCount = 0;
+    while (retryCount < MAX_BATCH_WRITE_RETRY_ATTEMPTS) {
+      // Find the indexes of all writes that failed with ABORTED.
+      const abortedIndexes = response.status.reduce(
```
schmidt-sebastian, line 592: Let's make this generic and base our retries on the behavior defined in https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/v1/firestore_client_config.json

This is a multi-step process though: First, we need to modify the `firestore_client_config.json` in Google3. We likely want to retry writes on both UNAVAILABLE and ABORTED. We can then use https://github.com/googleapis/nodejs-firestore/blob/f175236bde2f64365f140b14641f848bd4eb34d9/dev/src/util.ts#L165 to extract the retry codes.

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
...
+        (arr: number[], status, i) => {
+          if (status.code === Status.ABORTED) {
+            arr.push(i);
```
schmidt-sebastian, line 595: Ideally, we would combine this with loop below. This might work:

```
const writesToRetry  = response.status.reduce((writesToRetry, status, i) => {
    if (shouldRetry(status)) {
       writesToRetry.push(writes[i]);
    }
   return writesToRetry
}, []);
```

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
...
+      }
+      logger(
+        'WriteBatch.bulkCommit',
+        tag,
+        'Current batch failed at retry #' +
+          retryCount +
+          '. Num failures: ' +
+          abortedIndexes.length +
+          '/' +
+          response.status.length
```
schmidt-sebastian, line 613: Nit: Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
       api.BatchWriteResponse
     >('batchWrite', request, tag, retryCodes);
 
-    return (response.writeResults || []).map((result, i) => {
+    let retryCount = 0;
```
schmidt-sebastian, line 589: This method is getting very long. Can we factor out the retry logic?

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
...
+      retryRequest.writes = retryRequest.writes?.filter((write, i) => {
+        return abortedIndexes.includes(i);
+      });
+      const retriedResponse = await this._firestore.request<
+        api.IBatchWriteRequest,
+        api.BatchWriteResponse
+      >('batchWrite', retryRequest, tag, retryCodes);
+
+      // Map the results of the retried request back to the original response.
+      for (let i = 0; i < abortedIndexes.length; i++) {
```
schmidt-sebastian, line 627: I was hoping we could just extract a new BulkCommitBatch and send it off using the existing `bulkCommit()`. Is that feasible?

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
...
+        const originalIndex = abortedIndexes[i];
+        response.writeResults[originalIndex] = retriedResponse.writeResults[i];
+        response.status[originalIndex] = retriedResponse.status[i];
+      }
+      await this._backoff.backoffAndWait();
```
schmidt-sebastian, line 632: The first `backoffAndWait()` doesn't actually do any backoff. Since you call it only after you have already send the first retry this essentially means that only your third attempt will be executed with backoff.

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
       api.BatchWriteResponse
     >('batchWrite', request, tag, retryCodes);
 
-    return (response.writeResults || []).map((result, i) => {
+    let retryCount = 0;
+    while (retryCount < MAX_BATCH_WRITE_RETRY_ATTEMPTS) {
```
schmidt-sebastian, line 590: `retryCount` is pretty ambiguous and doesn't clarify if the first retry loop is part of the count. I usually go with "attempt" to clarify that it is.

dev/src/write-batch.ts
```diff
@@ -577,7 +586,54 @@ export class WriteBatch implements firestore.WriteBatch {
...
+      for (let i = 0; i < abortedIndexes.length; i++) {
+        const originalIndex = abortedIndexes[i];
+        response.writeResults[originalIndex] = retriedResponse.writeResults[i];
+        response.status[originalIndex] = retriedResponse.status[i];
+      }
+      await this._backoff.backoffAndWait();
+      retryCount++;
+    }
+
+    return response.writeResults.map((result, i) => {
```
schmidt-sebastian, line 636: This code now seems to delay all write results for this batch even if just a single write failed. Can we find a way where the rest of the writes are returned as successful right away?

dev/test/bulk-writer.ts
```diff
@@ -546,9 +545,46 @@ describe('BulkWriter', () => {
...
+        },
+        {
+          request: createRequest([setOp('doc3', 'bar')]),
+          response: mergeResponses([successResponse(3)]),
+        },
+      ]);
+
+      bulkWriter.set(firestore.doc('collectionId/doc1'), {foo: 'bar'});
+      bulkWriter.set(firestore.doc('collectionId/doc2'), {foo: 'bar'});
+      bulkWriter.set(firestore.doc('collectionId/doc3'), {foo: 'bar'});
```
schmidt-sebastian, line 584: Don't you need to verify that these Promises are resolved successfully?

dev/src/v1/firestore_client_config.json
```diff
@@ -11,6 +11,10 @@
           "DEADLINE_EXCEEDED",
           "INTERNAL",
           "UNAVAILABLE"
+        ],
```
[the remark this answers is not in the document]
thebrianchen: done.

thebrianchen:
Made a pass at moving the retry handling to BulkWriter. One part of complexity that I wish I could remove is mapping the subsequent batch retries back to the original index on resultsMap.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +224,97 @@ class BulkCommitBatch {
...
+          const originalIndex = originalIndexMap.get(i)!;
+          this.resultsMap.get(originalIndex)!.reject(error);
+        }
       }
     }
-    this.completedDeferred.resolve();
+    if (indexesToRetry.length === 0) {
+      this.completedDeferred.resolve();
+    }
+    return indexesToRetry;
```
schmidt-sebastian, line 312: Can we combine the different error handling code paths here?

Something like:


```
    if (error) {
      results = Array.from({length:this.opCount}, () => { status: { code: error.code }})
    } 
    
    for (let i = 0; i < results.length; i++) {
      const writeResult = results[i];
      if (ok) {
        const originalIndex = originalIndexMap.get(i)!;
        this.resultsMap.get(originalIndex)!.resolve(results[i]);
      } else if (this.shouldRetry(writeResult.status.code)
        indexesToRetry.push(i);
      } else {
         reject;
      }
    }
    
    if (indexesToRetry.length === 0) {
      this.completedDeferred.resolve();
    }
    
    return indexesToRetry;
```

dev/src/bulk-writer.ts
```diff
@@ -211,25 +224,97 @@ class BulkCommitBatch {
...
+      Array.from(new Array(this.opCount), (_, i) => [i, i])
+    );
+
+    let retryAttempts = 0;
+    while (retryAttempts < MAX_RETRY_ATTEMPTS) {
+      let retryIndexes: number[] = [];
+      try {
+        await this.backoff.backoffAndWait();
+        const results = await this.writeBatch.bulkCommit();
+        retryIndexes = this.processResults(originalIndexMap, results);
```
schmidt-sebastian, line 237: Can we rewrite this so that `processResults` doesn't take `originalIndexMap`? It seems like we could apply the  `retryIndexes` indexes in this function. I haven't played with this, but I wonder if this would lead to a simplification.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +224,97 @@ class BulkCommitBatch {
...
-    });
+    let originalIndexMap: Map<number, number> = new Map(
+      Array.from(new Array(this.opCount), (_, i) => [i, i])
+    );
+
+    let retryAttempts = 0;
+    while (retryAttempts < MAX_RETRY_ATTEMPTS) {
+      let retryIndexes: number[] = [];
+      try {
+        await this.backoff.backoffAndWait();
```
schmidt-sebastian, line 235: Nit: Move this out of the try/catch block.

dev/src/v1/firestore_client_config.json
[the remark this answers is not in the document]
schmidt-sebastian: It still seems missing?

dev/src/write-batch.ts
```diff
@@ -594,6 +594,29 @@ export class WriteBatch implements firestore.WriteBatch {
...
   }
 
+  /**
+   * Removes all operations from the WriteBatch except for the ones located
+   * at the provided indexes.
+   *
+   * @param indexes List of operation indexes to keep
+   * @private
+   */
+  _sliceIndexes(indexes: number[]): void {
```
schmidt-sebastian, line 604: The Array `slice()` method returns a copy, which I think would be preferably here as well.  Can we update this to create a new WriteBatch that is marked '_committed=false`? The reason I am asking is that I would ideally not expose a method that mocks with the internals of a WriteBatch, as the potential for future misuse is high.

dev/test/bulk-writer.ts
```diff
@@ -601,6 +651,46 @@ describe('BulkWriter', () => {
     });
   });
 
+  it('retries retryable batchWrite failures', async () => {
+    setTimeoutHandler(setImmediate);
+    let retryCount = 0;
```
schmidt-sebastian, line 656: s/retry/attempt

dev/test/bulk-writer.ts
```diff
@@ -601,6 +651,46 @@ describe('BulkWriter', () => {
     });
   });
 
+  it('retries retryable batchWrite failures', async () => {
```
schmidt-sebastian, line 654: Is it possible to make these two tests more distinguishable?

Something like:

`retries batchWrite when single operation fails` vs `retries batchWrite when entire RPC fails`.

Furthermore, should we test what happens when we have both non-rertyable and retryable writes?

dev/src/bulk-writer.ts
thebrianchen: It's possible to not pass in `originalIndexMap` to `processResults` -- the successful results are located at the indexes not in `retryIndexes`. From there we could resolve/reject the individual promises and reject the rest.

However, the main issue is that the `resultsMaps` resolution logic would have to live in `bulkCommit()`, which then means that `processResults()` would become `findRetryableIndexes()`. From there, it would make more sense to have all the logic in `bulkCommit`, which would make the method much longer. 

With the new refactor, using `originalIndexMap` in `processResults` seems a lot cleaner. WDYT?

dev/src/v1/firestore_client_config.json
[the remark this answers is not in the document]
thebrianchen: update: in cl/318124350.

thebrianchen:
Moved to #1243 in order to merge against master.

PR #1228, opened 2020-06-18
fix: another attempt at fixing the flaky BulkWriter test

thebrianchen:


dev/test/bulk-writer.ts
```diff
@@ -549,55 +553,63 @@ describe('BulkWriter', () => {
...
+        for (let i = 0; i < 499; i++) {
+          bulkWriter.set(firestore.doc('collectionId/doc' + i), {foo: 'bar'});
+        }
+        bulkWriter.flush();
+
+        // Wait until the initial batch has been sent by BulkWriter. Because
+        // there are floating promises, not waiting here can cause a race
+        // condition where the first batch's request has not been made by the
+        // time the second request is sent.
+        await requestMadeDeferred.promise;
```
schmidt-sebastian, line 589: I am a little confused here... isn't the test meant to send off a lot of writes as fast as possible, and as such it would be beneficial not to wait?

It seems like what we should do is:

```
setTimeoutHandler((_, timeout) => {
  if (!timeoutCalled) {
    timeoutCalled = true;
    // expect(requestCounter).to.equal(0); This is racy without your 
   //  deferred promise, but not needed for the purpose of the test
    expect(timeout).to.be.greaterThan(0);
    // done(); Maybe remove this so we don't have any outstanding 
    //  work when the test completes.
  }
});

for (let i = 0; i < 600; i++) {
   bulkWriter.set(firestore.doc('collectionId/doc' + i), {foo: 'bar'});
}

await bulkWriter.close();
expect(timeoutCalled).to.be.true;

```

thebrianchen: Post conversation:
The original issue with this simplified test was that it still passes even if the timeout handler is called for the first batch. To fix this, the BulkWriter tests now check that the timeout handler wasn't called after each individual test.

I tried removing the `done()`, but since it's based on the actual clock, the timeout handler will be called multiple times. If we just call the timeout function directly in the overridden handler, we will loop until enough time passes for the rate limiter to allow the request. Though there could still be outstanding work once the test completes, it shouldn't affect other tests since each test creates a new BulkWriter instance.

thebrianchen:
I also removed the response generation logic since the rate limiter test isn't testing the request/response logic. This also allows the test to pass independently of what the maximum batch size is.

dev/test/bulk-writer.ts
```diff
@@ -547,57 +561,47 @@ describe('BulkWriter', () => {
...
-        // Check that BulkWriter has not yet sent the 2nd batch.
-        timeoutCalled = true;
-        expect(requestCounter).to.equal(0);
-        expect(timeout).to.be.greaterThan(0);
+        for (let i = 0; i < 600; i++) {
+          bulkWriter.set(firestore.doc('collectionId/doc' + i), {foo: 'bar'});
+        }
+        await bulkWriter.close();
+        if (!timeoutCalled) {
+          done(new Error('Expected the timeout handler to be called'));
```
schmidt-sebastian, line 602: It's actually an error to invoke done() twice. This should show up in the console.

dev/test/bulk-writer.ts
```diff
@@ -547,57 +561,47 @@ describe('BulkWriter', () => {
...
-      let timeoutCalled = false;
-      setTimeoutHandler((_, timeout) => {
-        // Check that BulkWriter has not yet sent the 2nd batch.
-        timeoutCalled = true;
-        expect(requestCounter).to.equal(0);
-        expect(timeout).to.be.greaterThan(0);
+        for (let i = 0; i < 600; i++) {
+          bulkWriter.set(firestore.doc('collectionId/doc' + i), {foo: 'bar'});
+        }
+        await bulkWriter.close();
```
schmidt-sebastian, line 600: If you don't wait for the close to finish (i.e. no dangling promise in this test), then `verifyInstance` (line 189) might become flaky since it verifies that at the end of each test all operations have finished executing.

dev/test/bulk-writer.ts
```diff
@@ -547,57 +561,47 @@ describe('BulkWriter', () => {
...
+        firestore = firestoreClient;
+        return firestore._bulkWriter();
+      });
+    }
 
-    // TODO(chenbrian): Actually fix this test...
-    it.skip('does not send batches if doing so exceeds the rate limit', async () => {
-      // The test is considered a success if BulkWriter tries to send the second
-      // batch again after a timeout.
+    afterEach(() => setTimeoutHandler(setTimeout));
```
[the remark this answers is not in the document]
thebrianchen: removed.

dev/test/bulk-writer.ts
thebrianchen: Either the `done()` in the timeout handler will be called, setting `timeoutCalled` to `true`, or the error `done()` will be called. Calling `done()` twice will cause the test to fail.

dev/test/bulk-writer.ts
thebrianchen: talked separately: The 2nd request will never fire since we are not invoking the callback function in the overridden handler. This means that the the handler will be trigger twice: once when the 2nd request is made, and once when the first request is returned.

dev/test/bulk-writer.ts
```diff
@@ -547,57 +558,46 @@ describe('BulkWriter', () => {
...
+      });
+    }
 
-      const arrayRange = Array.from(new Array(500), (_, i) => i);
-      const requests1 = arrayRange.map(i => setOp('doc' + i, 'bar'));
-      const responses1 = arrayRange.map(i => successResponse(i));
-      const arrayRange2 = [500, 501, 502, 503, 504];
-      const requests2 = arrayRange2.map(i => setOp('doc' + i, 'bar'));
-      const responses2 = arrayRange2.map(i => successResponse(i));
+    it('does not send batches if doing so exceeds the rate limit', done => {
```
schmidt-sebastian, line 581: Final nit: This test should get a more obvious name "Uses timeout for batches that exceed the rate limit" for example.

PR #1229, opened 2020-06-18
chore: add eslint check for console.log statements

thebrianchen:
I allowed `console.error` because we pass in `console.error` as the `onError` handler in `Watch.onSnapshot`. 

Should I add this rule to the tests directory as well? If so, should I just c/p the rule, or make another overrides object that includes both `/src` and `/test`?


schmidt-sebastian:
I have no opinion on whether to add this to our test code.

PR #1232, opened 2020-06-21
fix: Extended support for generics on some firestore methods.

### LINKED ISSUE #1231: Extending existing Generic support.
The Transaction class found in https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L309, allows the user to provide a type to describe a Document. But in some other methods, we do not have this ability, ex: https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L190 and https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L199.

To make things consistent and greatly improve compile time type safety and autocompletion (even if firestore does not provide type safety as noted in #109 ), it would be good to extends the currently existing Generic support. The use case of no compile time type safety should still be supported for those that do not want to use it.

The following use case is not supported:
```ts
interface Doc {
  name: string;
}

var batch = admin.firestore().batch();

// Set the value of 'NYC'
var nycRef = admin.firestore().collection("cities").doc("NYC");
batch.set<Doc>(nycRef, {name: "New York City"});
```
This fails with the following error:
```
Argument of type 'DocumentReference<DocumentData>' is not assignable to parameter of type 'DocumentReference<Doc>'.
```

This should be possible:
```ts
interface Doc {
  name: string;
}

var batch = admin.firestore().batch();

// Set the value of 'NYC'
var nycRef = admin.firestore().collection<Doc>("cities").doc("NYC");
// or: var nycRef = admin.firestore().collection("cities").doc<Doc>("NYC");
batch.set<Doc>(nycRef, {name: "New York City"});
```

Ranguna:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-firestore/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Code coverage does not decrease (only the types were changed, this should be ok)
- [X] Appropriate docs were updated (not necessary)

As for tests: `npm test` and `lint` both work but I can't run `npm run samples-test` because `gcloud` installation fails with the following:
```
node-pre-gyp http 403 https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v0.14.1/node-v64-linux-x64.tar.gz
node-pre-gyp ERR! Tried to download: https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v0.14.1/node-v64-linux-x64.tar.gz 
node-pre-gyp ERR! Pre-built binaries not found for grpc@0.14.1 and node@10.17.0 (node-v64 ABI) (falling back to source compile with node-gyp) 
node-pre-gyp http 403 status code downloading tarball https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v0.14.1/node-v64-linux-x64.tar.gz 
n
```
So I can't run `npm run samples-test` and `npm run system-test`.

Fixes #1231 🦕

### DIFF
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -32,7 +32,7 @@ declare namespace FirebaseFirestore {
    * paths (e.g. 'foo' or 'foo.baz') mapped to values. Fields that contain dots
    * reference nested fields within the document.
    */
-  export type UpdateData = {[fieldPath: string]: any};
+  export type UpdateData<T = DocumentData> = {[fieldPath in keyof T]: T[fieldPath]};
 
   /**
    * Sets or disables the log function for all active Firestore instances.
@@ -187,7 +187,7 @@ declare namespace FirebaseFirestore {
      * @param collectionPath A slash-separated path to a collection.
      * @return The `CollectionReference` instance.
      */
-    collection<T>(collectionPath: string): CollectionReference<T>;
+    collection<T = DocumentData>(collectionPath: string): CollectionReference<T>;
 
     /**
      * Gets a `DocumentReference` instance that refers to the document at the
@@ -196,7 +196,7 @@ declare namespace FirebaseFirestore {
      * @param documentPath A slash-separated path to a document.
      * @return The `DocumentReference` instance.
      */
-    doc<T>(documentPath: string): DocumentReference<T>;
+    doc<T = DocumentData>(documentPath: string): DocumentReference<T>;
 
     /**
      * Creates and returns a new Query that includes all documents in the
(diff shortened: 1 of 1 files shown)


Ranguna:
I believe the `conventionalcommits.org` action is not passing because I forced push the commit message change to my branch and the action probably got confused. Should I redo this PR ?

schmidt-sebastian:
As pointed out in https://github.com/googleapis/nodejs-firestore/issues/1231, this is not something we can take. I am sorry about this!

PR #1235, opened 2020-06-24
chore: run synthtool after node10 merge

schmidt-sebastian:


schmidt-sebastian:
Excluded the Kokoro conf changes to make the CI green.

PR #1243, opened 2020-06-25
fix: retry ABORTED writes in bulkCommit

thebrianchen:
Recreating #1222 to merge against PR without deleting comment history.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 import * as firestore from '@google-cloud/firestore';
+import {Status} from 'google-gax';
 
 import * as assert from 'assert';
 
@@ -28,7 +29,6 @@ import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
 import {Deferred, getRetryCodes, wrapError} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
-import {Status} from 'google-gax';
 import {logger} from './logger';
 
 /*!
@@ -280,6 +280,8 @@ class BulkCommitBatch {
     error?: Error
   ): number[] {
     const indexesToRetry: number[] = [];
+    // If the BatchWrite RPC failed, the results array will be empty. Create
+    // a results array containing the error for each write in the request.
     if (error) {
       results = Array.from({length: originalIndexMap.size}, () => {
         return {
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -229,11 +229,11 @@ class BulkCommitBatch {
         results = await this.writeBatch.bulkCommit();
       } catch (err) {
         // Map the failure to each individual write's result.
-        results = [...this.pendingOps.keys()].map(key => {
-          return {key, writeTime: null, status: wrapError(err, stack)};
+        results = [...this.pendingOps.keys()].map(path => {
+          return {key: path, writeTime: null, status: wrapError(err, stack)};
         });
       }
-      this.processResults(results, /* shouldRetry= */ true);
+      this.processResults(results);
 
       if (this.pendingOps.size > 0) {
         logger(
@@ -252,16 +252,16 @@ class BulkCommitBatch {
       }
     }
 
-    this.processResults(results, /* shouldRetry= */ false);
+    this.failRemainingOperations(results);
     this.completedDeferred.resolve();
   }
 
   /**
    * Resolves the individual operations in the batch with the results.
    */
-  processResults(results: BatchWriteResult[], shouldRetry: boolean): void {
+  private processResults(results: BatchWriteResult[]): void {
     for (const result of results) {
-      if (result.status.code === Status.OK || !shouldRetry) {
+      if (result.status.code === Status.OK) {
         this.pendingOps.get(result.key)!.resolve(result);
         this.pendingOps.delete(result.key);
       } else if (!this.shouldRetry(result.status.code)) {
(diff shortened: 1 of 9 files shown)


dev/src/bulk-writer.ts
```diff
@@ -211,25 +224,92 @@ class BulkCommitBatch {
...
+    );
+
+    let retryAttempts = 0;
+    while (retryAttempts < MAX_RETRY_ATTEMPTS) {
+      let retryIndexes: number[] = [];
+      await this.backoff.backoffAndWait();
+      try {
+        const results = await this.writeBatch.bulkCommit();
+        retryIndexes = this.processResults(originalIndexMap, results);
+      } catch (err) {
```
schmidt-sebastian, line 238: I tried to see if I could get rid of this re-assignment and got completely carried away. Here is a cleanup:

https://gist.github.com/schmidt-sebastian/702ba33ce0c3515d096089226a762934

Use at your own risk and take at your own consideration :)

dev/src/bulk-writer.ts
```diff
@@ -211,25 +224,92 @@ class BulkCommitBatch {
...
+      } else if (this.shouldRetry(writeResult.status.code)) {
+        indexesToRetry.push(i);
+      } else {
+        this.resultsMap.get(originalIndex)!.reject(writeResult.status);
       }
     }
-    this.completedDeferred.resolve();
+
+    if (indexesToRetry.length === 0) {
+      this.completedDeferred.resolve();
```
schmidt-sebastian, line 305: We also need to resolve this when all retry attempts failed. Please also add a test for this.

dev/src/bulk-writer.ts
thebrianchen: Huge fan of this organization. I implemented what you did, but changed `keys` to `docPaths` since I was a bit confused about what the keys were at first. I think this approach will work with the UpdateBuilder pattern on java as well.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +221,70 @@ class BulkCommitBatch {
...
-    });
+    let results: BatchWriteResult[] = [];
+    for (let attempt = 0; attempt < MAX_RETRY_ATTEMPTS; attempt++) {
+      await this.backoff.backoffAndWait();
+
+      try {
+        results = await this.writeBatch.bulkCommit();
+      } catch (err) {
+        // Map the failure to each individual write's result.
+        results = [...this.pendingOps.keys()].map(key => {
```
schmidt-sebastian, line 232: Should "key" be "path" then?

dev/src/bulk-writer.ts
```diff
@@ -211,25 +221,70 @@ class BulkCommitBatch {
...
    */
-  processResults(results: BatchWriteResult[], error?: Error): void {
-    if (error === undefined) {
-      for (let i = 0; i < this.opCount; i++) {
-        this.resultsMap.get(i)!.resolve(results[i]);
-      }
-    } else {
-      for (let i = 0; i < this.opCount; i++) {
-        this.resultsMap.get(i)!.reject(error);
+  processResults(results: BatchWriteResult[], shouldRetry: boolean): void {
```
schmidt-sebastian, line 262: s/shouldRetry/mayRetry (or allowRetry)?

Also might be worth adding `@param` comments.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +221,70 @@ class BulkCommitBatch {
...
+        this.writeBatch = new WriteBatch(this.firestore, this.writeBatch, [
+          ...this.pendingOps.keys(),
+        ]);
+      } else {
+        this.completedDeferred.resolve();
+        return;
+      }
+    }
+
+    this.processResults(results, /* shouldRetry= */ false);
```
schmidt-sebastian, line 255: This might be cleaner if this was `this.failRemainingOperations(results)` and a separate call.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +221,70 @@ class BulkCommitBatch {
...
+          `Current batch failed at retry #${attempt}. Num failures: ` +
+            `${this.pendingOps.size}.`
+        );
+
+        this.writeBatch = new WriteBatch(this.firestore, this.writeBatch, [
+          ...this.pendingOps.keys(),
+        ]);
+      } else {
+        this.completedDeferred.resolve();
+        return;
```
schmidt-sebastian, line 251: This could just be a break,  this.processResult() would be a no-op.

dev/src/bulk-writer.ts
```diff
@@ -211,25 +221,70 @@ class BulkCommitBatch {
...
     }
-    this.completedDeferred.resolve();
+  }
+
+  private shouldRetry(code: Status | undefined): boolean {
+    const retryCodes = getRetryCodes('batchWrite');
+    return code !== undefined && retryCodes.includes(code);
+  }
+
+  has(path: string): boolean {
```
schmidt-sebastian, line 279: Let's use a more descriptive name here (hasDoc(), hasWrite(), ...).

dev/src/write-batch.ts
```diff
@@ -141,10 +144,26 @@ export class WriteBatch implements firestore.WriteBatch {
...
-  constructor(firestore: Firestore) {
+  constructor(
+    firestore: Firestore,
+    retryBatch: WriteBatch,
+    keysToRetry: string[]
+  );
+  constructor(firestore: Firestore);
+  constructor(
+    firestore: Firestore,
+    retryBatch?: WriteBatch,
```
schmidt-sebastian, line 155: Please document.

dev/src/write-batch.ts
```diff
@@ -141,10 +144,26 @@ export class WriteBatch implements firestore.WriteBatch {
...
+    docsToRetry?: string[]
```
schmidt-sebastian, line 156: Please document.

dev/src/bulk-writer.ts
thebrianchen: The map entry is deleted from `pendingOps` after the promise is resolved in `processResults`, so an error would be thrown if `processResults` is called again.

dev/src/bulk-writer.ts
thebrianchen: I originally didn't want to copy/paste code, but in this case, it clarifies the behavior much better.

dev/src/bulk-writer.ts
thebrianchen: N/A since we're using `failRemainingOperations`.

dev/src/bulk-writer.ts

dev/src/write-batch.ts
thebrianchen: done.

dev/src/write-batch.ts
thebrianchen: done.

PR #1251, opened 2020-06-30
feat: add deprecation message to WriteBatch

thebrianchen:


dev/src/index.ts
```diff
@@ -659,6 +659,9 @@ export class Firestore implements firestore.Firestore {
    * @returns {WriteBatch} A WriteBatch that operates on this Firestore
    * client.
    *
+   * @deprecated This class is now deprecated. Use Transaction for atomic,
```
schmidt-sebastian, line 662: "Atomic and parallel" seems a little incorrect.

How about: Use `runTransaction()` to update multiple documents atomically or `bulkWriter()` to update a large number of document in parallel.


We should add this in four places:
- WriteBatch in src/
- batch() in src/
- WriteBatch in types/
- batch() in types/

thebrianchen: done

dev/src/index.ts
```diff
@@ -659,6 +659,10 @@ export class Firestore implements firestore.Firestore {
    * @returns {WriteBatch} A WriteBatch that operates on this Firestore
    * client.
    *
+   * @deprecated This class is now deprecated. Use `runTransaction()` to update
+   * multiple documents atomically or `bulkWriter()` to update a large number
+   * of document in parallel.
```
schmidt-sebastian, line 664: s/document/documents

PR #1260, opened 2020-07-10
build: use bazel build in synthtool

alexander-fenster:
This PR moves Firestore client library generation from Docker to Bazel. 

Note: the `synthtool` execution may be slow if you run it for the first time (it will rebuild protobuf sources). The subsequent runs are fast.

I re-ran `synthtool` to make sure everything works. Not sure what is the reason of jsdoc comment change for enums though.


schmidt-sebastian (approved):
LGTM. The type change in the JSDoc comments don't seem to affect the `.d.ts` files and as such they don't matter much. It would be nice if we knew why this happened, but I am not overly concerned.

PR #1275, opened 2020-07-27
feat: allow `Settings.host` to be used when `Settings.servicePath` is set

### LINKED ISSUE #1265: [Firestore] firestore.settings() cannot have 'host' when FIRESTORE_EMULATOR_HOST is set
#### Environment details

  - OS: macOS
  - Node.js version: 10.x
  - npm version: 6.14.5
  - `@google-cloud/firestore` version: 4.1.1 (via `firebase-admin` 9.0.0)

#### Steps to reproduce

index.js
```js
const admin = require('firebase-admin');

admin.initializeApp({
  projectId: 'abc-123'
});

admin.firestore().settings({
  host: 'localhost:8080',
  ssl: false
});

console.log("Ok!");
```

```bash
$ node index.js 
Ok!
```

```bash
$ FIRESTORE_EMULATOR_HOST=localhost:8080 node index.js 
/private/var/folders/xl/6lkrzp7j07581mw8_4dlt3b000643s/T/tmp.4ukcSYBP/node_modules/@google-cloud/firestore/build/src/index.js:424
                throw new Error('Cannot set both "settings.host" and "settings.servicePath".');
                ^

Error: Cannot set both "settings.host" and "settings.servicePath".
    at Firestore.validateAndApplySettings (/private/var/folders/xl/6lkrzp7j07581mw8_4dlt3b000643s/T/tmp.4ukcSYBP/node_modules/@google-cloud/firestore/build/src/index.js:424:23)
    at Firestore.settings (/private/var/folders/xl/6lkrzp7j07581mw8_4dlt3b000643s/T/tmp.4ukcSYBP/node_modules/@google-cloud/firestore/build/src/index.js:413:14)
    at Object.<anonymous> (/private/var/folders/xl/6lkrzp7j07581mw8_4dlt3b000643s/T/tmp.4ukcSYBP/index.js:7:19)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
```

`settings.host` should **override** the environment variable, and it definitely shouldn't throw if the values are the same in any case.

thebrianchen:
Fixes #1265.

The existing behavior prioritized `FIRESTORE_EMULATOR_HOST` over any user-specified host when set, and prohibits setting the host more than once. This PR allows `Settings.host` to take precedence over `Settings.servicePath` and `Settings.apiEndpoint` if `FIRESTORE_EMULATOR_HOST` is not set and logs a warning when the host is overridden. Otherwise, `FIRESTORE_EMULATOR_HOST` is used as the host, regardless of any values that are passed via `Settings`.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -408,21 +408,7 @@ export class Firestore implements firestore.Firestore {
       libraryHeader.libVersion += ' fire/' + settings.firebaseVersion;
     }
 
-    if (process.env.FIRESTORE_EMULATOR_HOST) {
-      validateHost(
-        'FIRESTORE_EMULATOR_HOST',
-        process.env.FIRESTORE_EMULATOR_HOST
-      );
-
-      settings = {
-        ...settings,
-        host: process.env.FIRESTORE_EMULATOR_HOST,
-        ssl: false,
-      };
-    } else {
-      settings = {...settings, ...libraryHeader};
-    }
-    this.validateAndApplySettings(settings);
+    this.validateAndApplySettings({...settings, ...libraryHeader});
 
     const retryConfig = serviceConfig.retry_params.default;
     this._backoffSettings = {
@@ -496,7 +482,22 @@ export class Firestore implements firestore.Firestore {
     }
 
     let url: URL | null = null;
-    if (settings.host !== undefined) {
+
+    // If the environment variable is set, it should always take precedence
+    // over any user passed in settings.
+    if (process.env.FIRESTORE_EMULATOR_HOST) {
+      validateHost(
+        'FIRESTORE_EMULATOR_HOST',
+        process.env.FIRESTORE_EMULATOR_HOST
+      );
+
+      settings = {
+        ...settings,
+        host: process.env.FIRESTORE_EMULATOR_HOST,
+        ssl: false,
+      };
+      url = new URL(`http://${settings.host}`);
+    } else if (settings.host !== undefined) {
       validateHost('settings.host', settings.host);
       url = new URL(`http://${settings.host}`);
       if (
@@ -507,7 +508,7 @@ export class Firestore implements firestore.Firestore {
       ) {
         // eslint-disable-next-line no-console
         console.warn(
-          `The provided host (${settings.host}) in "settings" does not ` +
+          `The provided host (${url.hostname}) in "settings" does not ` +
             `match the existing host (${
               settings.servicePath ?? settings.apiEndpoint
             }). Using the provided host.`
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -504,14 +504,24 @@ export class Firestore implements firestore.Firestore {
 
     if (settings.host !== undefined) {
       validateHost('settings.host', settings.host);
-      if (settings.servicePath !== undefined) {
-        throw new Error(
-          'Cannot set both "settings.host" and "settings.servicePath".'
+      if (
+        settings.servicePath !== undefined &&
+        settings.servicePath !== settings.host
+      ) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          '"settings.servicePath" does not match "settings.host". ' +
+            'Using "settings.host" as host.'
         );
       }
-      if (settings.apiEndpoint !== undefined) {
-        throw new Error(
-          'Cannot set both "settings.host" and "settings.apiEndpoint".'
+      if (
+        settings.apiEndpoint !== undefined &&
+        settings.apiEndpoint !== settings.host
+      ) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          '"settings.apiEndpoint" does not match "settings.host". ' +
+            'Using "settings.host" as host.'
         );
       }
 
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -500,6 +500,10 @@ export class Firestore implements firestore.Firestore {
     } else if (settings.host !== undefined) {
       validateHost('settings.host', settings.host);
       url = new URL(`http://${settings.host}`);
+    }
+
+    // Only store the host if a valid value was provided in `host`.
+    if (url !== null) {
       if (
         (settings.servicePath !== undefined &&
           settings.servicePath !== url.hostname) ||
@@ -514,21 +518,18 @@ export class Firestore implements firestore.Firestore {
             }). Using the provided host.`
         );
       }
-    }
 
-    // Only store the host if a valid value was provided in `host`.
-    if (url !== null) {
       settings.servicePath = url.hostname;
       if (url.port !== '' && settings.port === undefined) {
         settings.port = Number(url.port);
       }
-    }
 
-    // We need to remove the `host` and `apiEndpoint` setting, in case a user
-    // calls `settings()`, which will compare the the provided `host` to the
-    // existing hostname stored on `servicePath`.
-    delete settings.host;
-    delete settings.apiEndpoint;
+      // We need to remove the `host` and `apiEndpoint` setting, in case a user
+      // calls `settings()`, which will compare the the provided `host` to the
+      // existing hostname stored on `servicePath`.
+      delete settings.host;
+      delete settings.apiEndpoint;
+    }
 
     if (settings.ssl !== undefined) {
       validateBoolean('settings.ssl', settings.ssl);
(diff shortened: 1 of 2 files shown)


dev/src/index.ts
```diff
@@ -504,14 +504,24 @@ export class Firestore implements firestore.Firestore {
...
     if (settings.host !== undefined) {
       validateHost('settings.host', settings.host);
-      if (settings.servicePath !== undefined) {
-        throw new Error(
-          'Cannot set both "settings.host" and "settings.servicePath".'
+      if (
+        settings.servicePath !== undefined &&
+        settings.servicePath !== settings.host
+      ) {
+        // eslint-disable-next-line no-console
```
thebrianchen, line 511: Not sure about the error message here, since the host is stored on `settings.servicePath`. Could we refer to hosts set on `settings.servicePath` and `settings.apiEndpoint` as "existing hosts"? That way, we could simply and combine the two error messages into something like "settings.host does not match the existing host..."

schmidt-sebastian:
@thebrianchen Did you look at why this failed? I don't see how this can fail:

https://github.com/googleapis/nodejs-firestore/blob/56355f1c83b055fdbd3ed47e069df813696a1fc0/dev/src/index.ts#L427
unsets the `servicePath` if FIRESTORE_EMULATOR_HOST is set. https://github.com/googleapis/nodejs-firestore/blob/56355f1c83b055fdbd3ed47e069df813696a1fc0/dev/src/index.ts#L496 doesn't seem to change it again.

dev/src/index.ts
```diff
@@ -505,23 +505,20 @@ export class Firestore implements firestore.Firestore {
...
-      ) {
-        // eslint-disable-next-line no-console
-        console.warn(
-          '"settings.apiEndpoint" does not match "settings.host". ' +
-            'Using "settings.host" as host.'
+          `The provided host (${settings.host}) in "settings" does not match the ` +
+            `existing host (${
+              settings.servicePath !== undefined
+                ? settings.servicePath
+                : settings.apiEndpoint
```
schmidt-sebastian, line 519:
```suggestion
              settings.servicePath ?? settings.apiEndpoint
```

dev/src/index.ts
```diff
@@ -505,23 +505,20 @@ export class Firestore implements firestore.Firestore {
     if (settings.host !== undefined) {
       validateHost('settings.host', settings.host);
       if (
-        settings.servicePath !== undefined &&
-        settings.servicePath !== settings.host
+        (settings.servicePath !== undefined &&
+          settings.servicePath !== settings.host) ||
+        (settings.apiEndpoint !== undefined &&
+          settings.apiEndpoint !== settings.host)
```
schmidt-sebastian, line 511: Host is "hostname:port". At least servicePath is just hostname. Can you verify that this works as expected?

dev/src/index.ts
```diff
@@ -530,10 +527,11 @@ export class Firestore implements firestore.Firestore {
...
         settings.port = Number(url.port);
       }
-      // We need to remove the `host` setting, in case a user calls `settings()`,
-      // which will again enforce that `host` and `servicePath` are not both
-      // specified.
+      // We need to remove the `host` and `apiEndpoint` setting, in case a user
+      // calls `settings()`, which will compare the the provided `host` to the
+      // existing host stored on `servicePath`.
       delete settings.host;
+      delete settings.apiEndpoint;
```
schmidt-sebastian, line 534: I think we can now delete the code that also does this in the constructor.

By the way, I am still a little uneasy about the order of operations here:

```
process.env.FIRESTORE_EMULATOR_HOST = 'chicago';
const firestore = new Firestore({host: 'sanfrancisco'});
// Now talking to 'chicago'

vs:

process.env.FIRESTORE_EMULATOR_HOST = 'chicago';
const firestore = new Firestore()
firstore.setting({host: 'sanfrancisco'});
// Now talking to 'sanfrancisco'
```

Maybe all the constructor code should be moved to `validateAndApplySettings`, where we can then use a unified and sane behavior?

dev/src/index.ts
thebrianchen: thanks, done.

dev/src/index.ts
thebrianchen: Changed logic to compare hostnames.

dev/src/index.ts
thebrianchen: moved the deletion code over to `validateAndApplySettings`. However, we need to keep the env var check in the constructor in order to preserve existing behavior where the env var overrides anything passed into the constructor.

dev/src/index.ts
```diff
@@ -502,30 +481,55 @@ export class Firestore implements firestore.Firestore {
...
-      // which will again enforce that `host` and `servicePath` are not both
-      // specified.
-      delete settings.host;
     }
 
+    // We need to remove the `host` and `apiEndpoint` setting, in case a user
+    // calls `settings()`, which will compare the the provided `host` to the
+    // existing hostname stored on `servicePath`.
+    delete settings.host;
+    delete settings.apiEndpoint;
```
schmidt-sebastian, line 531: This should be moved into the if-statement above, shouldn't it?

dev/src/index.ts
```diff
@@ -502,30 +481,55 @@ export class Firestore implements firestore.Firestore {
       this._projectId = settings.projectId;
     }
 
-    if (settings.host !== undefined) {
+    let url: URL | null = null;
+
+    // If the environment variable is set, it should always take precedence
+    // over any user passed in settings.
```
schmidt-sebastian, line 487: If we feel strongly that this is the wrong thing to do, should add a TODO to fix this?

dev/src/index.ts
```diff
@@ -502,30 +481,55 @@ export class Firestore implements firestore.Firestore {
...
-          'Cannot set both "settings.host" and "settings.apiEndpoint".'
+      url = new URL(`http://${settings.host}`);
+      if (
+        (settings.servicePath !== undefined &&
+          settings.servicePath !== url.hostname) ||
+        (settings.apiEndpoint !== undefined &&
+          settings.apiEndpoint !== url.hostname)
+      ) {
+        // eslint-disable-next-line no-console
+        console.warn(
```
schmidt-sebastian, line 510: I am debating if it makes sense to log this warning even when FIRESTORE_EMULATOR_HOST is used. Most of the time, this shouldn't warn since servicePath and apiEndpoint are rarely set.

dev/src/index.ts
```diff
@@ -502,30 +481,55 @@ export class Firestore implements firestore.Firestore {
...
       if (url.port !== '' && settings.port === undefined) {
         settings.port = Number(url.port);
       }
-      // We need to remove the `host` setting, in case a user calls `settings()`,
-      // which will again enforce that `host` and `servicePath` are not both
-      // specified.
-      delete settings.host;
     }
 
+    // We need to remove the `host` and `apiEndpoint` setting, in case a user
```
schmidt-sebastian, line 527: Comment no longer applicable.

dev/src/index.ts
thebrianchen: I don't see how we would fix it without creating an inconvenient breaking change for developers. Since this is only used for emulators and 1p testing, I think the current behavior isn't "wrong". I mainly wanted to document the priority order to make the code easier to understand.

dev/src/index.ts
thebrianchen: good catch, done.

dev/src/index.ts
thebrianchen: Yeah, I think that makes more sense. If anything, it will make it easier for developers to figure out what host is actually being used.

dev/src/index.ts
thebrianchen: Isn't this still applicable if the user does not provide the env var? In that case, creating a new Firestore with a host passed in and then calling `settings()` with `host` specified would cause this situation.

schmidt-sebastian (approved):
LGTM. Please update PR title to match new behavior.

JustinBeckwith (approved):

PR #1280, opened 2020-08-06
feat: Build query/document mapping in bundles.

wu-hui:


dev/protos/firestore/bundle.proto
```diff
@@ -79,6 +79,9 @@ message BundledDocumentMetadata {
 
   // Whether the document exists.
   bool exists = 3;
+
+  // The query name in this bundle, if the document is added via a named query.
+  string query = 4;
```
schmidt-sebastian, line 84: A document can match multiple queries and should be a list that is merged with each update, shouldn't it?

dev/test/bundle.ts
```diff
@@ -158,7 +168,14 @@ describe('Bundle Buidler', () => {
...
     // Verify named query
-    const namedQuery = (elements[1] as IBundleElement).namedQuery;
+    let namedQuery = (elements[1] as IBundleElement).namedQuery;
+    let newNamedQuery = (elements[2] as IBundleElement).namedQuery;
+    if (namedQuery?.name === 'test-query-new') {
+      // Swap
+      const q = namedQuery;
+      namedQuery = newNamedQuery;
+      newNamedQuery = q;
+    }
```
schmidt-sebastian, line 178: How about this:

```suggestion
    elements.sort((l, r) => l.nameQuery.name < r.nameQuery.name));
    let namedQuery = (elements[1] as IBundleElement).namedQuery;
    let newNamedQuery = (elements[2] as IBundleElement).namedQuery;
```

Or this:

```suggestion
    let namedQuery = elements.find(l.nameQuery.name === 'foo')!;
    let newNamedQuery = elements.find(l.nameQuery.name === 'bar')!;
```

dev/src/bundle.ts
```diff
@@ -84,16 +84,33 @@ export class BundleBuilder {
...
+        },
+      });
+    }
+
+    // Update `queries` to include both original and `queryName`.
+    const newDocument = this.documents.get(snap.id)!;
+    newDocument.metadata.queries = originalQueries || [];
+    if (queryName) {
+      newDocument.metadata.queries!.push(queryName);
+    }
```
schmidt-sebastian, line 112:
```suggestion
    newDocument.metadata.queries = [queryName, ...originalQueries];
```

dev/test/bundle.ts
```diff
@@ -158,7 +169,17 @@ describe('Bundle Buidler', () => {
...
+      .namedQuery;
+    let newNamedQuery = elements.find(
+      e => e.namedQuery?.name === 'test-query-new'
+    )!.namedQuery;
+    if (namedQuery?.name === 'test-query-new') {
+      // Swap
+      const q = namedQuery;
+      namedQuery = newNamedQuery;
+      newNamedQuery = q;
+    }
```
schmidt-sebastian, line 182: This is not needed anymore.

dev/src/bundle.ts
```diff
@@ -84,16 +84,33 @@ export class BundleBuilder {
...
+
     if (snap.readTime > this.latestReadTime) {
       this.latestReadTime = snap.readTime;
     }
```
schmidt-sebastian, line 116: Can you change the error message right below to:

'Query name conflict: "${name}" has already been added.'

schmidt-sebastian (approved):

dev/src/bundle.ts
wu-hui: I wrote it that way to avoid copying the queries every time a document is added, might not be as useful as i though, but i think the code is not too hard to understand?

dev/src/bundle.ts
wu-hui: Done.

dev/test/bundle.ts
wu-hui: Oops, fixed.

dev/src/bundle.ts
schmidt-sebastian: My comment wasn't based on improving code readability but brevity. If you are concerned about the extra copy then we can of course keep as is.

PR #1298, opened 2020-09-16
fix: bulkWriter: writing to the same document does not create a new batch

thebrianchen:


dev/src/bulk-writer.ts
```diff
@@ -229,56 +235,62 @@ class BulkCommitBatch {
...
-  private failRemainingOperations(results: BatchWriteResult[]): void {
-    for (const result of results) {
-      assert(
-        result.status.code !== Status.OK,
-        'Should not fail successful operation'
-      );
-      this.pendingOps.get(result.key)!.reject(result.status);
-      this.pendingOps.delete(result.key);
+    for (let i = resolvedOps.length - 1; i >= 0; i--) {
+      this.pendingOps.splice(resolvedOps[i], 1);
```
schmidt-sebastian, line 293: Aren't these indices off after the first splice? I wonder if we can simply build up the reverse array in the loop above and replace this.pendingOps with it.

dev/test/bulk-writer.ts
```diff
@@ -348,23 +332,20 @@ describe('BulkWriter', () => {
     expect(() => bulkWriter.close()).to.throw(expected);
   });
 
-  it('sends writes to the same document in separate batches', async () => {
+  it('sends writes to the same documents in the same batch', async () => {
```
schmidt-sebastian, line 335: Does this test still have value over the other tests?

schmidt-sebastian (commented):
This is mostly good, though I wonder if we need to fix the "docsToRetry" logic in the WriteBatch constructor?

dev/src/bulk-writer.ts
thebrianchen: Since we're going in reverse order and deleting the last element in the array each time, the remaining indices won't be affected.

dev/test/bulk-writer.ts
thebrianchen: This test would fail if pendingOps got changed back to its original map type since the second write to the same document path would override the existing entry, so it serves to make sure that writes to the same document are added properly.

dev/src/bulk-writer.ts
schmidt-sebastian: Can you add a comment?

dev/test/bulk-writer.ts
schmidt-sebastian: Makes sense. Maybe prefix the test name with "can send"?

dev/src/bulk-writer.ts
thebrianchen: done.

dev/test/bulk-writer.ts
thebrianchen: done.

schmidt-sebastian (commented):
Thanks for the updates. Can you find a way to address the comment in the summary response?

thebrianchen:
> Thanks for the updates. Can you find a way to address the comment in the summary response?

Updated to use an index-based tracking system instead. Also modified the retry test to include retries to the same document path, which would have failed with the previous iteration.

dev/src/bulk-writer.ts
```diff
@@ -229,75 +241,76 @@ class BulkCommitBatch {
...
    * Resolves the individual operations in the batch with the results.
    */
-  private processResults(results: BatchWriteResult[]): void {
-    for (const result of results) {
-      if (result.status.code === Status.OK) {
-        this.pendingOps.get(result.key)!.resolve(result);
-        this.pendingOps.delete(result.key);
+  private processResults(
+    results: BatchWriteResult[],
+    failRemainingOperations = false
```
schmidt-sebastian, line 278: I think this should be "allowRetry" or something like that, which also means that you can just merge the corresponding branch with `if (!this.shouldRetry(result.status.code))`.  In the context of this method, `failRemainingOperations` is quite self-explanatory.

dev/src/bulk-writer.ts
```diff
@@ -229,75 +241,76 @@ class BulkCommitBatch {
...
+        );
+        op.deferred.reject(result.status);
+      } else if (result.status.code === Status.OK) {
+        op.deferred.resolve(result);
       } else if (!this.shouldRetry(result.status.code)) {
-        this.pendingOps.get(result.key)!.reject(result.status);
-        this.pendingOps.delete(result.key);
+        op.deferred.reject(result.status);
+      } else {
+        // Retry the operation if it is not processed.
```
schmidt-sebastian, line 295: Maybe: 

```suggestion
        // Retry the operation if it has not been processed.
```

dev/src/write-batch.ts
```diff
@@ -144,29 +144,29 @@ export class WriteBatch implements firestore.WriteBatch {
...
     this._allowUndefined = !!firestore._settings.ignoreUndefinedProperties;
 
     if (retryBatch) {
-      // Creates a new WriteBatch containing only the operations from the
-      // provided document paths to retry.
+      // Creates a new WriteBatch containing only the indexes from the provided
+      // indexes to retry.
       this._ops = retryBatch._ops.filter(
-        v => docsToRetry!.indexOf(v.docPath) !== -1
+        (op, index) => indexesToRetry!.indexOf(index) !== -1
```
schmidt-sebastian, line 169: This is a bit confusing to read. Could we make indexesToRetry a `Set`? Optional.

thebrianchen: Good call. Done.

dev/src/bulk-writer.ts
thebrianchen: Done.

schmidt-sebastian (approved):

PR #1303, opened 2020-09-18
fix: change typings for select() to return Query<DocumentData>

### LINKED ISSUE #1299: Improve the types for Query.select
Currently `Query.select` returns a `Query<T>`. However, by specifying the fields we are interested in, we break the guarantee that the returned documents will be of type `T`.

https://github.com/googleapis/nodejs-firestore/blob/786e52f8c8b7b9c6b84ffc988190470a063d5855/dev/src/reference.ts#L1293

I think that instead, the return type should be set to `Query<DocumentData>`

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/1299

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2314,7 +2314,8 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
  * @class
  * @extends Query
  */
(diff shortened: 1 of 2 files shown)


thebrianchen (approved):
Thanks for fixing! LGTM after lint

PR #1304, opened 2020-09-22
fix: allow `setLogFunction(null)`

novemberborn:
`setLogFunction()` is typed to take a `null` value, but this doesn't actually work since it validates that `logger` is a function.

Change the logic so it only validates when `logger` is not `null`.

Looking at how `setLogFunction()` is used, it seems that it should be called with `null` instead of a no-op function. If this is contentious that commit can be reverted.

### DIFF
--- a/dev/src/logger.ts
+++ b/dev/src/logger.ts
@@ -55,7 +55,7 @@ export function logger(
  * `null` to turn off logging.
  */
 export function setLogFunction(logger: ((msg: string) => void) | null): void {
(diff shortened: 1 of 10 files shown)


novemberborn:
Is the commit convention check complaining about the PR title?!

schmidt-sebastian (approved):

schmidt-sebastian:
Thanks for this change! I am happy to take it.

PR #1305, opened 2020-09-22
feat: add starting/max rates to BulkWriterOptions

thebrianchen:
Following up from the 555 discussion by adding option to set the initial and maximum rate of the throttler in BulkWriter.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -30,7 +30,11 @@ import {Timestamp} from './timestamp';
 import {Deferred, getRetryCodes, isObject, wrapError} from './util';
 import {BatchWriteResult, WriteBatch, WriteResult} from './write-batch';
 import {logger} from './logger';
-import {invalidArgumentMessage, validateOptional} from './validate';
+import {
+  invalidArgumentMessage,
+  validateInteger,
+  validateOptional,
+} from './validate';
 
 /*!
  * The maximum number of writes that can be in a single batch.
@@ -757,57 +761,33 @@ function validateBulkWriterOptions(arg: string | number, value: unknown): void {
 
   const options = value as {[k: string]: unknown};
 
-  if ('disableThrottling' in options) {
-    if (typeof options.disableThrottling !== 'boolean') {
-      throw new Error(
-        `${invalidArgumentMessage(
-          arg,
-          'bulkWriter() options argument'
-        )} "disableThrottling" is not a boolean.`
-      );
-    }
-
-    if ('initialOpsPerSecond' in options || 'maxOpsPerSecond' in options) {
-      throw new Error(
-        `${invalidArgumentMessage(
-          arg,
-          'bulkWriter() options argument'
-        )} "disableThrottling" cannot be set with "initialOpsPerSecond" or "maxOpsPerSecond".`
-      );
-    }
-  }
-
   if (
-    'initialOpsPerSecond' in options &&
-    (typeof options.initialOpsPerSecond !== 'number' ||
-      options.initialOpsPerSecond <= 0 ||
-      options.initialOpsPerSecond % 1 !== 0)
+    'disableThrottling' in options &&
+    ('initialOpsPerSecond' in options || 'maxOpsPerSecond' in options)
   ) {
     throw new Error(
       `${invalidArgumentMessage(
         arg,
         'bulkWriter() options argument'
-      )} "initialOpsPerSecond" is not a positive integer.`
+      )} "disableThrottling" cannot be set with "initialOpsPerSecond" or "maxOpsPerSecond".`
     );
   }
 
+  if ('initialOpsPerSecond' in options) {
+    validateInteger('initialOpsPerSecond', options.initialOpsPerSecond, {
+      minValue: 0,
+    });
+  }
+
   if ('maxOpsPerSecond' in options) {
-    if (
-      typeof options.maxOpsPerSecond !== 'number' ||
-      options.maxOpsPerSecond <= 0 ||
-      options.maxOpsPerSecond % 1 !== 0
-    ) {
-      throw new Error(
-        `${invalidArgumentMessage(
-          arg,
-          'bulkWriter() options argument'
-        )} "maxOpsPerSecond" is not a positive integer.`
-      );
-    }
+    validateInteger('maxOpsPerSecond', options.maxOpsPerSecond, {
+      minValue: 0,
+    });
 
     if (
       'initialOpsPerSecond' in options &&
-      (options.initialOpsPerSecond as number) > options.maxOpsPerSecond
+      (options.initialOpsPerSecond as number) >
+        (options.maxOpsPerSecond as number)
     ) {
       throw new Error(
         `${invalidArgumentMessage(
@@ -819,7 +799,7 @@ function validateBulkWriterOptions(arg: string | number, value: unknown): void {
 
     if (
       !('initialOpsPerSecond' in options) &&
-      STARTING_MAXIMUM_OPS_PER_SECOND > options.maxOpsPerSecond
+      STARTING_MAXIMUM_OPS_PER_SECOND > (options.maxOpsPerSecond as number)
     ) {
       throw new Error(
         `${invalidArgumentMessage(
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -361,7 +361,7 @@ export class BulkWriter {
     this.firestore._incrementBulkWritersCount();
     validateBulkWriterOptions('options', options);
 
-    if (options?.disableThrottling !== false) {
+    if (options?.throttling === false) {
       this.rateLimiter = new RateLimiter(
         Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY,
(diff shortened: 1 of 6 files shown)


types/firestore.d.ts
```diff
@@ -613,12 +616,28 @@ declare namespace FirebaseFirestore {
...
+   * An options object to configure throttling on BulkWriter.
    */
   export interface BulkWriterOptions {
-    /** Whether to disable throttling. */
+    /**
+     * Whether to disable throttling. If set, initialOpsPerSecond and
+     * maxOpsPerSecond cannot be set.
+     */
     readonly disableThrottling?: boolean;
+
```
schmidt-sebastian, line 627: A lot of the APIs I have been dealing with lately do the following:

```
export interface BulkWriterOptions {
readonly throttling?: boolean : {initialOpsPerSecond?: number; maxOpsPerSecond?: number; };
}
```

This would match terser and eslint, but we might want to gather more feedback (maybe from @feiyang1).

dev/src/bulk-writer.ts
```diff
@@ -351,22 +352,25 @@ export class BulkWriter {
...
+    if (options && !options.disableThrottling) {
       this.rateLimiter = new RateLimiter(
+        Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY
       );
+    } else {
+      this.rateLimiter = new RateLimiter(
+        options?.initialOpsPerSecond || STARTING_MAXIMUM_OPS_PER_SECOND,
```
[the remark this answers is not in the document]
thebrianchen: That's a good tip! I'll try to keep that in mind for the future

dev/src/bulk-writer.ts
```diff
@@ -728,3 +732,101 @@ export class BulkWriter {
...
+
+  if ('disableThrottling' in options) {
+    if (typeof options.disableThrottling !== 'boolean') {
+      throw new Error(
+        `${invalidArgumentMessage(
+          arg,
+          'bulkWriter() options argument'
+        )} "disableThrottling" is not a boolean.`
+      );
+    }
```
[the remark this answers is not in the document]
thebrianchen: done.

thebrianchen (commented):
Used the union types for BulkWriterOptions. I think my constructor logic could be cleaner -- please advise :) 

I'll assign to Fei after you think it's in a good place.

types/firestore.d.ts
thebrianchen: attempted

dev/src/bulk-writer.ts
```diff
@@ -351,22 +356,36 @@ export class BulkWriter {
...
         Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY,
         Number.POSITIVE_INFINITY
       );
+    } else {
+      const startingRate =
+        typeof options?.throttling !== 'boolean' &&
+        options?.throttling?.initialOpsPerSecond !== undefined
+          ? options.throttling.initialOpsPerSecond
+          : STARTING_MAXIMUM_OPS_PER_SECOND;
```
schmidt-sebastian, line 376: I think this is:

```suggestion
      const startingRate = options?.throttling?.initialOpsPerSecond ??
          STARTING_MAXIMUM_OPS_PER_SECOND;
```

I tested it with this code:

```
const STARTING_MAXIMUM_OPS_PER_SECOND = 1337;
const options1 : any = {throttling:false};
const options2 : any = {throttling:true};
const options3 : any = {throttling:{initialOpsPerSecond:10}};

function printThrottler(options?:any) {
    console.log(options?.throttling?.initialOpsPerSecond ??STARTING_MAXIMUM_OPS_PER_SECOND)
}

printThrottler(options1);
printThrottler(options2);
printThrottler(options3);
```

This prints 1337, 1337, 10.

Same below.

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
     this.maxBatchSize = size;
   }
 }
+
+/**
+ * Validates the use of 'value' as BulkWriterOptions.
+ *
+ * @private
+ * @param arg The argument name or argument index (for varargs methods).
```
schmidt-sebastian, line 755: You can probably be more specific since this is only used for BulkWriterOptions.

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+ * @param value The object to validate.
+ * @throws if the input is not a valid BulkWriterOptions object.
+ */
+function validateBulkWriterOptions(arg: string | number, value: unknown): void {
```
schmidt-sebastian, line 759: This can just be a string.

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+  ) {
+    return;
+  }
+
+  if ('initialOpsPerSecond' in options.throttling) {
+    validateInteger(
+      'initialOpsPerSecond',
+      options.throttling.initialOpsPerSecond,
+      {
+        minValue: 0,
```
schmidt-sebastian, line 787: This needs to be 1, doesn't it?

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+      }
+    );
+  }
+
+  if ('maxOpsPerSecond' in options.throttling) {
+    validateInteger('maxOpsPerSecond', options.throttling.maxOpsPerSecond, {
+      minValue: 0,
```
schmidt-sebastian, line 794: 0 also doesn't quite work here.

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+    validateInteger(
+      'initialOpsPerSecond',
+      options.throttling.initialOpsPerSecond,
+      {
+        minValue: 0,
+      }
+    );
+  }
+
+  if ('maxOpsPerSecond' in options.throttling) {
```
schmidt-sebastian, line 792: If you change this to `options.throttling.maxOpsPerSecond !== undefined`, you can get rid of the bang below.

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+    validateInteger('maxOpsPerSecond', options.throttling.maxOpsPerSecond, {
+      minValue: 0,
+    });
+
+    if (
+      'initialOpsPerSecond' in options.throttling &&
```
schmidt-sebastian, line 798: Same: `options.throttling. initialOpsPerSecond !== undefined`

dev/src/bulk-writer.ts
```diff
@@ -728,3 +747,76 @@ export class BulkWriter {
...
+          'bulkWriter() options argument'
+        )} "maxOpsPerSecond" cannot be less than "initialOpsPerSecond".`
+      );
+    }
+
+    if (
+      !('initialOpsPerSecond' in options.throttling) &&
+      STARTING_MAXIMUM_OPS_PER_SECOND > options.throttling.maxOpsPerSecond!
+    ) {
+      throw new Error(
```
schmidt-sebastian, line 814: This is pretty unexpected, especially since the default value is not known. We should instead adjust the starting value.

BTW, I think we should somehow squeeze `DEFAULT` into the name of STARTING_MAXIMUM_OPS_PER_SECOND, since the constant is no longer required to be the maximum.

types/firestore.d.ts
```diff
@@ -613,12 +616,24 @@ declare namespace FirebaseFirestore {
...
+     * second allowed by the throttler. If this field is not set, the default
+     * is 500 operations per second.
+     * @param maxOpsPerSecond The maximum number of operations per second
+     * allowed by the throttler. If this field is set, the throttler's allowed
+     * operations per second does not ramp up past the specified operations per
+     * second.
+     */
+    readonly throttling?:
+      | boolean
+      | {initialOpsPerSecond?: number; maxOpsPerSecond?: number};
```
schmidt-sebastian, line 636: Thanks for the update. Did you check with @Feiyang1 if he thinks this is a useable API?

Feiyang1: This looks fine to me. Autocomplete will help people figure it out. Currently we are only describing the config params in the comment. I would make it clear in the comment that it can be either a boolean or a config object (boolean to enable/disable, config to enable and set config values).

thebrianchen: Was going to ask him after, but he jumped in :)

dev/src/bulk-writer.ts

dev/src/bulk-writer.ts
thebrianchen: Since it's a union type, `initialOpsPerSecond` is not a number unless I explicity check it's not a boolean first. 

With the additional change to lower the initial rate to match the max rate if it's below the default, I refactored it to use if statements, which is hopefully more readable.

dev/src/bulk-writer.ts
thebrianchen: removed used of `arg` as a parameter.

dev/src/bulk-writer.ts
```diff
@@ -351,22 +356,46 @@ export class BulkWriter {
 
   constructor(
     private readonly firestore: Firestore,
-    enableThrottling: boolean
+    private readonly options?: firestore.BulkWriterOptions
```
schmidt-sebastian, line 359:
```suggestion
    options?: firestore.BulkWriterOptions
```

types/firestore.d.ts
schmidt-sebastian: Do you also want to update the comment per Fei's suggestion?

schmidt-sebastian (approved):

types/firestore.d.ts
thebrianchen: I thought I had already added some more to the comments, but this time around I also explicitly stated that setting it to false disables throttling.

dev/src/bulk-writer.ts
thebrianchen: done.

PR #1307, opened 2020-09-23
fix: fix flaky BulkWriter rate limiter test

thebrianchen:
The test flakes when the CI slows down, which allows enough time to pass such that the test's writes can be made without exceeding the rate limit. To fix this, I created a mock rate limiter that always rejects any request, which makes this test timing-agnostic.


thebrianchen:
Update: also added an additional check in the BulkWriter constructor to lower the max batch size to match the starting rate if the starting rate is lower than the max batch size.

schmidt-sebastian (approved):

PR #1308, opened 2020-09-23
docs: expand on FieldPath docstr

crwilcox:
docs: expand on FieldPath docstr

### DIFF
--- a/dev/src/path.ts
+++ b/dev/src/path.ts
@@ -489,7 +489,7 @@ export function validateResourcePath(
 }
 
 /**
- * A dot-separated path for navigating sub-objects, for example a map, within a document. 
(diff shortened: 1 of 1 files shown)


dev/src/path.ts
```diff
@@ -489,7 +489,7 @@ export function validateResourcePath(
 }
 
 /**
- * A dot-separated path for navigating sub-objects within a document.
+ * A dot-separated path for navigating sub-objects, for example a map, within a document.
```
schmidt-sebastian, line 492:
```suggestion
 * A dot-separated path for navigating sub-objects (e.g. nested maps) within a document.
```

schmidt-sebastian (approved):

PR #1315, opened 2020-09-30
feat: add onWriteError() and onWriteResult() handlers to BulkWriter

### LINKED ISSUE #1282: [FR] Improve the BulkWriter API to prevent UnhandledRejections
The `BulkWriter` class was made public in version 4.1.0. however, it has some use disadvantages compared to using a `BatchWriter`.

The problem lies in the fact that `close` never fails, and the fact that writing (`update`, `delete`, `set`) returns a promise.

This makes it difficult and awkward to correctly handle errors, which is absolutely necessary on cloud functions, as `UnhandledRejections` will make the whole function to fail with a very cryptic error.

It'd be great if we could write something like:

```ts
const writer = db.bulkWriter();
writer.update(ref1, {...});
writer.update(ref2, {...});
try {
  await writer.commit();
} catch (error) {
  // decide whether to continue, fail, log or do something else based on the error.
  // error could be an object having a `failures` and `succeses` fields, where each failure or success has the all the arguments of the write operation. 
  // Or something similar to the result of promiseAllSettled. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled
}
```

Note that using `WriteBatch` resembles the proposed API.

Compare that to what one needs to do today.

```ts
let sharedState = ...;
const writer = db.bulkWriter();
 // depending on whether we want to skip failures or actually fail what the catch function will be
writer.update(ref1, {...}).catch(...);
writer.update(ref2, {...}).catch(...);
await writer.close();
// Check shared state
```

thebrianchen:
Fixes #1282.
- Adds global `onWriteError()` error and `onWriteResult()` success handlers to BulkWriter instances.
- Adds `BulkWriter.retry(operation: BulkWriterOperation)` method to retry failed operations.
- All failed operations throw a `BulkWriterError` that contain an underlying `BulkWriterOperation`.


dev/src/write-batch.ts
```diff
@@ -180,6 +180,15 @@ export class WriteBatch implements firestore.WriteBatch {
     return this._ops.length === 0;
   }
 
+  /**
+   * Returns the last operation that was added to the WriteBatch.
+   *
+   * @private
+   */
+  get lastOp(): PendingWriteOp {
```
schmidt-sebastian, line 188: s/lastOp/_lastOp

dev/src/write-batch.ts
```diff
@@ -547,6 +556,19 @@ export class WriteBatch implements firestore.WriteBatch {
     return this;
   }
 
+  /**
+   * Adds the provided document reference and PendingWriteOp to the list of
+   * WriteBatch operations.
+   *
+   * @private
+   */
+  addOperation(
```
schmidt-sebastian, line 565: s/addOperation/_ addOperation

dev/src/bulk-writer.ts
```diff
@@ -190,6 +195,15 @@ class BulkCommitBatch {
     return this.processOperation(documentRef);
   }
 
+  /**
+   * Adds the provided operation to the WriteBatch. Returns a promise that
+   * resolves with the result of the operation.
+   */
+  addOperation(operation: BulkWriterOperation): Promise<WriteResult> {
+    this.writeBatch.addOperation(operation.documentRef, operation._data);
```
schmidt-sebastian, line 203: My dream of somehow removing the BulkCommitBatch is growing more and more vivid. Certainly not in this PR though :)

dev/src/bulk-writer.ts
```diff
@@ -436,6 +484,15 @@ export class BulkWriter {
...
     const resultPromise = bulkCommitBatch.create(documentRef, data);
+    const pendingOp = bulkCommitBatch.getLastOp();
+    resultPromise.catch(err => {
+      const operation = new BulkWriterOperation(
+        documentRef,
+        'create',
+        pendingOp
+      );
+      this.errorFn(err, operation);
+    });
```
schmidt-sebastian, line 495: Ideally, we would only attach our own Promise if `errorFn` is defined, but that would require users to register an `error` function before any other API call. I think that we could re-throw the original exception here though if the error callback is not defined. This can be a simple change if you change the default error function to just throw.

dev/src/bulk-writer.ts
```diff
@@ -589,6 +660,43 @@ export class BulkWriter {
...
+    this.sendReadyBatches();
+    return resultPromise;
+  }
+
+  /**
+   * Attaches a listener that is run every time a BulkWriter operation fails.
+   *
+   * @param fn A callback to be called every time a BulkWriter operation
```
schmidt-sebastian, line 679: We don't use `fn` in our other APIs. Can this be `callback`?

dev/src/bulk-writer.ts
```diff
@@ -589,6 +660,43 @@ export class BulkWriter {
...
+  }
+
+  /**
+   * Retry a BulkWriter operation that has failed.
+   *
+   * The BulkWriterOperation object is a parameter on the `onError()`
+   * callback function that is called whenever a write fails.
+   *
+   * @param operation The operation to retry.
+   * @return A promise that resolves with the result of the operation. Throws
```
schmidt-sebastian, line 693: s/Throws/Rejects if the retry fails (with Brianified and improved grammer)

dev/test/bulk-writer.ts
```diff
@@ -477,6 +482,146 @@ describe('BulkWriter', () => {
...
+        request: createRequest([createOp('doc', 'bar')]),
+        response: failedResponse(Status.FAILED_PRECONDITION),
+      },
+      {
+        request: createRequest([createOp('doc', 'bar')]),
+        response: successResponse(1),
+      },
+    ]);
+    let op: BulkWriterOperation | undefined = undefined;
+    bulkWriter.onError(async (error, operation) => {
```
schmidt-sebastian, line 497:
```suggestion
    bulkWriter.onError((error, operation) => {
```

dev/test/bulk-writer.ts
```diff
@@ -477,6 +482,146 @@ describe('BulkWriter', () => {
...
+    let writeResult: WriteResult;
+    bulkWriter.retry(op!).then(res => {
+      writeResult = res;
+    });
+    return bulkWriter.close().then(() => {
+      expect(writeResult.writeTime.isEqual(new Timestamp(1, 0))).to.be.true;
+    });
+  });
+
+  it('can retry failed set operations with retry()', async () => {
```
schmidt-sebastian, line 512: Do you think these duplicate tests add value? If so, you should try to extract helpers to reduce code duplication.

dev/test/bulk-writer.ts
```diff
@@ -477,6 +482,146 @@ describe('BulkWriter', () => {
...
+      expect(writeResult.writeTime.isEqual(new Timestamp(1, 0))).to.be.true;
+    });
+  });
+
+  it('retried operations are enqueued after other operations', async () => {
+    const bulkWriter = await instantiateInstance([
+      {
+        request: createRequest([setOp('doc1', 'bar'), updateOp('doc2', 'bar')]),
+        response: mergeResponses([
+          failedResponse(Status.FAILED_PRECONDITION),
```
schmidt-sebastian, line 598: How about `Internal`? `FAILED_PRECONDITION` is a bit strange since the retry should also fail.

types/firestore.d.ts
```diff
@@ -582,6 +582,28 @@ declare namespace FirebaseFirestore {
...
+     */
+    onError(
+      errorFn: (error: Error, operation: BulkWriterOperation) => void
+    ): void;
+
+    /**
+     * Retry a BulkWriter operation that has failed.
+     *
+     * The BulkWriterOperation object is a parameter on the `onError()`
+     * callback function that is called whenever a write fails.
```
schmidt-sebastian, line 599: We should mention that is can only be used in combination with `flush()`. This should probably also be part of the error message when a retry is enqueue after `close()` is called.

schmidt-sebastian (commented):
This looks good. Before you address the feedback, we should see if we want to incorporate @bcoe's API suggestion.

dev/src/bulk-writer.ts
```diff
@@ -648,16 +850,33 @@ export class BulkWriter {
...
-    this.closed = true;
-    return flushPromise;
+    this.closeCalled = true;
+    await flushPromise;
+
+    // Wait for pending retry operations to run.
+    if (this.batchQueue.length > 0) {
+      await this._flush();
+      await this.closedDeferred.promise;
+    }
```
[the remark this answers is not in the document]
tilgovi: I now see that `verifyNotClosed` only throws if fully closed when called from `retry`. In which case, I think my second question still stands. Do we know that the `await this._flush()` won't cause more retries? Do we either need to do this in a loop or at least await another flush after `this.closeDeferred.promise` resolves?

dev/src/bulk-writer.ts
schmidt-sebastian: Something like:

```
const defaultCatch = (e) => throw e;

set() {
	this.op().catch(() => defaultCatch())
}

onError(errorFn) {
  defaultCatch = errorFn;
}
```

This should avoid the "uncaught Promise rejection" handling only when `onError` is called.

dev/src/bulk-writer.ts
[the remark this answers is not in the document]
schmidt-sebastian: I think we can use Promise chaining here as proposed in the diff above. Instead of requiring `void` as the return type of `onWriteError`, we can require the user to call `retry` and to pass the result of `retry` to use via the error handler. The signature would then be:

`onWriteError((err:BulkWriterError) => Promise<WriteResult>)`

Once we do that, we can chain the result of this error callback and block on the retry attempt in the flush call.

If we don't do this, we have to somehow ensure that all error callbacks have fired once we process the results. This would be a bit tricky and may not port to Java or other languages:

```
error.reject(...);
setImmediate(() => {
  // This only works in Node, but due to the way that ticks are processed in the VM, we can use a `setImmediate`
  // to ensure that our code only runs once all Promise chains are processed.
});
```

dev/src/bulk-writer.ts
```diff
@@ -762,17 +850,29 @@ export class BulkWriter {
...
+   */
+  private verifyNotClosed(checkIsClosed = false): void {
+    if (!checkIsClosed && this.closeCalled) {
+      throw new Error('BulkWriter has already been closed.');
+    }
+
+    if (checkIsClosed && this.isClosed) {
       throw new Error('BulkWriter has already been closed.');
     }
   }
```
schmidt-sebastian, line 882: Should these be two different functions?

types/firestore.d.ts
```diff
@@ -663,12 +672,30 @@ declare namespace FirebaseFirestore {
...
+   */
+  export class BulkWriterError extends Error {
+    /** The operation that failed. Can be used to retry the operation. */
+    readonly operation: BulkWriterOperation;
+
+    /** The status code of the error. */
+    readonly code: GrpcStatus;
+
+    /** The error message of the error. */
+    readonly message: string;
```
schmidt-sebastian, line 688: I think we may need to include a retry count/attempt number to let users decide when to give up.

dev/src/bulk-writer.ts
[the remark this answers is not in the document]
tilgovi: That would be elegant, I think. The ultimate result of the original write would become the result of the retry. It would just be important to document well so that callers don't forget to return the result of `retry()` from their error handler or the retry promise would end up unhandled and the original write would resolve with undefined.

thebrianchen:
@tilgovi Quick update -- Still working on it in a separate branch https://github.com/googleapis/nodejs-firestore/pull/1330.

At this point, I have the auto-retry functionality working. `onWriteError()` now takes a callback with the signature `(error: BulkWriterError): boolean` where returning `true` automatically retries, and returning `false` throws. Aside from a lot of code cleanup to do, the last bug that is in the implementation is that `close()` resolves a tick before the last retry operation resolves. Hoping to fix that tomorrow.

dev/src/bulk-writer.ts
```diff
@@ -738,29 +852,97 @@ export class BulkWriter {
...
+            throw bulkWriterError;
+          }
+        }
+      }
+    } finally {
+      operationCompletedDeferred.resolve();
+      this._pendingOps.delete(operationCompletedDeferred.promise);
+    }
+  }
+
```
schmidt-sebastian, line 937: My last effort cleanup attempt:

```suggestion
  /**
   * Schedules and runs the provided operation.
   */
  private async _executeWrite(
    documentRef: firestore.DocumentReference,
    operationType: 'create' | 'set' | 'update' | 'delete',
    operationFn: (bulkCommitBatch: BulkCommitBatch) => Promise<WriteResult>
  ): Promise<WriteResult> {
    let result : WriteResult | Promise<WriteResult> | undefined;
    
    // A deferred promise that resolves when operationFn completes.
    const operationCompletedDeferred = new Deferred<void>();
    this._pendingOps.add(operationCompletedDeferred.promise);

    let batchQueue = this._batchQueue;
    this.sendReadyBatches(batchQueue);
    
    for (let failedAttempts = 0; result != undefined; ++failedAttempts) {
      const bulkCommitBatch = this.getEligibleBatch(batchQueue);

      try {
        result = await operationFn(bulkCommitBatch);
        this._successFn(documentRef, result);
      } catch (error) {
        const bulkWriterError = new BulkWriterError(
          error.code,
          error.message,
          documentRef,
          operationType,
          failedAttempts
        );
        const shouldRetry = this._errorFn(bulkWriterError);
        logger(
          'BulkWriter.errorFn',
          null,
          'Running error callback on error code:',
          error.code,
          ', shouldRetry:',
          shouldRetry
        );
        if (shouldRetry) {
          batchQueue = this._retryBatchQueue;
        } else {
          result = Promise.reject(bulkWriterError);
        }
      }
    }
    
    operationCompletedDeferred.resolve();
    this._pendingOps.delete(operationCompletedDeferred.promise);
    return result!;
  }
```

Let me know what you think.

dev/system-test/firestore.ts
```diff
@@ -2612,6 +2613,24 @@ describe('BulkWriter class', () => {
     await writer.close();
     return firestore.terminate();
   });
+
+  it('can retry failed writes with a provided callback', async () => {
+    let retryCount = 0;
+    let code: Status = -1;
+    writer.onWriteError(error => {
+      retryCount = error.failedAttempts;
+      return error.failedAttempts < 30;
```
schmidt-sebastian, line 2622: Since this is a system test, can you lower this to 3 or something?

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+          setOp('doc', 'bar'),
+          updateOp('doc', 'bar'),
+          deleteOp('doc'),
+        ]),
+        response: mergeResponses([
+          failedResponse(Status.INTERNAL),
+          failedResponse(Status.INTERNAL),
+          failedResponse(Status.INTERNAL),
+          failedResponse(Status.INTERNAL),
+        ]),
```
schmidt-sebastian, line 540: Can you modify this test to only fail some of the initial writes?

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+      },
+    ]);
+    const ops: string[] = [];
+    const writeResults: number[] = [];
+    bulkWriter.onWriteError(error => {
+      ops.push(error.operationType);
+      return true;
+    });
+    bulkWriter.onWriteResult((documentRef, result) => {
+      writeResults.push(result.writeTime.seconds);
```
schmidt-sebastian, line 564: Suggestion: Add to the `ops` array here as well.

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+      },
+    ]);
+    const ops: number[] = [];
+    bulkWriter.onWriteError(() => {
+      return true;
+    });
+    bulkWriter
+      .set(firestore.doc('collectionId/doc'), {foo: 'bar'})
+      .then(res => {
+        ops.push(res.writeTime.seconds);
```
schmidt-sebastian, line 696:
```suggestion
        ops.push("before_flush");
```

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+      });
+    const flush = bulkWriter.flush().then(() => {
+      ops.push(42);
```
schmidt-sebastian, line 699:
```suggestion
      ops.push("flush");
```

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+    });
+    bulkWriter
+      .set(firestore.doc('collectionId/doc2'), {foo: 'bar'})
+      .then(res => {
+        ops.push(res.writeTime.seconds);
```
schmidt-sebastian, line 704: "after_flush"

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+      });
+
+    await flush;
+    expect(ops).to.deep.equal([1, 42]);
```
schmidt-sebastian, line 708:
```suggestion
    expect(ops).to.deep.equal(["before_flush", "flush"]);
```

dev/test/bulk-writer.ts
```diff
@@ -477,6 +493,256 @@ describe('BulkWriter', () => {
...
+    return bulkWriter.close().then(() => {
+      expect(ops).to.deep.equal([1, 42, 2]);
```
schmidt-sebastian, line 710:
```suggestion
      expect(ops).to.deep.equal(["before_flush", "flush", "after_flush"]);
```

schmidt-sebastian (commented):
Whee. Very nice. Some nits left but it looks like we are basically good to go

schmidt-sebastian (approved):
Whee!

PR #1326, opened 2020-10-08
docs: update getAll example in documentation

n370:
Help users avoid getting error message`getAll() no longer accepts an array as its first argument. Please unpack your array and call getAll() with individual arguments`.


n370:
@googlebot I signed it!

n370:
I figured `getAll` will also error if documentRefs is an empty array. If y'all think it's a good idea I can try to update it to make it return an empty array if it is called without arguments.

schmidt-sebastian (approved):

schmidt-sebastian:
@n370 Thanks for this!

PR #1328, opened 2020-10-14
feat: add implicit ordering for startAt(DocumentReference) calls

schmidt-sebastian:
The backend orders every query implicitly by document name, but if a cursor is specified, the order constraint has to be made explicit. This makes the Partition API harder to use, since the user needs to manually "prepend" and orderBy before they can use the cursor values in a QueryPartition.

This PR adds an implicit orderBy if the only cursor value is a DocumentReference. This is based on existing logic for DocumentSnapshots.


thebrianchen (approved):
lgtm

PR #1330, opened 2020-10-15
Bc/bulk error debug

schmidt-sebastian:

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -316,6 +316,7 @@ export class BulkWriter {
   private batchQueue: BulkCommitBatch[] = [];
 
   private _pendingCommits: Promise<void>[] = [];
+  private _pendingOps: Promise<void>[] = [];
 
   /**
    * Whether this BulkWriter instance has started to close. Afterwards, no
@@ -441,6 +442,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const op = this._create(documentRef,data);
     this.sendReadyBatches();
+    this._pendingOps.push(op.then(() => {}, () => {}));
     return op;
   }
 
@@ -457,6 +459,7 @@ export class BulkWriter {
           return res;
         })
         .catch(error => {
+          console.log('creating error');
           const bulkWriterError = new BulkWriterError(
               error.code,
               error.message,
@@ -468,7 +471,7 @@ export class BulkWriter {
           if (shouldRetry) {
             return this._create(documentRef, data, retryCount + 1);
           } else {
-            throw error;
+            throw bulkWriterError;
           }
         });
   }
@@ -508,6 +511,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const op = this._delete(documentRef,precondition);
     this.sendReadyBatches();
+    this._pendingOps.push(op.then(() => {}, () => {}));
     return op;
   }
 
@@ -535,7 +539,7 @@ export class BulkWriter {
           if (shouldRetry) {
             return this._delete(documentRef, precondition, retryCount + 1);
           } else {
-            throw error;
+            throw bulkWriterError;
           }
         });
   }
@@ -592,6 +596,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const op = this._set(documentRef, data, options);
     this.sendReadyBatches();
+    this._pendingOps.push(op.then(() => {}, () => {}));
     return op;
   }
 
@@ -620,7 +625,7 @@ export class BulkWriter {
           if (shouldRetry) {
             return this._set(documentRef, data, options, retryCount + 1);
           } else {
-            throw error;
+            throw bulkWriterError;
           }
         });
   }
@@ -676,6 +681,7 @@ export class BulkWriter {
     this.verifyNotClosed();
     const op = this._update(documentRef, dataOrField, preconditionOrValues);
     this.sendReadyBatches();
+    this._pendingOps.push(op.then(() => {}, () => {}));
     return op;
   }
 
@@ -716,7 +722,7 @@ export class BulkWriter {
                 retryCount + 1
             );
           } else {
-            throw error;
+            throw bulkWriterError;
           }
         });
   }
@@ -793,6 +799,8 @@ export class BulkWriter {
         retry = true;
       }
     }
+    // Make sure user promises resolve before _flush
+    await Promise.all(this._pendingOps);
   }
 
   /**
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -93,8 +93,13 @@ enum BatchState {
 interface PendingOp {
   writeBatchIndex: number;
   deferred: Deferred<BatchWriteResult>;
+  operationInfo: OperationInfo;
+}
+
+interface OperationInfo {
   documentRef: firestore.DocumentReference;
-  operation: 'create' | 'set' | 'update' | 'delete';
+  operationType: 'create' | 'set' | 'update' | 'delete';
+  retryCount: number;
   data: PendingWriteOp;
 }
 
@@ -143,11 +148,13 @@ class BulkCommitBatch {
     data: T
   ): Promise<WriteResult> {
     this.writeBatch.create(documentRef, data);
-    return this.processOperation(
+    const operationInfo = {
       documentRef,
-      'create',
-      this.writeBatch._lastOp
-    );
+      operationType: 'create',
+      retryCount: 0,
+      data: this.writeBatch._lastOp,
+    } as OperationInfo;
+    return this.processOperation(operationInfo);
   }
 
   /**
@@ -159,11 +166,13 @@ class BulkCommitBatch {
     precondition?: firestore.Precondition
   ): Promise<WriteResult> {
     this.writeBatch.delete(documentRef, precondition);
-    return this.processOperation(
+    const operationInfo = {
       documentRef,
-      'delete',
-      this.writeBatch._lastOp
-    );
+      operationType: 'delete',
+      retryCount: 0,
+      data: this.writeBatch._lastOp,
+    } as OperationInfo;
+    return this.processOperation(operationInfo);
   }
 
   set<T>(
@@ -190,7 +199,13 @@ class BulkCommitBatch {
     options?: firestore.SetOptions
   ): Promise<WriteResult> {
     this.writeBatch.set(documentRef, data, options);
-    return this.processOperation(documentRef, 'set', this.writeBatch._lastOp);
+    const operationInfo = {
+      documentRef,
+      operationType: 'set',
+      retryCount: 0,
+      data: this.writeBatch._lastOp,
+    } as OperationInfo;
+    return this.processOperation(operationInfo);
   }
 
   /**
@@ -205,24 +220,25 @@ class BulkCommitBatch {
     >
   ): Promise<WriteResult> {
     this.writeBatch.update(documentRef, dataOrField, ...preconditionOrValues);
-    return this.processOperation(
+    const operationInfo = {
       documentRef,
-      'update',
-      this.writeBatch._lastOp
-    );
+      operationType: 'update',
+      retryCount: 0,
+      data: this.writeBatch._lastOp,
+    } as OperationInfo;
+    return this.processOperation(operationInfo);
   }
 
   /**
    * Adds the provided operation to the WriteBatch. Returns a promise that
    * resolves with the result of the operation.
    */
-  addOperation(
-    documentRef: firestore.DocumentReference,
-    operation: 'create' | 'set' | 'update' | 'delete',
-    data: PendingWriteOp
-  ): Promise<WriteResult> {
-    this.writeBatch._addOperation(documentRef, data);
-    return this.processOperation(documentRef, operation, data);
+  addOperation(operationInfo: OperationInfo): Promise<WriteResult> {
+    this.writeBatch._addOperation(
+      operationInfo.documentRef,
+      operationInfo.data
+    );
+    return this.processOperation(operationInfo);
   }
 
   /**
@@ -230,9 +246,7 @@ class BulkCommitBatch {
    * return the result.
    */
   private processOperation<T>(
-    documentRef: firestore.DocumentReference,
-    operation: 'create' | 'set' | 'update' | 'delete',
-    data: PendingWriteOp
+    operationInfo: OperationInfo
   ): Promise<WriteResult> {
     assert(
       this.state === BatchState.OPEN,
@@ -242,9 +256,7 @@ class BulkCommitBatch {
     this.pendingOps.push({
       writeBatchIndex: this.opCount,
       deferred: deferred,
-      documentRef,
-      operation,
-      data,
+      operationInfo,
     });
 
     if (this.opCount === this.maxBatchSize) {
@@ -334,9 +346,7 @@ class BulkCommitBatch {
         const error = new BulkWriterError(
           (result.status.code as number) as GrpcStatus,
           result.status.message,
-          op.documentRef,
-          op.operation,
-          op.data
+          op.operationInfo
         );
         op.deferred.reject(error);
       } else {
@@ -346,9 +356,7 @@ class BulkCommitBatch {
         newPendingOps.push({
           writeBatchIndex: i,
           deferred: op.deferred,
-          documentRef: op.documentRef,
-          operation: op.operation,
-          data: op.data,
+          operationInfo: op.operationInfo,
         });
       }
     }
@@ -375,13 +383,6 @@ class BulkCommitBatch {
       this.state = BatchState.READY_TO_SEND;
     }
   }
-
-  /**
-   * Returns the last operation that was added to this batch.
-   */
-  getLastOp(): PendingWriteOp {
-    return this.writeBatch._lastOp;
-  }
 }
 
 /**
@@ -390,6 +391,18 @@ class BulkCommitBatch {
  * @class BulkWriterError
  */
 export class BulkWriterError extends Error {
+  /** The document reference the operation was performed on. */
+  readonly documentRef: firestore.DocumentReference;
+
+  /** The type of operation performed. */
+  readonly operationType: 'create' | 'set' | 'update' | 'delete';
+
+  /** How many times this operation has been retried. */
+  readonly retryCount: number;
+
+  /** Internal proto representation of the operation. */
+  readonly _data: PendingWriteOp;
+
   /** @hideconstructor */
   constructor(
     /** The status code of the error. */
@@ -398,16 +411,13 @@ export class BulkWriterError extends Error {
     /** The error message of the error. */
     readonly message: string,
 
-    /** The document reference the operation was performed on. */
-    readonly documentRef: firestore.DocumentReference,
-
-    /** The type of operation performed. */
-    readonly operation: 'create' | 'set' | 'update' | 'delete',
-
-    /** Internal proto representation of the operation. */
-    readonly _data: PendingWriteOp
+    operationInfo: OperationInfo
   ) {
     super(message);
+    this.documentRef = operationInfo.documentRef;
+    this.operationType = operationInfo.operationType;
+    this.retryCount = operationInfo.retryCount;
+    this._data = operationInfo.data;
   }
 }
 
@@ -465,7 +475,16 @@ export class BulkWriter {
    * The user-provided callback to be run every time a BulkWriter operation
    * fails.
    */
-  private errorFn: (error: BulkWriterError) => void;
+  private errorFn: (error: BulkWriterError) => boolean;
+
+  private errorOrBust = (error: BulkWriterError): Promise<WriteResult> => {
+    const shouldRetry = this.errorFn(error);
+    if (shouldRetry) {
+      return this._retry(error).catch(this.errorOrBust);
+    } else {
+      throw error;
+    }
+  };
 
   /** @hideconstructor */
   constructor(
@@ -473,7 +492,9 @@ export class BulkWriter {
     options?: firestore.BulkWriterOptions
   ) {
     this.firestore._incrementBulkWritersCount();
-    this.errorFn = (e) => {throw e;};
+    this.errorFn = () => {
+      return false;
+    };
     this.successFn = () => {};
     validateBulkWriterOptions(options);
 
@@ -553,15 +574,13 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch();
     const resultPromise = bulkCommitBatch.create(documentRef, data);
-    resultPromise
+    this.sendReadyBatches();
+    return resultPromise
       .then(res => {
         this.successFn(documentRef, res);
+        return res;
       })
-      .catch(err => {
-        this.errorFn(err);
-      });
-    this.sendReadyBatches();
-    return resultPromise;
+      .catch(this.errorOrBust);
   }
 
   /**
@@ -599,16 +618,13 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch();
     const resultPromise = bulkCommitBatch.delete(documentRef, precondition);
-    const pendingOp = bulkCommitBatch.getLastOp();
-    resultPromise
+    this.sendReadyBatches();
+    return resultPromise
       .then(res => {
         this.successFn(documentRef, res);
+        return res;
       })
-      .catch(err => {
-        this.errorFn(err);
-      });
-    this.sendReadyBatches();
-    return resultPromise;
+      .catch(this.errorOrBust);
   }
 
   set<T>(
@@ -663,16 +679,15 @@ export class BulkWriter {
     this.verifyNotClosed();
     const bulkCommitBatch = this.getEligibleBatch();
     const resultPromise = bulkCommitBatch.set(documentRef, data, options);
-    const pendingOp = bulkCommitBatch.getLastOp();
-    resultPromise
+    this.sendReadyBatches();
+    return resultPromise
       .then(res => {
         this.successFn(documentRef, res);
+        return res;
       })
-      .catch(err => {
-        this.errorFn(err);
+      .catch(error => {
+        return this.errorOrBust(error);
       });
-    this.sendReadyBatches();
-    return resultPromise;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


dev/src/bulk-writer.ts
```diff
@@ -434,10 +455,27 @@ export class BulkWriter {
     data: T
   ): Promise<WriteResult> {
     this.verifyNotClosed();
-    const bulkCommitBatch = this.getEligibleBatch();
-    const resultPromise = bulkCommitBatch.create(documentRef, data);
+    const op = this._create(documentRef, data);
```
schmidt-sebastian, line 458: Instead of having a helper function for each call that uses a lambda internally, can you just call `wrapRetry` and pass `bulkCommitBatch.create` directly?

dev/src/bulk-writer.ts
```diff
@@ -583,14 +656,65 @@ export class BulkWriter {
...
+      retryFn
     );
-    this.sendReadyBatches();
-    return resultPromise;
+  }
+
+  /**
+   * Attaches a listener that is run every time a BulkWriter operation
+   * successfully completes.
+   *
```
schmidt-sebastian, line 696: Can you add an `@example` block here?

dev/src/bulk-writer.ts
```diff
@@ -583,14 +656,65 @@ export class BulkWriter {
...
+      documentRef: firestore.DocumentReference,
+      result: WriteResult
+    ) => void
+  ): void {
+    this._successFn = callback;
+  }
+
+  /**
+   * Attaches a listener that is run every time a BulkWriter operation fails.
+   *
```
schmidt-sebastian, line 711: Please add an `@example` block that shows how to use the retry counter.

dev/src/bulk-writer.ts
```diff
@@ -583,14 +656,65 @@ export class BulkWriter {
...
+   * @param shouldRetryError A callback to be called every time a BulkWriter
+   * operation fails. Returning `true` will retry the operation. Returning
+   * `false` will stop the retry loop.
+   */
+  onWriteError(shouldRetryError: (error: BulkWriterError) => boolean): void {
```
schmidt-sebastian, line 716:
```suggestion
  onWriteError(shouldRetryCallback: (error: BulkWriterError) => boolean): void {
```

dev/src/bulk-writer.ts
```diff
@@ -620,16 +744,33 @@ export class BulkWriter {
...
+  }
+
+  private async _flush(): Promise<void> {
+    const trackedBatches = this._batchQueue;
+    trackedBatches.map(batch => batch.markReadyToSend());
+    let retriesCompleted = false;
+    while (!retriesCompleted) {
+      const before = this._pendingBatches.size;
+      this.sendReadyBatches();
+      const after = this._pendingBatches.size;
```
schmidt-sebastian, line 757: If a user adds a new batch, this will send the new user writes as well, won't it? I think this creates the "flush loops over all writes while any write is pending and is impossible to understand API" that I had in one of my earlier revisions. I think what you want is a single commit of all non-retry batches and then a final loop that drains the retry batches.

dev/src/bulk-writer.ts
```diff
@@ -620,16 +744,33 @@ export class BulkWriter {
...
     this.verifyNotClosed();
-    const trackedBatches = this.batchQueue;
-    const writePromises = trackedBatches.map(batch => batch.awaitBulkCommit());
-    this.sendReadyBatches();
-    await Promise.all(writePromises);
+    await this._flush();
+  }
+
+  private async _flush(): Promise<void> {
+    const trackedBatches = this._batchQueue;
```
schmidt-sebastian, line 751: s/trackedBatches/existingBatches ?

dev/src/bulk-writer.ts
```diff
@@ -648,14 +789,20 @@ export class BulkWriter {
    *   console.log('Executed all writes');
    * });
    */
-  close(): Promise<void> {
+  async close(): Promise<void> {
```
schmidt-sebastian, line 792: Public API methods shouldn't use the `async` keyword since it messes with the execution order. `this.closed = true;` should be set synchronously. If you make this method `async`, it executes one Promise tick later.

dev/src/bulk-writer.ts
```diff
@@ -648,14 +789,20 @@ export class BulkWriter {
...
     this.verifyNotClosed();
     this.firestore._decrementBulkWritersCount();
     const flushPromise = this.flush();
     this.closed = true;
-    return flushPromise;
+    await flushPromise;
+
+    // Make sure user promises resolve before close() resolves.
+    await Promise.all(this._pendingOps);
```
schmidt-sebastian, line 800: I think the exact problem that this line addresses also exists in `flush` (i.e. a retry can take longer to resolve than the Promise returned from this method). We should move this logic to `_flush`.

dev/src/bulk-writer.ts
```diff
@@ -668,14 +815,21 @@ export class BulkWriter {
    *
    * @private
    */
-  private getEligibleBatch<T>(): BulkCommitBatch {
-    if (this.batchQueue.length > 0) {
-      const lastBatch = this.batchQueue[this.batchQueue.length - 1];
-      if (lastBatch.state === BatchState.OPEN) {
+  private getEligibleBatch<T>(retryCount: number): BulkCommitBatch {
```
schmidt-sebastian, line 818: If we don't need the count here, we can pass in a boolean (or even the queue we would use)

dev/src/bulk-writer.ts
```diff
@@ -708,29 +867,44 @@ export class BulkWriter {
    * @private
    */
   private sendReadyBatches(): void {
```
schmidt-sebastian, line 869: This API might also be simpler if we passed a pointer with the queue we want to process.

dev/src/bulk-writer.ts
```diff
@@ -708,29 +867,44 @@ export class BulkWriter {
...
+    const sendRetries = this._retryBatchQueue.length > 0;
+    const unsentBatches = sendRetries
+      ? this._retryBatchQueue
+      : this._batchQueue;
+    for (let index = 0; index < unsentBatches.length; ++index) {
       const batch = unsentBatches[index];
 
+      // Send all retries whenever possible.
+      if (sendRetries) {
+        batch.markReadyToSend();
```
schmidt-sebastian, line 880: I was wondering if we need this call? Could we not transition from OPEN to SEND? If we do need it (I haven't thought it quite through) then maybe it makes sense to post-process the retry queue before entering here.

dev/src/bulk-writer.ts
```diff
@@ -708,29 +867,44 @@ export class BulkWriter {
...
+      }
+
+      if (batch.state !== BatchState.READY_TO_SEND) {
```
schmidt-sebastian, line 883: With the two different queues, I think we can once again go back to the original `while` loop and exit once we see the first batch that is not sendable.

dev/src/bulk-writer.ts
```diff
@@ -708,29 +867,44 @@ export class BulkWriter {
...
+          .then(batchDeferred.resolve, batchDeferred.reject)
+          .then(() => {
+            this._pendingBatches.delete(batchDeferred.promise);
+          });
       } else {
-        delayExecution(() => this.sendReadyBatches(), delayMs);
+        delayExecution(() => {
+          batchDeferred.resolve();
+          this._pendingBatches.delete(batchDeferred.promise);
+          this.sendReadyBatches();
```
schmidt-sebastian, line 904: Please add a comment here that explain that `batchDeferred` in this case isn't actually a batch, but rather just an operation that waits for the timeout to fire, which will ad more batches.

dev/src/bulk-writer.ts
```diff
@@ -708,29 +867,44 @@ export class BulkWriter {
...
       // attempt after the appropriate timeout.
       const delayMs = this.rateLimiter.getNextRequestDelayMs(batch.opCount);
       assert(delayMs !== -1, 'Batch size should be under capacity');
       if (delayMs === 0) {
-        this.sendBatch(batch);
+        this.sendBatch(batch, sendRetries)
+          .then(batchDeferred.resolve, batchDeferred.reject)
+          .then(() => {
+            this._pendingBatches.delete(batchDeferred.promise);
+          });
```
schmidt-sebastian, line 899: I think that rather than using a Set and a one-by-one cleanup you can just mass-delete all Promises you waited on at the end of the `_flush` call. Y

If you want to clean them up as the commit happens, can you think of a way to handle all cleanup in `this.sendBatch`?

dev/src/bulk-writer.ts
```diff
@@ -326,34 +281,99 @@ class BulkCommitBatch {
...
-  private batchQueue: BulkCommitBatch[] = [];
+  private _batchQueue: BulkCommitBatch[] = [];
+
+  private _retryBatchQueue: BulkCommitBatch[] = [];
+
+  /**
+   * A list of promises that represent sent batches. Each promise is resolved
+   * when the batch's response is received.
+   */
+  private _pendingBatches: Set<Promise<void>> = new Set();
```
schmidt-sebastian, line 334: If you get creative, we should indicate via the naming that `_pendingBatches` contains both `_batchQueue` and `_retryBatchQueue`.

dev/src/bulk-writer.ts
```diff
@@ -740,27 +914,69 @@ export class BulkWriter {
...
+      const batchIndex = batchQueue.indexOf(batch);
       assert(batchIndex !== -1, 'The batch should be in the BatchQueue');
-      this.batchQueue.splice(batchIndex, 1);
+      batchQueue.splice(batchIndex, 1);
 
       this.sendReadyBatches();
     });
   }
 
+  private wrapRetry(
```
schmidt-sebastian, line 936: Please add a method comment.

dev/src/bulk-writer.ts
```diff
@@ -740,27 +914,69 @@ export class BulkWriter {
...
+    promise: Promise<WriteResult>,
+    documentRef: firestore.DocumentReference,
+    retryCount: number,
```
schmidt-sebastian, line 939: Again, not sure what this means.

dev/src/bulk-writer.ts
```diff
@@ -740,27 +914,69 @@ export class BulkWriter {
...
       assert(batchIndex !== -1, 'The batch should be in the BatchQueue');
-      this.batchQueue.splice(batchIndex, 1);
+      batchQueue.splice(batchIndex, 1);
 
       this.sendReadyBatches();
     });
   }
 
+  private wrapRetry(
+    promise: Promise<WriteResult>,
```
schmidt-sebastian, line 937: What is this? Name is somewhat generic :)

dev/src/bulk-writer.ts
```diff
@@ -740,27 +914,69 @@ export class BulkWriter {
...
+          retryCount
+        );
+        const shouldRetry = this._errorFn(bulkWriterError);
+        this._pendingOps.delete(voidOperationPromise);
+        if (shouldRetry) {
+          return retryFn();
+        } else {
+          throw bulkWriterError;
+        }
+      });
```
schmidt-sebastian, line 966: Nit: This might be slightly prettier if we used try/catch with `async/await` here.

dev/src/bulk-writer.ts
```diff
@@ -740,27 +914,69 @@ export class BulkWriter {
...
       this.sendReadyBatches();
     });
   }
 
+  private wrapRetry(
+    promise: Promise<WriteResult>,
+    documentRef: firestore.DocumentReference,
+    retryCount: number,
+    operationType: 'create' | 'set' | 'update' | 'delete',
+    retryFn: () => Promise<WriteResult>
```
schmidt-sebastian, line 941: This is the original function isn't it? Could we just call it in this method ourselves, which would drop the `promise` argument.

schmidt-sebastian (commented):
Thanks for the update. We are pretty close. Feel free to ignore feedback that is contradicting :)

dev/src/bulk-writer.ts
thebrianchen: Couldn't think of a good way for naming, but added more detailed comments.

dev/src/bulk-writer.ts
```diff
@@ -326,34 +277,105 @@ class BulkCommitBatch {
...
+   * from this set before scheduling the next retry.
+   */
+  private _pendingOps: Set<Promise<void>> = new Set();
+
+  /**
+   * Whether this BulkWriter instance has started to close. Afterwards, no
+   * new operations can be enqueued, except for retry operations scheduled by
+   * the error handler.
+   */
+  private _closed = false;
```
schmidt-sebastian, line 349: s/_closed/_closing based on your comment.

dev/src/bulk-writer.ts
```diff
@@ -326,34 +277,105 @@ class BulkCommitBatch {
...
+    document: firestore.DocumentReference,
+    result: WriteResult
+  ) => void = () => {};
 
+  /**
+   * The user-provided callback to be run every time a BulkWriter operation
+   * fails.
+   */
+  private _errorFn: (error: BulkWriterError) => boolean = error => {
+    const retryCodes = getRetryCodes('batchWrite');
```
schmidt-sebastian, line 370: Regarding logging since the comment thread doesn't show up:

I think we need to add logging, otherwise we have no insight into what happens and why writes get re-scheduled. I am not worried about this being noisy, but we should log to the log function and not to `console.log`.

Note that a user-provided error handler would also overwrite this.

dev/src/bulk-writer.ts
```diff
@@ -582,15 +600,63 @@ export class BulkWriter {
...
+   * @param shouldRetryCallback A callback to be called every time a BulkWriter
+   * operation fails. Returning `true` will retry the operation. Returning
+   * `false` will stop the retry loop.
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   *
+   * bulkWriter
+   *   .onWriteError((error) => {
+   *     if (
+   *       shouldRetry(error.code) &&
```
schmidt-sebastian, line 648: Can you just use some hardcoded error check (`error.code === GrpcStatus.UNAVAILABLE`) in this example?

dev/src/bulk-writer.ts
```diff
@@ -618,18 +684,55 @@ export class BulkWriter {
...
-    const trackedBatches = this.batchQueue;
-    const writePromises = trackedBatches.map(batch => batch.awaitBulkCommit());
-    this.sendReadyBatches();
-    await Promise.all(writePromises);
+
+    // Copy the pending ops at the time flush() was called.
+    const pendingOps: Array<Promise<void>> = [];
+    this._pendingOps.forEach(op => pendingOps.push(op));
+
+    return this._flush(pendingOps);
```
schmidt-sebastian, line 694:
```suggestion
    return this._flush(this._pendingOps.slice());
```

dev/src/bulk-writer.ts
```diff
@@ -738,29 +849,124 @@ export class BulkWriter {
...
   }
 
+  /**
+   * Schedules and runs the provided operation.
+   *
+   * @param documentRef
+   * @param operationType
+   * @param operationFn
+   */
+  private runOperation(
```
schmidt-sebastian, line 889: s/runOperation/_executeWrite ? I think using "write" makes this slightly more descriptive.

dev/src/bulk-writer.ts
```diff
@@ -618,18 +684,55 @@ export class BulkWriter {
...
+    while (!retriesCompleted) {
+      const before = this._pendingBatches.size;
+      if (isFirstLoop) {
+        this.sendReadyBatches();
+        isFirstLoop = false;
+      } else {
+        this._retryBatchQueue.forEach(batch => batch.markReadyToSend());
+        this.sendReadyBatches(this._retryBatchQueue);
+      }
+      const after = this._pendingBatches.size;
```
schmidt-sebastian, line 716: private async _flush(pendingOps: Array<Promise<void>>): Promise<void> {
```suggestion
  private async _flush(pendingOps: Array<Promise<void>>): Promise<void> {
    let batchQueue = this._batchQueue;

    // Send all scheduled operations on the BatchQueue first. Afterwards,
    // send all retries that the initial batches generated in a loop. This way,
    // operations scheduled after flush() will not be sent until the retries
    // are completed.
    while (batchQueue.length > 0) {
      batchQueue.forEach(batch => batch.markReadyToSend());
      this.sendReadyBatches(batchQueue);
      await Promise.all(this._pendingBatches);
      batchQueue = this._retryBatchQueue;
    }

    // Make sure user promises resolve before flush() resolves.
    return voidPromiseNoThrow(Promise.all(pendingOps));
  }
```

How about this? (untested)

dev/src/bulk-writer.ts
```diff
@@ -668,68 +774,73 @@ export class BulkWriter {
...
       assert(delayMs !== -1, 'Batch size should be under capacity');
       if (delayMs === 0) {
-        this.sendBatch(batch);
+        this.sendBatch(batch, batchQueue, batchCompletedDeferred);
       } else {
-        delayExecution(() => this.sendReadyBatches(), delayMs);
+        delayExecution(() => {
+          batchCompletedDeferred.resolve();
+          this._pendingBatches.delete(batchCompletedDeferred.promise);
+          this.sendReadyBatches();
```
schmidt-sebastian, line 840: I forgot to mention this last time - I think this would be more future proof if `batchCompleteDeferred` was the last statement in this block.

dev/src/bulk-writer.ts
```diff
@@ -668,68 +774,73 @@ export class BulkWriter {
...
     while (
-      index < unsentBatches.length &&
-      unsentBatches[index].state === BatchState.READY_TO_SEND
+      index < batchQueue.length &&
+      batchQueue[index].state === BatchState.READY_TO_SEND
     ) {
-      const batch = unsentBatches[index];
+      const batch = batchQueue[index];
+
+      // Deferred promise that resolves when the current batch completes.
```
schmidt-sebastian, line 826: Maybe insert "or it's scheduling attempt"

dev/src/bulk-writer.ts
```diff
@@ -738,29 +849,124 @@ export class BulkWriter {
...
       assert(batchIndex !== -1, 'The batch should be in the BatchQueue');
-      this.batchQueue.splice(batchIndex, 1);
+      batchQueue.splice(batchIndex, 1);
+
+      batchCompletedDeferred.resolve();
+      this._pendingBatches.delete(batchCompletedDeferred.promise);
 
-      this.sendReadyBatches();
+      if (batchQueue === this._retryBatchQueue) {
+        this._retryBatchQueue.forEach(batch => batch.markReadyToSend());
```
schmidt-sebastian, line 873: Who is going to send these batches?  Don't we need to schedule a commit here so that the retries work even without `flush()`? If we do so, we might not need to retry in `_flush()`.

dev/src/bulk-writer.ts
```diff
@@ -738,29 +849,124 @@ export class BulkWriter {
...
+          documentRef,
+          operationType,
+          operationFn,
+          failedAttempts + 1
+        );
+      } else {
+        throw bulkWriterError;
+      }
+    }
+  }
```
schmidt-sebastian, line 960: Simplification suggestion:

```suggestion
private async runOperation(
    documentRef: firestore.DocumentReference,
    operationType: 'create' | 'set' | 'update' | 'delete',
    operationFn: (bulkCommitBatch: BulkCommitBatch) => Promise<WriteResult>
  ): Promise<WriteResult> {
    this.verifyNotClosed();
    this.sendReadyBatches();
    
    for (let failedAttempts = 0; ; ++failedAttempts) {
      // A deferred promise that resolves when operationFn completes.
      const operationCompletedDeferred = new Deferred<void>();
      this._pendingOps.add(operationCompletedDeferred.promise);
      
      const batchQueue =
          failedAttempts > 0 ? this._retryBatchQueue : this._batchQueue;
      const bulkCommitBatch = this.getEligibleBatch(batchQueue);

      try {
        const operationResult = await operationFn(bulkCommitBatch);
        this._successFn(documentRef, operationResult);
        return operationResult;
      } catch (error) {
        const bulkWriterError = new BulkWriterError(
            error.code,
            error.message,
            documentRef,
            operationType,
            failedAttempts
        );

        const shouldRetry = this._errorFn(bulkWriterError);
        if (!shouldRetry) {
          throw bulkWriterError;
        }
      } finally {
        this._pendingOps.delete(operationCompletedDeferred.promise);
      }
    }
  }
```

dev/system-test/firestore.ts
```diff
@@ -2612,6 +2613,26 @@ describe('BulkWriter class', () => {
...
+    let code: Status = -1;
+    writer.onWriteError(error => {
+      retryCount = error.failedAttempts;
+      return error.failedAttempts < 15;
+    });
+
+    // Use an invalid document name that the backend will reject.
+    const ref = randomCol.doc('__doc__');
+    writer.create(ref, {foo: 'bar'}).catch(err => {
+      console.log('error running');
```
schmidt-sebastian, line 2628: Please remove

dev/system-test/firestore.ts
```diff
@@ -2612,6 +2613,26 @@ describe('BulkWriter class', () => {
...
+      code = err.code;
+    });
+    await writer.close();
+    expect(retryCount).to.equal(15);
+    console.log('validating');
```
schmidt-sebastian, line 2633: Please remove

types/firestore.d.ts
```diff
@@ -474,8 +474,9 @@ declare namespace FirebaseFirestore {
      * @param documentRef A reference to the document to be
      * created.
      * @param data The object to serialize as the document.
-     * @returns A promise that resolves with the result
-     * of the write. Throws an error if the write fails.
+     * @returns {Promise<WriteResult>} A promise that resolves with the result
```
schmidt-sebastian, line 477: No need to include the return type in TypeScript. You should keep the type of the rejected result, since that is not apparent here.

dev/src/bulk-writer.ts
thebrianchen: Changed to power all retry batch scheduling to `sendBatch()`. Now, `flush()` schedules writes on the batchQueue, then kicks off the retryBatch if there are retries. Subsequent retries are now handled when the first retry batch comes back.

Also, I moved the `markReadyToSend()` mapping for retryBatches into `sendReadyBatches()`. There's two call points (flush and sendBatch) that would require a `forEach()` call, and I got bitten several times by forgetting to mark the batches as ready to send.

dev/src/bulk-writer.ts
```diff
@@ -583,14 +613,63 @@ export class BulkWriter {
...
+   * @param shouldRetryCallback A callback to be called every time a BulkWriter
+   * operation fails. Returning `true` will retry the operation. Returning
+   * `false` will stop the retry loop.
+   * @example
+   * let bulkWriter = firestore.bulkWriter();
+   *
+   * bulkWriter
+   *   .onWriteError((error) => {
+   *     if (
+   *       (error.code === GrpcStatus.UNAVAILABLE &&
```
schmidt-sebastian, line 661:
```suggestion
   *       error.code === GrpcStatus.UNAVAILABLE &&
```

dev/src/bulk-writer.ts
```diff
@@ -336,24 +315,76 @@ export class BulkWriter {
...
+   * The user-provided callback to be run every time a BulkWriter operation
+   * fails.
+   */
+  private _errorFn: (error: BulkWriterError) => boolean = error => {
+    const retryCodes = getRetryCodes('batchWrite');
+    const shouldRetry =
+      error.code !== undefined &&
+      retryCodes.includes(error.code) &&
+      error.failedAttempts < MAX_RETRY_ATTEMPTS;
+    logger(
```
schmidt-sebastian, line 376: As proposed in API meeting, we should move this logging to the callsite. This will also allow to us to log user-customized retries.

dev/src/bulk-writer.ts
```diff
@@ -618,18 +697,38 @@ export class BulkWriter {
...
+    this.sendReadyBatches(batchQueue);
+    await Promise.all(this._pendingBatches);
+
+    // Afterwards, send all accumulated retry operations. Wait until the
+    // retryBatchQueue is cleared. This way, operations scheduled after
+    // flush() will not be sent until the retries are completed.
+    batchQueue = this._retryBatchQueue;
+    if (batchQueue.length > 0) {
+      this.sendReadyBatches(batchQueue);
+    }
```
schmidt-sebastian, line 721: I wonder if this logic can be combined. The logic for the two different lists is essentially the same.

dev/src/bulk-writer.ts
```diff
@@ -668,68 +770,79 @@ export class BulkWriter {
...
    */
-  private sendReadyBatches(): void {
-    const unsentBatches = this.batchQueue.filter(
-      batch => batch.state === BatchState.READY_TO_SEND
-    );
+  private sendReadyBatches(batchQueue: BulkCommitBatch[]): void {
+    // Always mark retry batches as READY_TO_SEND.
+    if (batchQueue === this._retryBatchQueue) {
+      batchQueue.forEach(batch => batch.markReadyToSend());
+    }
```
schmidt-sebastian, line 818: This seems out of place here. Can we do this at the callsite?

dev/src/bulk-writer.ts
```diff
@@ -738,29 +851,85 @@ export class BulkWriter {
...
+          const shouldRetry = this._errorFn(bulkWriterError);
+          if (!shouldRetry) {
+            throw bulkWriterError;
+          }
+        }
+      }
+    } finally {
+      operationCompletedDeferred.resolve();
+      this._pendingOps.delete(operationCompletedDeferred.promise);
+    }
```
schmidt-sebastian, line 922: I think the double try/catch block is a bit harder to parse then just calling these two lines twice.

dev/test/bulk-writer.ts
```diff
@@ -372,12 +376,35 @@ describe('BulkWriter', () => {
...
     return bulkWriter.close().then(async () => verifyOpCount(1));
   });
 
+  it('throws UnhandledPromiseRejection if the error is not caught', async () => {
+    let errorThrown = false;
+    const unhandledDeferred = new Deferred<void>();
+    process.on('unhandledRejection', () => {
+      errorThrown = true;
+      unhandledDeferred.resolve();
+    });
```
schmidt-sebastian, line 392: Let's confirm with Yoshi team whether it would be ok to swallow these.

schmidt-sebastian (commented):
This is pretty good. I think we should verify some of the assumptions we made at API discussion, but we can probably take this back to the original PR

dev/test/bulk-writer.ts
thebrianchen: SG, started a conversation with them in the node room.

schmidt-sebastian:
Closing this as well are moving back to the original PR.

PR #1344, opened 2020-10-24
fix: speed up listDocuments pagination

### LINKED ISSUE #1342: listDocuments making too many requests
#### Environment details

  - OS: MacOS
  - Node.js version: 10
  - npm version: 6.14.8
  - `@google-cloud/firestore` version: 4.1.1

#### Steps to reproduce

We changed part of our code to use `listDocuments` instead of doing a `get` on the collection, as we are only interested in the ids. However we are seeing a considerable increase in the latency to get the results.

The weird part is that the `listDocuments` request seems to be retried/paginated? thousand of times, as shown by the OpenTelemetry data, where there are 1239 grpc calls to `grpc.google.firestore.v1.Firestore/ListDocuments`.

Screen Shot 2020-10-23 at 13 10 42

Waterfall showing the nested requests
Screen Shot 2020-10-23 at 13 11 27

Latency increased considerably
Screen Shot 2020-10-23 at 13 10 14

The code we are using is:

```js
async generateCode() {
    const codes = await getInviteCodes();
    let code = generateRandomCode();
    while (codes.has(code)) {
      code = generateRandomCode();
    }
}

async function getCodes() {
    const codeRefs = await db.collection('codes').listDocuments();
    return new Set(codeRefs.map((ref) => ref.id));
}
```

Note that all requests return with `grpc.status_code`: 0.

We are using `listDocuments` in other parts of our codebase (for example to recursively delete users) and we checked the behavior and we did not find a high number of calls to this grpc endpoint.

Any idea what could be triggering that many retries or how to disable them? Or is it that the endpoint is paginating the results? However the `listDocuments` call sets the page size to `2**32-1`, maybe that value is too high and the backend is using the default instead.

sk-:
A page size of 2**32 - 1 is rejected by the backend, making it to use the default page size of 20, which increases the latency of the `listDocuments` call.

Fixes #1342 🦕


sk-:
No idea what to do about code coverage, as nothing really changed there, only an extra comment line was added.

schmidt-sebastian:
> No idea what to do about code coverage, as nothing really changed there, only an extra comment line was added.

Don't worry about the coverage check. I will take a look at the PR today.

schmidt-sebastian (commented):
LGTM. I will reach out to the backend team to see if we can increase the page size limit.

schmidt-sebastian (approved):

wimpykid719:
Hi!
Now I'm writing scripts for delete Firestore data.
and I had stack for memory leak.
Because `listDocuments()` makes 100000 of array data.
I found this issues.
Is it possible to pagination `listDocuments()`?

schmidt-sebastian:
@wimpykid719 Are you able to use "Firestore.recursiveDelete()", which deletes documents in smaller batches? If that is not possible, you can use the Query API and combine it with limits or `startAt()`.

wimpykid719:
Yeah, I want to do that but It's so deep path and I'm stacking the middle of path.

`/collection/document/collection/document/collection/document/collection/documents(The id is like this 20210930114547 this is almost random)/collection/documents(random ID)/field
`

I'm trying to get documents id using `listDocuments()` for making delete path, but this documents return to so many array that makes memory leak.
and I think can't use Query API and `limits()`, `startAt()` because that documents doesn't have field.
Do you have any ideas? Please🙇🏻‍♂️

schmidt-sebastian:
You can also use https://github.com/googleapis/nodejs-firestore/blob/main/dev/src/v1/firestore_client.ts#L1343 directly, which lets you control the output size.

PR #1347, opened 2020-10-28
fix: retry transactions that fail with expired transaction IDs

schmidt-sebastian:
Fixes b/171204289

The original issue is that once a transaction expires, the backend returns INVALID_ARGUMENT. While INVALID_ARGUMENT is not generally retryable, we want to retry INVALID_ARGUMENT for expired transactions only. After talking with the backend TL, we decided that we can only do this based on the error message and if we have previously received a transaction ID.

This PR also changes the Rollback RPC to use the same retry semantics.


wilhuff (approved):
LGTM

PR #1348, opened 2020-11-02
docs(sample): normalize sample's region tags

laljikanjareeya:
Make sample's region tags consistent across all the languages, [another PR](https://github.com/firebase/snippets-node/pull/170) for other firestore samples.
//cc @crwilcox

### DIFF
--- a/samples/solution-counters.js
+++ b/samples/solution-counters.js
@@ -17,14 +17,17 @@ const {Firestore, FieldValue} = require('@google-cloud/firestore');
 
 async function main() {
   // [START increment_counter]
+  // [START firestore_solution_sharded_counter_increment]
   function incrementCounter(docRef, numShards) {
     const shardId = Math.floor(Math.random() * numShards);
     const shardRef = docRef.collection('shards').doc(shardId.toString());
     return shardRef.set({count: FieldValue.increment(1)}, {merge: true});
   }
+  // [END firestore_solution_sharded_counter_increment]
   // [END increment_counter]
 
   // [START get_count]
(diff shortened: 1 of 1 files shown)


samples/solution-counters.js
```diff
@@ -16,15 +16,15 @@
 const {Firestore, FieldValue} = require('@google-cloud/firestore');
 
 async function main() {
-  // [START increment_counter]
+  // [START firestore_solution_sharded_counter_increment]
```
crwilcox, line 19: similar to the other document, can we keep boht tags for a bit? this aids the transition. 

```
// [START add_lovelace]
// [START firestore_setup_dataset_pt1]


/* SAMPLE CODE */

// [END firestore_setup_dataset_pt1]
// [END add_lovelace]
```

crwilcox (changes requested):

samples/solution-counters.js
laljikanjareeya: Done!!

tmatsuo:
@laljikanjareeya 

Let me add some explanation on the snippet-bot comment:
> There are 3 possible violations for not having product prefix.

The bot thinks they are violations because the current data shows the `region_tag_prefix` should be `fs` for Firestore samples. However, it might be possible that the bot is using stale data. 

If you're sure the region tag prefix is `firestore`, just go ahead and merge. I'll appreciate if you file a data quality issue at
https://github.com/googleapis/repo-automation-bots/issues in that case.

Thanks,

laljikanjareeya:
Gentle ping @crwilcox

crwilcox (approved):

PR #1352, opened 2020-11-03
fix: create new batch for writes to the same doc

thebrianchen:
Fixing BulkWriter to add new writes to the same document into a new batch, since the BE endpoint validates that there are no duplicate docs in each batch.

(BulkCommitBatch refactoring coming next)

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -2607,7 +2607,7 @@ describe('BulkWriter class', () => {
     expect(deleteResult.writeTime).to.deep.equal(new Timestamp(0, 0));
   });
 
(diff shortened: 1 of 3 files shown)


schmidt-sebastian (approved):

dev/system-test/firestore.ts
```diff
@@ -2607,6 +2607,19 @@ describe('BulkWriter class', () => {
     expect(deleteResult.writeTime).to.deep.equal(new Timestamp(0, 0));
   });
 
+  it('has can write to the same document twice', async () => {
```
schmidt-sebastian, line 2610: Hm, my amazing Meme doesn't seem to show up, but please drop "has".

PR #1354, opened 2020-11-04
chore: refactor BulkCommitBatch

thebrianchen:
First commit refactored out `BulkCommitBatch` into `WriteBatch`, but it resulted in a bunch of one-off methods in `WriteBatch` that were only used to support `BulkWriter`. 

The second commit refactors `BulkCommitBatch` to extend `WriteBatch`, which allows for the BulkWriter-related methods to be separate from `WriteBatch`. I also refactored `_commit()` to be the helper method that makes the request and returns the response.


dev/test/write-batch.ts
```diff
@@ -503,21 +503,25 @@ describe('bulkCommit support', () => {
...
-    writeBatch.set(documentName, {foo: FieldValue.serverTimestamp()});
-    writeBatch.update(documentName, {foo: 'bar'});
-
-    return writeBatch.bulkCommit().then(resp => {
-      verifyResponse(resp);
+    const doc1 = firestore.doc('col/doc1');
+    const doc2 = firestore.doc('col/doc2');
+    const completed: string[] = [];
+
+    writeBatch.set(doc1, {foo: FieldValue.serverTimestamp()});
```
thebrianchen, line 511: This feels awkward to use.

schmidt-sebastian (approved):

schmidt-sebastian (changes requested):
Undo LGTM, Undo Approval since I haven't actually looked yet:)

dev/src/bulk-writer.ts
```diff
@@ -73,124 +76,91 @@ const RATE_LIMITER_MULTIPLIER = 1.5;
...
-   */
-  set<T>(
-    documentRef: firestore.DocumentReference<T>,
-    data: T | Partial<T>,
-    options?: firestore.SetOptions
-  ): Promise<WriteResult> {
-    this.writeBatch.set(documentRef, data, options);
-    return this.processLastOperation(documentRef);
-  }
+  async bulkCommit(): Promise<void> {
```
schmidt-sebastian, line 113: s/bulkCommit/_bulkCommit

But really should this maybe just be `commit` and overwrite the `commit` method in the base class? That way, it hides the existing implementation which we would never want to use.

dev/src/bulk-writer.ts
```diff
@@ -73,124 +76,91 @@ const RATE_LIMITER_MULTIPLIER = 1.5;
...
+    return this.docPaths.has(documentRef.path);
   }
 
-  /**
-   * The number of writes in this batch.
-   */
-  get opCount(): number {
-    return this.pendingOps.length;
+  _markReadyToSend(): void {
+    if (this.state === BatchState.OPEN) {
```
schmidt-sebastian, line 100: We should try to unify this: Either `state` lives in BulkCommitBatch and the helpers are here, or we should move the helpers to WriteBatch as well. What do you think?

dev/src/write-batch.ts
```diff
@@ -603,25 +595,26 @@ export class WriteBatch implements firestore.WriteBatch {
...
     requestTag?: string;
-  }): Promise<WriteResult[]> {
+    retryCodes?: number[];
+    methodName?: FirestoreUnaryMethod;
+  }): Promise<Resp> {
     // Note: We don't call `verifyNotCommitted()` to allow for retries.
-    this._committed = true;
+    this.state = BatchState.SENT;
 
     const tag = (commitOptions && commitOptions.requestTag) || requestTag();
```
schmidt-sebastian, line 607:
```suggestion
    const tag = commitOptions?.requestTag ?? requestTag();
```

dev/src/write-batch.ts
```diff
@@ -603,25 +595,26 @@ export class WriteBatch implements firestore.WriteBatch {
...
+    const request = {
+      database: this._firestore.formattedName,
       writes: this._ops.map(op => op.op()),
+      transaction: commitOptions?.transactionId,
     };
-    if (commitOptions?.transactionId) {
-      request.transaction = commitOptions.transactionId;
+
+    if (request.transaction === undefined) {
+      delete request.transaction;
```
schmidt-sebastian, line 617: The old way seemed better. Just add the transaction if it is present. To make this typesafe, you can do something like this: https://paste.googleplex.com/5733759430164480

dev/src/bulk-writer.ts
```diff
@@ -73,124 +76,91 @@ const RATE_LIMITER_MULTIPLIER = 1.5;
...
 
-  private readonly backoff: ExponentialBackoff;
-
-  constructor(
-    private readonly firestore: Firestore,
-    private writeBatch: WriteBatch,
-    private readonly maxBatchSize: number
-  ) {
-    this.backoff = new ExponentialBackoff();
+  _has(documentRef: firestore.DocumentReference): boolean {
```
schmidt-sebastian, line 95: This is a private class and it doesn't need underscores. I don't mind if you leave them for consistency.

thebrianchen: Removed.

dev/src/bulk-writer.ts
thebrianchen: It's a little messy right now, since the WriteBatch uses the `BatchState.SENT` after the batch has been committed, but does not adhere to the READY_TO_SEND checks.

We have two options:
1. Move all the BatchState code into BulkCommitBatch and add back a `committed` boolean flag in WriteBatch. (more code, appropriate separation)
2. Move all the BatchState over into WriteBatch. (slightly less code, less-intuitive behavior)

dev/src/bulk-writer.ts
thebrianchen: Why are we adding the underscore for this method, but not the others?

I originally wanted to just override the base class, but `WriteBatch.commit()` returns `Promise<WriteResult[]>`, which doesn't match the signature here.

dev/src/write-batch.ts
thebrianchen: done.

dev/src/write-batch.ts
thebrianchen: done. Added comment and removed cast to `unknown` below.

dev/src/bulk-writer.ts
schmidt-sebastian: I thought this was a public class. It's not, so we don't need to add it. Thanks for pointing out the different return types. I think it might be best to leave as is then.

dev/src/bulk-writer.ts
thebrianchen: I opted to move everything back into BulkCommitBatch, which only results in an extra `_commited` variable on WriteBatch. The main issue I had with option 2 is that the helpers that I moved over are only used by BulkWriter, which didn't seem right.

dev/src/write-batch.ts
```diff
@@ -603,23 +580,25 @@ export class WriteBatch implements firestore.WriteBatch {
    * this request.
    * @returns  A Promise that resolves when this batch completes.
    */
-  async _commit(commitOptions?: {
+  async _commit<Req, Resp>(commitOptions?: {
```
schmidt-sebastian, line 583: This should be `Req extends api.ICommitRequest`, otherwise your code below isn't really typesafe and you are enforcing `api.ICommitRequest` only when you create the request. Let me know if that adds any type issues (I believe it didn't for me).

thebrianchen: Good catch, thanks!

schmidt-sebastian (approved):
Shiiiiip it!

PR #1355, opened 2020-11-04
feat: add ability to specify custom headers for individual RPC types

schmidt-sebastian:
This makes it possible to specify custom headers for a single RPC type as such:

```
 firestore = new Firestore({commit: {customHeaders: {'X-Header: 'Value'}}});
```


bcoe (approved):
A test would be nice, but can understand wanting to get this out to test.

PR #1363, opened 2020-11-10
fix: ignore 'undefined' in update() with UpdateMap

### LINKED ISSUE #1359: Update throws error with top level undefined when ignoreUndefinedProperties is set
Calling `.update()` with a data object that contains `undefined` in the top level fields is throwing an error when `ignoreUndefinedProperties` is set.

I tracked it to the `validateFieldValue` function on `write-batch.ts` that calls `validateUserInput` without a `level` parameter to inform that the value being validated is already 1 level deep.

https://github.com/googleapis/nodejs-firestore/blob/dc94946f261db527f913d1ff89a2572cd9539f70/dev/src/write-batch.ts#L841-L854

I also noticed that the test that should catch this doesn't test for a top level field with `undefined`.

https://github.com/googleapis/nodejs-firestore/blob/dc94946f261db527f913d1ff89a2572cd9539f70/dev/test/ignore-undefined.ts#L90-L115

#### Code to reproduce
```typescript
const firestore = new Firestore({
  ignoreUndefinedProperties: true,
});

const col = firestore.collection('collection');
const doc = col.doc('uid');

await doc.update({
  required: 'string',
  optional: undefined,
});
```

#### Environment details
  - OS: Ubuntu 20.04 on WSL 2 on Windows 10 19041.572
  - `node` version: 14.14.0
  - `yarn` version: 1.22.5
  - `@google-cloud/firestore` version: 4.7.0

schmidt-sebastian:
When `ignoreUndefinedProperties` is set, `docRef.update({"foo": undefined})` should not be an error. This PR fixes this by passing the correct level to the validator, and then updating the validator to still work for deletes. 

Fixes https://github.com/googleapis/nodejs-firestore/issues/1359

### DIFF
--- a/dev/src/write-batch.ts
+++ b/dev/src/write-batch.ts
@@ -45,9 +45,8 @@ import {
   validateMinNumberOfArguments,
   validateOptional,
 } from './validate';
-
-import api = google.firestore.v1;
 import {GoogleError, Status} from 'google-gax';
+import api = google.firestore.v1;
 
 /**
  * A WriteResult wraps the write time set by the Firestore servers on sets(),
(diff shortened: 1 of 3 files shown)


dev/src/write-batch.ts
```diff
@@ -900,16 +903,8 @@ function validateUpdateMap(
...
-      isEmpty = false;
-      validateFieldValue(arg, obj[prop], allowUndefined, new FieldPath(prop));
-    }
-  }
-
-  if (isEmpty) {
+  if (Object.keys(obj).length === 0) {
     throw new Error('At least one field must be updated.');
   }
+  validateFieldValue(arg, obj, allowUndefined);
```
thebrianchen, line 909: Can you explain why we no longer validate each field on the object anymore?

thebrianchen (approved):

dev/src/write-batch.ts
schmidt-sebastian: We do. We loop through the object here - https://github.com/googleapis/nodejs-firestore/blob/56355f1c83b055fdbd3ed47e069df813696a1fc0/dev/src/serializer.ts#L334

PR #1365, opened 2020-11-16
feat: Release Bundles

wu-hui:
PR to release bundles. Will merge when we decide to release.

### DIFF
--- a/dev/src/bundle.ts
+++ b/dev/src/bundle.ts
@@ -30,8 +30,6 @@ const BUNDLE_VERSION = 1;
 
 /**
  * Builds a Firestore data bundle with results from the given document and query snapshots.
- *
- * @private
  */
 export class BundleBuilder {
   // Resulting documents for the bundle, keyed by full document path.
--- a/dev/src/bundle.ts
+++ b/dev/src/bundle.ts
@@ -40,7 +40,7 @@ export class BundleBuilder {
   // The latest read time among all bundled documents and queries.
   private latestReadTime = new Timestamp(0, 0);
 
-  constructor(public readonly bundleId: string) {}
+  constructor(readonly bundleId: string) {}
 
   add(documentSnapshot: DocumentSnapshot): BundleBuilder;
   add(queryName: string, querySnapshot: QuerySnapshot): BundleBuilder;
--- a/dev/src/bundle.ts
+++ b/dev/src/bundle.ts
@@ -40,7 +40,7 @@ export class BundleBuilder {
   // The latest read time among all bundled documents and queries.
   private latestReadTime = new Timestamp(0, 0);
 
-  constructor(private bundleId: string) {}
+  constructor(public readonly bundleId: string) {}
 
   add(documentSnapshot: DocumentSnapshot): BundleBuilder;
   add(queryName: string, querySnapshot: QuerySnapshot): BundleBuilder;
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -1920,6 +1920,9 @@ declare namespace FirebaseFirestore {
    * Builds a Firestore data bundle with results from the given document and query snapshots.
    */
   export class BundleBuilder {
+    /** The ID of this bundle. */
+    readonly bundleId: string;
+
     /**
      * Adds a Firestore document snapshot or query snapshot to the bundle.
      * Both the documents data and the query read time will be included in the bundle.
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -1924,7 +1924,7 @@ declare namespace FirebaseFirestore {
     readonly bundleId: string;
 
     /**
-     * Adds a Firestore document snapshot to the bundle. Both the documents data and the document
+     * Adds a Firestore `DocumentSnapshot` to the bundle. Both the documents data and the document
      * read time will be included in the bundle.
      *
      * @param documentSnapshot A `DocumentSnapshot` to add.
@@ -1933,8 +1933,8 @@ declare namespace FirebaseFirestore {
     add<T>(documentSnapshot: DocumentSnapshot<T>): BundleBuilder;
 
     /**
-     * Adds a Firestore query snapshot to the bundle. Both the query data and the query read time
-     * will be included in the bundle.
+     * Adds a Firestore `QuerySnapshot` to the bundle. Both the documents in the query results and
+     * the query read time will be included in the bundle.
      *
      * @param queryName The name of the query to add.
      * @param querySnapshot A `QuerySnapshot` to add to the bundle.
--- a/types/firestore.d.ts
+++ b/types/firestore.d.ts
@@ -301,7 +301,7 @@ declare namespace FirebaseFirestore {
      * Creates a new `BundleBuilder` instance to package selected Firestore data into
      * a bundle.
      *
-     * @param bundleId. The id of the bundle. When loaded on clients, client SDKs use this id
+     * @param bundleId The ID of the bundle. When loaded on clients, client SDKs use this ID
      * and the timestamp associated with the built bundle to tell if it has been loaded already.
      * If not specified, a random identifier will be used.
      *
@@ -316,7 +316,7 @@ declare namespace FirebaseFirestore {
      *                            .build()
      * // Save `bundleBuffer` to CDN or stream it to clients.
      */
-    bundle(name?: string): BundleBuilder;
+    bundle(bundleId?: string): BundleBuilder;
   }
 
   /**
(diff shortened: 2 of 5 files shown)


schmidt-sebastian (changes requested):
I would expect this PR to contain updates to types/firestore.d.ts. Can you double check what needs to be added there?

wu-hui:
> I would expect this PR to contain updates to types/firestore.d.ts. Can you double check what needs to be added there?

Added. Thanks for the reminder.

types/firestore.d.ts
```diff
@@ -296,6 +296,27 @@ declare namespace FirebaseFirestore {
      * behavior for the underlying BulkWriter.
      */
     bulkWriter(options?: BulkWriterOptions): BulkWriter;
+
+    /**
+     * Creates a new `BundleBuilder` instance to package selected Firestore data into
+     * a bundle.
+     *
+     * @param bundleId. The id of the bundle. When loaded on clients, client SDKs use this id
```
schmidt-sebastian, line 304: In the text part: s/id/ID here and everywhere (including the main sources)

Also drop the period after "bundleId".

types/firestore.d.ts
```diff
@@ -1895,6 +1916,34 @@ declare namespace FirebaseFirestore {
...
+  export class BundleBuilder {
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
+     *
+     * @param {DocumentSnapshot=} documentSnapshot A document snapshot to add.
+     * @returns {BundleBuilder} This instance.
+     */
+    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
+    /**
```
schmidt-sebastian, line 1931: Nit: Missing empty line.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,34 @@ declare namespace FirebaseFirestore {
...
+  /**
+   * Builds a Firestore data bundle with results from the given document and query snapshots.
+   */
+  export class BundleBuilder {
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
+     *
+     * @param {DocumentSnapshot=} documentSnapshot A document snapshot to add.
+     * @returns {BundleBuilder} This instance.
```
schmidt-sebastian, line 1928: This is TypeScript. You should drop the JSDoc type markers.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,34 @@ declare namespace FirebaseFirestore {
...
+     */
+    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
+     *
+     * @param {string=} queryName The name of the query to add.
+     * @param {QuerySnapshot=} querySnapshot A query snapshot to add to the bundle.
+     * @returns {BundleBuilder} This instance.
```
schmidt-sebastian, line 1937: Drop types.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,34 @@ declare namespace FirebaseFirestore {
...
+   */
+  export class BundleBuilder {
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
+     *
+     * @param {DocumentSnapshot=} documentSnapshot A document snapshot to add.
+     * @returns {BundleBuilder} This instance.
+     */
+    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
```
schmidt-sebastian, line 1930: Should `DocumentSnapshot` (and `QuerySnapshot`) below be generic (for converter support)? Does the BundleBuilder respect the converters as it generates the Proto?

wu-hui: All it does is call `toDocumentProto()` on the snapshot, I don't think it is necessary to make it generic?

types/firestore.d.ts
wu-hui: Done.

types/firestore.d.ts
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -42,7 +40,7 @@ export class BundleBuilder {
   // The latest read time among all bundled documents and queries.
   private latestReadTime = new Timestamp(0, 0);
 
-  constructor(private bundleId: string) {}
+  constructor(public readonly bundleId: string) {}
```
schmidt-sebastian, line 43: The TS style guide recommends using only "readonly" instead of "public readonly" as they are equivalent.

types/firestore.d.ts
```diff
@@ -296,6 +296,27 @@ declare namespace FirebaseFirestore {
      * behavior for the underlying BulkWriter.
      */
     bulkWriter(options?: BulkWriterOptions): BulkWriter;
+
+    /**
+     * Creates a new `BundleBuilder` instance to package selected Firestore data into
+     * a bundle.
+     *
+     * @param bundleId The ID of the bundle. When loaded on clients, client SDKs use this ID
+     * and the timestamp associated with the built bundle to tell if it has been loaded already.
```
schmidt-sebastian, line 305:
```suggestion
     * and the timestamp associated with the bundle to tell if it has been loaded already.
```

or:

```
     * and bundle's timestamp associated to tell if the bundle has been loaded already.
```

types/firestore.d.ts
```diff
@@ -1895,6 +1916,38 @@ declare namespace FirebaseFirestore {
...
+  /**
+   * Builds a Firestore data bundle with results from the given document and query snapshots.
+   */
+  export class BundleBuilder {
+    /** The ID of this bundle. */
+    readonly bundleId: string;
+
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
```
schmidt-sebastian, line 1928: This comment should only apply to DocumentSnapshots.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,38 @@ declare namespace FirebaseFirestore {
...
+     *
+     * @param documentSnapshot A document snapshot to add.
```
schmidt-sebastian, line 1930:
```suggestion
     * @param documentSnapshot A `DocumentSnapshot` to add.
```

types/firestore.d.ts
schmidt-sebastian: There is actually no such thing as a non-generic DocumentSnapshot in our API. The DocumentSnapshot type defaults to `DocumentSnapshot<DocumentData>`, which means that a custom type such as `DocumentSnapshot<Foo>` cannot be passed. This should be:

```
add<T>(documentSnapshot: DocumentSnapshot<T>): BundleBuilder;
```

Same for QuerySnapshot.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,38 @@ declare namespace FirebaseFirestore {
...
+     * @returns This instance.
+     */
+    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
+
+    /**
+     * Adds a Firestore document snapshot or query snapshot to the bundle.
+     * Both the documents data and the query read time will be included in the bundle.
```
schmidt-sebastian, line 1937: Please update to only refer to QuerySnapshots.

types/firestore.d.ts
```diff
@@ -1895,6 +1916,38 @@ declare namespace FirebaseFirestore {
...
+     *
+     * @param queryName The name of the query to add.
+     * @param querySnapshot A query snapshot to add to the bundle.
+     * @returns This instance.
+     */
+    add(queryName: string, querySnapshot: QuerySnapshot): BundleBuilder;
+
+    /**
+     * Builds the bundle into a `Buffer` instance.
```
schmidt-sebastian, line 1946: Maybe:

```suggestion
     * Builds the bundle and returns the result as a`Buffer` instance.
```

types/firestore.d.ts
wu-hui: Fixed.

types/firestore.d.ts
wu-hui: Done.

types/firestore.d.ts
wu-hui: Done, thanks.

types/firestore.d.ts
wu-hui: Done.

types/firestore.d.ts
wu-hui: Done.

types/firestore.d.ts
```diff
@@ -1924,26 +1924,26 @@ declare namespace FirebaseFirestore {
...
+     * @param documentSnapshot A `DocumentSnapshot` to add.
      * @returns This instance.
      */
-    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
+    add<T>(documentSnapshot: DocumentSnapshot<T>): BundleBuilder;
 
     /**
-     * Adds a Firestore document snapshot or query snapshot to the bundle.
-     * Both the documents data and the query read time will be included in the bundle.
+     * Adds a Firestore query snapshot to the bundle. Both the query data and the query read time
```
schmidt-sebastian, line 1936: We are also adding the resulting documents, no?

s/query snapshot/`QuerySnapshot`

types/firestore.d.ts
```diff
@@ -1924,26 +1924,26 @@ declare namespace FirebaseFirestore {
     readonly bundleId: string;
 
     /**
-     * Adds a Firestore document snapshot or query snapshot to the bundle.
-     * Both the documents data and the query read time will be included in the bundle.
+     * Adds a Firestore document snapshot to the bundle. Both the documents data and the document
```
schmidt-sebastian, line 1927: s/document snapshot/`DocumentSnapshot`

schmidt-sebastian (approved):

types/firestore.d.ts
wu-hui: Done.

types/firestore.d.ts
wu-hui: Done.

PR #1373, opened 2020-12-03
fix: stop using GRPC channels after RST_STREAM

schmidt-sebastian:
This PR is based on a theory (based on some anecdotal evidence) that once a GRPC Channel receives the first RST_STREAM error it will only every throw RST_STREAM errors. This PR changes our client pool to mark any GAX client that throws RST_STREAM as failed and ensures that the SDK spins up a new client for the next request.

Addresses https://github.com/googleapis/google-cloud-node/issues/7490

### DIFF
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {GoogleError} from 'google-gax';
 import * as assert from 'assert';
 
 import {logger} from './logger';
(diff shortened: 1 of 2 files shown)


dev/src/pool.ts
```diff
@@ -140,10 +149,19 @@ export class ClientPool<T> {
    * @private
    */
   private shouldGarbageCollectClient(client: T): boolean {
+    // Don't garbagae collect clients that have active requests.
```
alexander-fenster, line 152: nit: typo

alexander-fenster (approved):
LGTM with a minor nit

dev/src/pool.ts
schmidt-sebastian: Fixed

PR #1394, opened 2021-01-24
feat: add support for applying default converter in withConverter()

thebrianchen:


dev/src/collection-group.ts
```diff
@@ -168,18 +172,23 @@ export class CollectionGroup<T = firestore.DocumentData>
    *   post.someNonExistentProperty; // TS error
    * }
    *
-   * @param {FirestoreDataConverter} converter Converts objects to and from
-   * Firestore.
+   * @param {FirestoreDataConverter | null} converter Converts objects to and
```
dconeybe, line 175: ultranit: The "C" in "Converts" should be lowercase. Also in `reference.ts`.

dconeybe (commented):
Just one drive-by comment.

dev/src/collection-group.ts
thebrianchen: Thanks for catching!

thebrianchen: Actually, it should be capitalized, since it's the description of the `converter` parameter. The other docs descriptions capitalize the first letter as well.

dev/src/collection-group.ts
```diff
@@ -168,18 +172,23 @@ export class CollectionGroup<T = firestore.DocumentData>
...
+   * from Firestore. Passing in `null` applies the default `DocumentData` typed
+   * converter.
```
schmidt-sebastian, line 177: Maybe: `Passing in `null` removes the current converter.`

Up to you.

dev/src/collection-group.ts
```diff
@@ -168,18 +172,23 @@ export class CollectionGroup<T = firestore.DocumentData>
...
     converter: firestore.FirestoreDataConverter<U>
+  ): CollectionGroup<U>;
+  withConverter<U>(
+    converter: firestore.FirestoreDataConverter<U> | null
   ): CollectionGroup<U> {
     return new CollectionGroup<U>(
       this.firestore,
       this._queryOptions.collectionId,
-      converter
+      converter !== null ? converter : defaultConverter()
```
schmidt-sebastian, line 191: Since this is NodeJS and we don't care as much about code size, you could also use `converter ?? defaultConverter`.

dev/system-test/firestore.ts
```diff
@@ -148,6 +148,7 @@ describe('Firestore class', () => {
   });
 
   it('getAll() supports generics', async () => {
+    randomCol.doc('lo')._converter;
```
schmidt-sebastian, line 151: Hm.

dev/test/query.ts
```diff
@@ -723,6 +723,28 @@ describe('query interface', () => {
...
+    return createInstance(overrides).then(async firestore => {
+      const coll = await firestore
+        .collection('collectionId')
+        .withConverter(postConverter)
+        .withConverter(null);
+
+      const posts = await coll.where('title', '==', 'post').get();
+      expect(posts.size).to.equal(1);
+      // Check that the Post class's toString() override does not apply.
+      expect(posts.docs[0].data().toString()).to.equal('[object Object]');
```
schmidt-sebastian, line 745: Can we not do an `instanceof` check here?

schmidt-sebastian (commented):
Thanks. Some nits, some optional nits :)

We also need to update the types file.

dev/src/collection-group.ts
dconeybe: Ohhhh I'm a dummy. Sorry for the false alarm!

dev/test/query.ts
```diff
@@ -741,8 +742,7 @@ describe('query interface', () => {
 
       const posts = await coll.where('title', '==', 'post').get();
       expect(posts.size).to.equal(1);
-      // Check that the Post class's toString() override does not apply.
-      expect(posts.docs[0].data().toString()).to.equal('[object Object]');
+      expect(posts.docs[0].data() instanceof Post).to.be.false;
```
schmidt-sebastian, line 745: to.not.be.an.instanceof

thebrianchen: learn2chai

schmidt-sebastian (approved):

PR #1395, opened 2021-01-24
fix: support byte values in Bundles

wu-hui:


dev/src/bundle.ts
```diff
@@ -142,7 +142,11 @@ export class BundleBuilder {
   private elementToLengthPrefixedBuffer(
     bundleElement: firestore.IBundleElement
   ): Buffer {
-    const buffer = Buffer.from(JSON.stringify(bundleElement), 'utf-8');
+    // Convert to a valid proto message object then take its json representation.
+    // This take cares of stuff like converting internal byte array fields
+    // to base64 encodings.
+    const message = BundleElement.fromObject(bundleElement).toJSON();
+    const buffer = Buffer.from(JSON.stringify(message), 'utf-8');
```
schmidt-sebastian, line 149: Let's talk about this offline later, but this is quite an expensive operation. The steps are:

- Manually serialize documents in ProtobufJS format
- Read elements back Protobuf JS and convert to Proto3 JSON
- Parse Proto3 JSON into string

This seems like an okay "quick fix", but not the most elegant solution. We should figure out whether we are ok with this quick fix or whether we should directly produce Proto3 JSON. We could use a similar approach as the Web SDK and use different serialization options in our serializer depending on the target. While this might end up being a fair amount of work, it also helps us with the timestamp conversion.

Furthermore, can you verify whether the current conversion correctly translates numbers greater than 2^53 bits? You would have to turn on [useBigInt](https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L158) and then write a large integer value like such: `2147483647n`. I am asking because the JSON roundtrip might convert this type to a lossy floating point representation.

wu-hui: Added a test for BigInt.

dev/system-test/firestore.ts
```diff
@@ -2773,19 +2776,21 @@ describe('Bundle building', () => {
...
+        // TODO(wuandy): Fix query.toProto to skip undefined fields, so we can stop using `extend` here.
+        bundledQuery: extend(
+          true,
+          {},
+          {
+            parent: query.toProto().parent,
+            structuredQuery: query.toProto().structuredQuery,
+          }
+        ),
+      }).toJSON()
```
schmidt-sebastian, line 2792: Do we have to do this round-trip here (and below)? Can we not just compare against the resulting JSON directly?

wu-hui: The trick here is the expected object we use here will have int64 encoded as number, while the actual object we get from bundles have them encoded as strings, for example.

I have created some utility functions to help with the boilerplate. Unfortunately i cannot figure out how to make them generic, because `fromObject` is static, and returns different types.

dev/src/bundle.ts
```diff
@@ -41,7 +42,6 @@ export class BundleBuilder {
   private latestReadTime = new Timestamp(0, 0);
 
   constructor(readonly bundleId: string) {}
-
```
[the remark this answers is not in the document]
wu-hui: Done.

dev/system-test/firestore.ts
schmidt-sebastian: Is it not possible to manually create the desired Proto output? Would it be possible to write `expect(...).to.deep.equal({integerValue:'1337'}))` instead of `expect(...).to.deep.equal(fromJson({integerValue:1337}).toJSON()))`?

wu-hui: This works for the example you give. But the tests deal a lot with timestamps, and `expect(...).to.deep.equal(snap.readTime.toProto.timestampValue)` will compare encoded string `seconds` to the actual integer value. So you get tests failures like:

```
      -    "seconds": "1577836805"
      +    "seconds": 1577836805
```

dev/test/bundle.ts
```diff
@@ -37,17 +41,19 @@ export function verifyMetadata(
...
+    expect(parseInt(meta.totalBytes!.toString())).greaterThan(0);
   } else {
-    expect(meta.totalBytes).to.equal(0);
+    expect(parseInt(meta.totalBytes!.toString())).to.equal(0);
   }
   expect(meta.id).to.equal(TEST_BUNDLE_ID);
   expect(meta.version).to.equal(TEST_BUNDLE_VERSION);
   expect(meta.totalDocuments).to.equal(totalDocuments);
-  expect(meta.createTime).to.deep.equal(createTime);
+  expect(google.protobuf.Timestamp.fromObject(meta.createTime!)).to.deep.equal(
```
schmidt-sebastian, line 51: Fix for this: https://gist.github.com/schmidt-sebastian/3f53054d1e1c2d8238e473bfc133a1de

schmidt-sebastian (changes requested):
Please update the tests to not use the JSON helpers as it hides bugs in the JSON creation.

dev/test/bundle.ts
wu-hui: Done.

dev/src/bundle.ts
```diff
@@ -142,7 +143,11 @@ export class BundleBuilder {
   private elementToLengthPrefixedBuffer(
     bundleElement: firestore.IBundleElement
   ): Buffer {
-    const buffer = Buffer.from(JSON.stringify(bundleElement), 'utf-8');
+    // Convert to a valid proto message object then take its json representation.
```
schmidt-sebastian, line 146:
```suggestion
    // Convert to a valid proto message object then take its JSON representation.
```

dev/src/bundle.ts
```diff
@@ -142,7 +143,11 @@ export class BundleBuilder {
...
+    // This take cares of stuff like converting internal byte array fields
+    // to base64 encodings.
```
schmidt-sebastian, line 148:
```suggestion
    // to Base64 encodings.
```

dev/system-test/firestore.ts
```diff
@@ -43,12 +43,14 @@ import {
   postConverterMerge,
   verifyInstance,
 } from '../test/util/helpers';
-import IBundleElement = firestore.IBundleElement;
```
schmidt-sebastian, line 46: Thanks for re-ordering. Can you remove the empty line?

dev/test/bundle.ts
```diff
@@ -37,17 +38,17 @@ export function verifyMetadata(
   expectEmptyContent = false
 ): void {
   if (!expectEmptyContent) {
-    expect(meta.totalBytes).greaterThan(0);
+    expect(parseInt(meta.totalBytes!.toString())).greaterThan(0);
```
schmidt-sebastian, line 41: I think you can probably emit toString() here (totalBytes should be a string).

schmidt-sebastian (approved):

dev/system-test/firestore.ts
wu-hui: Done.

dev/test/bundle.ts
wu-hui: It's there to silent the type error, it can be number or string.

JustinBeckwith (approved):
Rubber stamp because I broke CODEOWNERs

PR #1396, opened 2021-01-26
fix: handle ignoreUndefinedProperties in set(merge: true)

### LINKED ISSUE #1392: ignoreUndefinedProperties is misleading
In the past I did remove myself undefined properties before inserting data into Firestore. Now, I use the _ignoreUndefinedProperties_ setting from the Firestore constructor instead. I found out that if I merge a document with undefined properties, the existing properties in Firestore corresponding to the undefined ones are deleted.

I don't know if it's an issue or the intended behavior, but I think this settings name is misleading since undefined properties are not ignored.

#### Environment details

  - OS: Debian
  - Node.js version: v14.7.0
  - npm version: 6.14.7
  - `@google-cloud/firestore` version: 4.8.1

#### Steps to reproduce

```
import { Firestore } from "@google-cloud/firestore";

const db = new Firestore({
    ignoreUndefinedProperties: true
});

db.collection('collection')
    .doc('doc')
    .set({
        prop1: "prop1",
        prop2: "prop2"
    }, { merge: true });

// In Firestore we have prop1 and prop2

db.collection('collection')
    .doc('doc')
    .set({
        prop1: "prop1",
        prop2: undefined
    }, { merge: true });

// In Firestore we have only prop1 (prop2 is deleted)

```

Thanks!

wilhuff:
Previously any field in the document would be propagated into the DocumentMask regardless of whether or not it had an undefined value, which led to the client acting as if the user had passed `FieldValue.delete()`, which wasn't intended.

Fixes #1392

### DIFF
--- a/dev/src/document.ts
+++ b/dev/src/document.ts
@@ -694,7 +694,11 @@ export class DocumentMask {
           }
         } else if (isPlainObject(value)) {
           extractFieldPaths(value, childPath);
-        } else {
+        } else if (value !== undefined) {
+          // If the value is undefined it can never participate in the document
+          // mask. With `ignoreUndefinedProperties` set to false,
+          // `validateDocumentData` will reject an undefined value before even
+          // computing the document mask.
           fieldPaths.push(childPath);
         }
       }
--- a/dev/test/ignore-undefined.ts
+++ b/dev/test/ignore-undefined.ts
@@ -64,6 +64,33 @@ describe('ignores undefined values', () => {
     );
   });
 
+  it('in set({ merge: true })', () => {
+    const overrides: ApiOverride = {
+      commit: request => {
+        requestEquals(
+          request,
+          set({
+            document: document('documentId', 'foo', 'foo'),
+            mask: updateMask('foo'),
+          })
+        );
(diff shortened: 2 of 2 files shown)


dev/src/document.ts
```diff
@@ -694,7 +694,12 @@ export class DocumentMask {
           }
         } else if (isPlainObject(value)) {
           extractFieldPaths(value, childPath);
-        } else {
+        } else if (value !== undefined) {
+          // If the value is undefined it can never participate in the document
+          // mask. With ignoreUndefinedProperties set to true, this shouldn't
+          // affect the document mask. With ignoreUndefinedProperties set to
+          // false, we'll throw an error when trying to create a proto from the
```
schmidt-sebastian, line 701: Nit: This error is thrown first by `validateDocumentData`: "Cannot use "undefined" as a Firestore value (found in field "bar"). If you want to ignore undefined values, enable `ignoreUndefinedProperties`."

schmidt-sebastian (approved):

dev/src/document.ts
wilhuff: Fixed.

PR #1401, opened 2021-01-29
fix: reject select and offset for bundles.

wu-hui:


dev/src/bundle.ts
```diff
@@ -118,6 +118,12 @@ export class BundleBuilder {
     if (this.namedQueries.has(name)) {
       throw new Error(`Query name conflict: ${name} has already been added.`);
     }
+    if (querySnap.query._hasSelect()) {
+      throw new Error('Bundled query cannot have projections (select()).');
+    }
+    if (querySnap.query._hasOffset()) {
+      throw new Error('Bundled query cannot have offset().');
+    }
```
schmidt-sebastian, line 126: This is a limitation on the Mobile SDKs, and as such I think we should perform the validation there. It will allow to us gradually lift the limitation if we ever end up supporting these features, as we can simple enable it for the Web SDK for example, while iOS and Android might still prohibit these queries. If we perform this validation on the Server SDKs, we have to finish implementing it in all Mobile SDKs and then remove this from 7 Server SDKs before this becomes useable. And it still won't work for developers that didn't update their Mobile SDK...

wu-hui: This applies only to `offset`?

schmidt-sebastian: This applies to all features I think.

PR #1404, opened 2021-02-01
docs: Add callback details for updateFunction in runTransaction()

### LINKED ISSUE #1402: Node.js API transaction object undocumented
https://googleapis.dev/nodejs/firestore/latest/Firestore_.html#runTransaction

The transaction object provided to the updateFunction is not directly documented (only indirectly via examples)

With a starting point of https://firebase.google.com/docs/firestore/manage-data/transactions, I assumed t.create/t.set were both async since its Nodejs example shows t.update being async.

That does not seem to be the case, though without the API documentation I hesitate to say it is wrong, since I don't know if it is intended to be async or not.

thebrianchen:
Fixes #1402.

This provides `updateFunction` with detailed callback typings.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -869,6 +869,19 @@ export class Firestore implements firestore.Firestore {
   }
 
   /**
+   * Function executed by {@link Firestore#runTransaction} within the transaction
+   * context.
+   *
+   * @callback Firestore~updateFunction
+   * @template T
+   * @param  {Transaction} transaction The transaction object for this
+   * transaction.
+   * @returns {Promise<T>} The promise returned at the end of the transaction.
+   * This promise will be returned by {@link Firestore#runTransaction} if the transaction
+   * completed successfully.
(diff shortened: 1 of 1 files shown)


dev/src/index.ts
```diff
@@ -868,6 +868,19 @@ export class Firestore implements firestore.Firestore {
...
+   * Function executed by {@link Firestore#runTransaction} within the transaction
+   * context.
+   *
+   * @callback Firestore~updateFunction
+   * @template T
+   * @param  {Transaction} transaction The transaction object for this
+   * transaction.
+   * @returns {Promise<T>} The promise returned at the end of the transaction.
+   * This promise will be returned by {@link Firestore#runTransaction} if the transaction
+   * completed successfully.
```
schmidt-sebastian, line 881: This looks like it could use some manual word wrap.

schmidt-sebastian (approved):

PR #1410, opened 2021-02-03
chore: small BulkWriter cleanup

schmidt-sebastian:
This removes an indirection between BulkWriter and BulkWriterOperation.


thebrianchen (approved):
lgtm, thanks!

PR #1412, opened 2021-02-04
chore: add option to fetch all descendants

thebrianchen:
I'm trying to separate out the PRs into smaller chunks where I can. This PR contains only the logic to fetch all descendants using a StructuredQuery. I only included a single system test as a sanity check. Once I add recursive delete functionality, the test coverage should include the code to fetch all descendants.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -40,7 +40,7 @@ import {
   validateResourcePath,
 } from './path';
 import {ClientPool} from './pool';
-import {CollectionReference} from './reference';
+import {CollectionReference, Query, QueryOptions} from './reference';
 import {DocumentReference} from './reference';
 import {Serializer} from './serializer';
 import {Timestamp} from './timestamp';
@@ -1184,6 +1184,39 @@ export class Firestore implements firestore.Firestore {
    */
   _decrementBulkWritersCount(): void {
     this.bulkWritersCount -= 1;
+  }
+
+  /**
+   * Retrieves all descendant documents nested under the provided reference.
+   *
+   * @private
+   */
+  // TODO(chenbrian): Make this a private method after adding recursive delete.
+  _getAllDescendants(
+    ref: CollectionReference | DocumentReference
+  ): Promise<DocumentReference[]> {
+    const firestore =
+      ref instanceof CollectionReference
+        ? ref._firestore
+        : ref.parent._firestore;
+    const segments = ref.path.split('/');
+    const isDocumentPath = segments.length % 2 === 0;
+
+    // The parent is the closest ancestor document to the location we're
+    // deleting. If we are deleting a document, the parent is the path of that
+    // document. If we are deleting a collection, the parent is the path of the
+    // document containing that collection (or the database root, if it is a
+    // root collection.
+    if (!isDocumentPath) {
+      segments.pop();
+    }
+    const parentPath = ResourcePath.EMPTY.construct(segments);
+
+    const query = new Query(
+      firestore,
+      QueryOptions.forDocumentNamesOnly(parentPath, ref.id, isDocumentPath)
+    );
+    return query._getAllDescendants();
   }
 
   /**
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1220,7 +1220,7 @@ export class Firestore implements firestore.Firestore {
     const collectionId =
       ref instanceof CollectionReference ? ref.id : ref.parent.id;
 
-    let query: Query<firestore.DocumentData> = new Query(
+    let query: Query = new Query(
       this,
       QueryOptions.forKindlessAllDescendants(parentPath, collectionId)
     );
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -157,6 +157,18 @@ const DEFAULT_MAX_IDLE_CHANNELS = 1;
 const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
 
 /**
+ * Datastore allowed numeric IDs where Firestore only allows strings. Numeric
+ * IDs are exposed to Firestore as __idNUM__, so this is the lowest possible
+ * negative numeric value expressed in that format.
+ *
+ * This constant is used to specify startAt/endAt values when querying for all
+ * descendants in a single collection.
+ *
+ * @private
+ */
+const REFERENCE_NAME_MIN_ID = '__id-9223372036854775808__';
+
+/**
  * Document data (e.g. for use with
  * [set()]{@link DocumentReference#set}) consisting of fields mapped
  * to values.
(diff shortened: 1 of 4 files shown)


dev/src/index.ts
```diff
@@ -1186,6 +1186,39 @@ export class Firestore implements firestore.Firestore {
...
   }
 
+  /**
+   * Retrieves all descendant documents nested under the provided reference.
+   *
+   * @private
+   */
+  // TODO(chenbrian): Make this a private method after adding recursive delete.
+  _getAllDescendants(
+    ref: CollectionReference | DocumentReference
```
schmidt-sebastian, line 1196: Does this also need to take Firestore?

dev/src/index.ts
```diff
@@ -1186,6 +1186,39 @@ export class Firestore implements firestore.Firestore {
...
+  ): Promise<DocumentReference[]> {
+    const firestore =
```
schmidt-sebastian, line 1198: Isn't `firestore` `this`?

dev/src/index.ts
```diff
@@ -1186,6 +1186,39 @@ export class Firestore implements firestore.Firestore {
...
+      ref instanceof CollectionReference
+        ? ref._firestore
+        : ref.parent._firestore;
+    const segments = ref.path.split('/');
```
schmidt-sebastian, line 1202: Can you make CollectionReference and DocumentReference implement the same ResourcePath-exposing property? Then you can access the path as ref._resourcePath or something and don't need to split.

dev/src/index.ts
```diff
@@ -1186,6 +1186,39 @@ export class Firestore implements firestore.Firestore {
...
+    const isDocumentPath = segments.length % 2 === 0;
```
schmidt-sebastian, line 1203: `ref instanceof DocumentReference` (can be inline too)

dev/src/reference.ts
```diff
@@ -92,6 +91,18 @@ const comparisonOperators: {
   'array-contains-any': 'ARRAY_CONTAINS_ANY',
 };
 
+/**
+ * Datastore allowed numeric IDs where Firestore only allows strings. Numeric
+ * IDs are exposed to Firestore as __idNUM__, so this is the lowest possible
+ * negative numeric value expressed in that format.
```
schmidt-sebastian, line 97: I learned something today! :)

dev/src/reference.ts
```diff
@@ -92,6 +91,18 @@ const comparisonOperators: {
...
+ *
+ * This constant is used to specify startAt/endAt values when querying for all
+ * descendants in a single collection.
+ *
+ * @private
+ */
+const MIN_ID = '__id-9223372036854775808__';
```
schmidt-sebastian, line 104: This file is gigantic. Can you use a more specific name?

dev/src/reference.ts
```diff
@@ -1035,6 +1047,33 @@ export class QueryOptions<T> {
...
+    let options = new QueryOptions<T>(
+      parent,
+      id,
+      defaultConverter(),
+      /*allDescendants=*/ true,
+      /*fieldFilters=*/ [],
+      /*fieldOrders=*/ []
+    );
+
+    options = options.with({
```
schmidt-sebastian, line 1070: Maybe use `select()` here (and maybe `FieldPath.documentId()`)

dev/src/reference.ts
```diff
@@ -1879,6 +1920,34 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
...
+        .on('data', result => {
+          if (result.document) {
+            docs.push(result.document);
+          }
+        })
+        .on('end', () => {
+          resolve(docs);
+        });
+    });
+  }
```
schmidt-sebastian, line 1949: Can we not use the existing method and return DocumentSnapshots with empty data() (similar to what `Query.select(FieldPath.documentId())` already does?)

dev/src/reference.ts
```diff
@@ -2088,21 +2205,27 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
...
-          );
-          // Recreate the QueryDocumentSnapshot with the DocumentReference
-          // containing the original converter.
-          finalDoc.fieldsProto = document._fieldsProto;
-          finalDoc.readTime = document.readTime;
-          finalDoc.createTime = document.createTime;
-          finalDoc.updateTime = document.updateTime;
-          lastReceivedDocument = finalDoc.build() as QueryDocumentSnapshot<T>;
-          callback(undefined, {document: lastReceivedDocument, readTime});
+          if (this._queryOptions.ancestorType) {
```
schmidt-sebastian, line 2208: I think you could just return DocumentSnapshots here as well.

dev/src/reference.ts
```diff
@@ -2088,21 +2205,27 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
...
+            const finalDoc = new DocumentSnapshotBuilder<T>(
+              document.ref.withConverter(this._queryOptions.converter)
+            );
+            // Recreate the QueryDocumentSnapshot with the DocumentReference
+            // containing the original converter.
+            finalDoc.fieldsProto = document._fieldsProto;
+            finalDoc.readTime = document.readTime;
+            finalDoc.createTime = document.createTime;
+            finalDoc.updateTime = document.updateTime;
+            lastReceivedDocument = finalDoc.build() as QueryDocumentSnapshot<T>;
```
schmidt-sebastian, line 2226: You probably want to hit the code path because `lastReceivedDocument` implements implicit query resumption for deadline failures.

dev/system-test/firestore.ts
```diff
@@ -2628,6 +2628,56 @@ describe('BulkWriter class', () => {
...
+      tx.set(randomCol.doc('bob/parentsCol/charlie'), {name: 'charlie'});
+      tx.set(randomCol.doc('bob/parentsCol/daniel'), {name: 'daniel'});
+      tx.set(randomCol.doc('bob/parentsCol/daniel/childCol/ernie'), {
+        name: 'ernie',
+      });
+      tx.set(randomCol.doc('bob/parentsCol/daniel/childCol/francis'), {
+        name: 'francis',
+      });
+      return Promise.resolve();
+    });
```
schmidt-sebastian, line 2657: Don't have to fix this, but this is what WriteBatch is for.

dev/src/reference.ts
thebrianchen: Done. Also used `where()` for the MIN_ID logic to standardize in a single place.

thebrianchen: I don't want to add the extra setting to QueryOptions just to remove the `collectionId` field, but I can't figure out a clean way to. Any ideas?

dev/src/reference.ts
thebrianchen: Done.

dev/src/reference.ts
thebrianchen: Done.

thebrianchen (commented):
Thanks for the review!

dev/src/reference.ts
schmidt-sebastian: Can you add a `parent` to `forCollectionGroupQuery` and make the `collectionId` EMPTY (which removes `omitCollectionId` entirely)? 

FWIW, the query options object this method returns doesn't actually do the filtering of document names. If you keep this method, it should get a different name.

dev/src/index.ts

dev/src/index.ts
```diff
@@ -1186,6 +1198,69 @@ export class Firestore implements firestore.Firestore {
...
+  _getAllDescendants(
+    ref: CollectionReference | DocumentReference
+  ): Promise<DocumentReference[]> {
+    const segments = ref.path.split('/');
+
+    // The parent is the closest ancestor document to the location we're
+    // deleting. If we are deleting a document, the parent is the path of that
+    // document. If we are deleting a collection, the parent is the path of the
+    // document containing that collection (or the database root, if it is a
+    // root collection.
```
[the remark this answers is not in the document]

dev/src/index.ts
```diff
@@ -1186,6 +1198,69 @@ export class Firestore implements firestore.Firestore {
...
+        })
+        .on('data', result => {
+          if (result) {
+            docs.push(result);
+          }
+        })
+        .on('end', () => {
+          resolve(docs);
+        });
+    });
```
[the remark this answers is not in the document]

dev/src/reference.ts
thebrianchen: Renamed to `forKindlessAllDescendants()`. 

We need the `collectionId` to not be empty since `Query.id` would then incorrectly return an empty string.

dev/src/index.ts
```diff
@@ -1186,6 +1198,53 @@ export class Firestore implements firestore.Firestore {
...
+    // document containing that collection (or the database root, if it is a
+    // root collection).
+    let parentPath = ref._resourcePath;
+    if (ref instanceof CollectionReference) {
+      parentPath = parentPath.popLast();
+    }
+    const collectionId =
+      ref instanceof CollectionReference ? ref.id : ref.parent.id;
+
+    let query: Query<firestore.DocumentData> = new Query(
```
schmidt-sebastian, line 1223:
```suggestion
    let query = new Query(
```

schmidt-sebastian (approved):
LGTM. Thanks for the updates

dev/src/index.ts
thebrianchen: Actually, i have to revert this change, since the tests are failing:
`Type 'Query<DocumentData>' is not assignable to type 'Query<unknown>'.`

schmidt-sebastian: Hm, shouldn't it default to `DocumentData`? I will take a look.

PR #1425, opened 2021-02-13
fix: promote "protobufjs" to a dependency from a devDependency

Robin-Hoodie:
Using Yarn 2 in PnP mode will break this package as it seems to require "protobufjs" in runtime code, while it's not specified as a dependency, resulting in 

> "@google-cloud/firestore tried to access protobufjs, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound"

I did not open an issue for this beforehand as the problem it's trying to solve seems obvious

You can view more info on this here: https://yarnpkg.com/configuration/yarnrc#packageExtensions and here https://yarnpkg.com/features/pnp

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
   "dependencies": {
     "fast-deep-equal": "^3.1.1",
     "functional-red-black-tree": "^1.0.1",
-    "google-gax": "^2.9.2"
+    "google-gax": "^2.9.2",
+    "protobufjs": "^6.8.6"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


JustinBeckwith (approved):

Robin-Hoodie:
I'm not sure what the last 2 CI tests are being blocked by. Is there anything that's expected from me before being able to merge this PR?

schmidt-sebastian:
We have to kick off the tests on your behalf. They are running now.

PR #1427, opened 2021-02-17
feat: add recursiveDelete()

thebrianchen:
First pass at adding recursive delete. Merging to a feature branch for iterative development.

Some things that I still have yet to add:
- Firestore instance deletion
- BulkWriter backoffs (extra for RESOURCE_EXHAUSTED)
- More aggressive delete retry policy (ex: RST_STREAM errors).
- Add toString() for document references
- No throttling on deletes.


dev/src/index.ts
```diff
@@ -1198,15 +1219,100 @@ export class Firestore implements firestore.Firestore {
...
+        writer.delete(docRef).catch(err => {
+          errorCount++;
+          lastError = err;
+        });
+      })
+      .on('end', () => {
+        writer.flush().then(async () => {
+          if (ref instanceof DocumentReference) {
+            try {
+              await ref.delete();
```
thebrianchen, line 1274: I can't think of a more elegant way to differentiate between the final document reference delete failing vs. the BulkWriter deletes. Any suggestions?

schmidt-sebastian: I haven't looked at the rest of the PR, but if you want to work on it before I do - couldn't you just let the BulkWriter handle this delete as well?

schmidt-sebastian:
One more note: Can you ping the backend chat to determine whether we can retry deletes that fail with RST_STREAM?

dev/src/index.ts
```diff
@@ -1198,15 +1219,100 @@ export class Firestore implements firestore.Firestore {
...
+    // Capture the error stack to preserve stack tracing across async calls.
+    const stack = Error().stack!;
+
+    docStream
+      .on('error', err => {
+        err.code = Status.FAILED_PRECONDITION;
+        err.message =
+          'Failed to fetch children documents. ' +
+          'The provided reference was not deleted.';
+        deleteCompleted.reject(wrapError(err, stack));
```
thebrianchen, line 1261: Is not deleting the provided reference in the case of a failed StructuredQuery the ideal, expected behavior? I need a quick sanity check. Here are the options:
1. Do not delete the provided reference if StructuredQuery stream ever errors. 
2. Do not delete provided reference iff StructuredQuery stream errors before any deletes are enqueued.
3. Delete the reference regardless of what happens.

I think we originally agreed on Option 2. It makes logical sense -- if Firestore can't fetch any of the descendants, that's a failed precondition, so Firestore shouldn't delete the provided reference. However, what's tripping me up is if the StructuredQuery stream errors out midway through the deletion process. In this case, an argument can be made both ways -- Firestore shouldn't delete the provided reference because it couldn't fetch the descendants (failed precondition argument) OR Firestore should delete the provided reference to be consistent in behavior (the provided reference is deleted even if BulkWriter fails to delete some references). 

Trying to document all these edge cases led me to consider Option 3 -- if I'm a developer who simply wants the collection or document deleted, I don't care about the provided reference, and Firestore should just nuke whatever it can get a hold of.

dev/src/index.ts
thebrianchen: I still have a few more changes/cleanup I need to push upstream, so reviewing now might result in some wasted work.

Thanks for the suggestion! I originally figured a normal delete is less verbose than enqueuing another operation to BulkWriter and calling flush(), but using BulkWriter allows for any user callbacks to also run.

dev/src/index.ts
schmidt-sebastian: The query can error at any point in time. I would suggest we make no guarantees whatsoever, which makes the overall behavior of the API much more predictable. If any operation fails, the database tree is in an undefined state, and the user needs to try again.

thebrianchen: Done. Changed it to always delete the provided reference when possible, but kept the error message in for query stream failures. Not sure if we want the FAILED_PRECONDITION status code though.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+          deleteCompletedDeferred.resolve();
+        } else {
+          let error = new GoogleError(
+            `${errorCount} ` +
+              `${errorCount > 1 ? 'deletes' : 'delete'} ` +
+              'failed. The last delete failed with: '
+          );
+          if (lastError instanceof BulkWriterError) {
+            error.code = (lastError.code as number) as Status;
+          }
```
thebrianchen, line 1272: Not sure why these lines aren't covered by Codecov, but I'm pretty sure they're being called. Maybe it's b/c I put it in a lambda function to access `lastError` and `errorCount`. Is this an anti-pattern?

dev/src/index.ts
```diff
@@ -76,6 +76,7 @@ const serviceConfig = interfaces['google.firestore.v1.Firestore'];
 
 import api = google.firestore.v1;
 import {CollectionGroup} from './collection-group';
+import {DocumentData} from '@google-cloud/firestore';
```
schmidt-sebastian, line 79: This is already imported. If you don't like typing `firestore.DocumentData` you should instead do `const DocumentData = firestore.DocumentData`.

dev/src/index.ts
```diff
@@ -399,6 +400,26 @@ export class Firestore implements firestore.Firestore {
    */
   private registeredListenersCount = 0;
 
+  /**
+   * A lazy-loaded BulkWriter instance to be used with recursiveDelete() if no
+   * BulkWriter instance is provided.
+   *
+   * @private
+   */
+  private _bulkWriter: BulkWriter | undefined;
```
schmidt-sebastian, line 409: Could we move this to BulkWriter (kind of like a singleton pattern)? This could then be a shared instance that could be used by other operations in the future as well.

I also wonder whether the default ramp up makes sense here. It will slow down the delete (especially when compared to the CLI) and it also doesn't quite fit the use case for 5/5/5. We do not have to ramp up to give the database a chance to shard - the data is already sharded, we just want to delete it.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
     this.bulkWritersCount -= 1;
   }
 
+  /**
+   * Recursively deletes all documents and subcollections at and under the
+   * specified level.
+   *
+   * If any deletes fail, the promise is rejected with an error message
```
schmidt-sebastian, line 1226: "delete fails"?

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+   * containing the number of failed deletes and the stack trace of the last
+   * failed delete. The provided reference is deleted regardless of whether
+   * all deletes succeeded, except when Firestore fails to fetch the provided
+   * reference's descendants.
```
schmidt-sebastian, line 1230: I would just remove the last sentence, otherwise it forces us to keep the implementation as is.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+   *
+   * Firestore uses a BulkWriter instance with default settings to perform the
```
schmidt-sebastian, line 1232:
```suggestion
   * `recursiveDelete()` uses a BulkWriter instance with default settings to perform the
```

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+   * deletes. To customize throttling rates or add success/error callbacks,
+   * pass in a custom BulkWriter instance.
+   *
```
schmidt-sebastian, line 1235: You should add a code snippet here that shows how to collect the references for the failed delete.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+   * @param ref The reference of a document or collection to delete.
+   * @param bulkWriter Custom BulkWriter instance used to perform the deletes.
+   * @return A promise that resolves when all deletes have been performed.
+   * The promise is rejected if any of the deletes fail.
+   */
+  recursiveDelete<T = DocumentData>(
+    ref: CollectionReference<T> | DocumentReference<T>,
```
schmidt-sebastian, line 1242: You don't need `T` here. It can be `unknown`.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+    bulkWriter?: BulkWriter
+  ): Promise<void> {
+    const writer = bulkWriter ?? this.getBulkWriter();
```
schmidt-sebastian, line 1245: Why do we need the extra `writer` variable?

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+    writer._verifyNotClosed();
+    const docStream = this.getAllDescendants(ref);
```
schmidt-sebastian, line 1247:
```suggestion
    const docStream = this._getAllDescendants(ref);
```

I also wonder if we should move this somewhere else (not that there is a good place for this). The class is getting a bit big.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+    const incrementErrorCount = (err: Error): void => {
+      errorCount++;
+      lastError = err;
+    };
+    const onStreamEnd = (): void => {
+      writer.flush().then(async () => {
+        if (ref instanceof DocumentReference) {
+          writer.delete(ref).catch(err => incrementErrorCount(err));
+          await writer.flush();
+        }
```
schmidt-sebastian, line 1260:
```suggestion
      if (ref instanceof DocumentReference) {
          writer.delete(ref).catch(err => incrementErrorCount(err));
        }
      writer.flush().then(async () => {
```

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+          // Wrap the BulkWriter error last to provide the full stack trace.
+          deleteCompletedDeferred.reject(
+            lastError.stack ? wrapError(error, lastError.stack ?? '') : error
+          );
+        }
+      });
+    };
+
+    // Capture the error stack to preserve stack tracing across async calls.
+    const stack = Error().stack!;
```
schmidt-sebastian, line 1284: It might be cleaner to make this the first line of the method.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+
+    docStream
+      .on('error', err => {
+        err.code = Status.FAILED_PRECONDITION;
+        err.message =
+          'Failed to fetch children documents. ' +
+          'The provided reference was not deleted.';
```
schmidt-sebastian, line 1291: The stream can fail mid-stream, so any number of deletes could have taken place.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+        incrementErrorCount(err);
```
schmidt-sebastian, line 1292: This shouldn't bump up the error count that is used for the error message, should it?

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+        onStreamEnd();
+      })
+      .on('data', (snap: QueryDocumentSnapshot) => {
+        const docRef = new DocumentReference(this, snap.ref._path);
+        writer.delete(docRef).catch(err => incrementErrorCount(err));
```
schmidt-sebastian, line 1297:
```suggestion
        writer.delete(snap.ref).catch(err => incrementErrorCount(err));
```

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
+   * The promise is rejected if any of the deletes fail.
+   */
+  recursiveDelete<T = DocumentData>(
+    ref: CollectionReference<T> | DocumentReference<T>,
+    bulkWriter?: BulkWriter
+  ): Promise<void> {
+    const writer = bulkWriter ?? this.getBulkWriter();
+    writer._verifyNotClosed();
+    const docStream = this.getAllDescendants(ref);
+    const deleteCompletedDeferred = new Deferred<void>();
```
schmidt-sebastian, line 1248: Should this just be `result`? While generic, it makes it clear that this is what the method returns.

dev/src/index.ts
```diff
@@ -1198,15 +1219,96 @@ export class Firestore implements firestore.Firestore {
...
   /**
    * Retrieves all descendant documents nested under the provided reference.
    *
    * @private
    * @return {Stream<QueryDocumentSnapshot>} Stream of descendant documents.
    */
-  // TODO(chenbrian): Make this a private method after adding recursive delete.
-  _getAllDescendants(
-    ref: CollectionReference | DocumentReference
+  private getAllDescendants<T = DocumentData>(
```
schmidt-sebastian, line 1310: This should still be underscore prefixed to prevent "usage by code completion".

schmidt-sebastian (commented):
Feedback for your implementation. I hope that I can get to the tests soon. They look a bit scary in that you wrote a lot of them :)

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
     });
   });
 });
+
+describe('recursiveDelete() method:', () => {
```
schmidt-sebastian, line 1300: Consider moving these to a separate file.

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+  });
+
+  describe('deletes', () => {
+    it('collection', async () => {
+      firestore = await instantiateInstance([
+        ['anna', 'bob', 'bob/children/charlie', 'bob/children/daniel'],
+      ]);
+      await firestore.recursiveDelete(firestore.collection('collectionId'));
+      expect(batchWriteError, 'batchWrite should not have errored').to.be
+        .undefined;
```
schmidt-sebastian, line 1450: Should this be part of `afterEach`?

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+        );
+      },
+      batchWrite: request => {
+        const expected = createRequest(
+          childrenDocs[requestCount].map(docId => deleteOp(docId))
+        );
+        try {
+          expect(request.writes).to.deep.equal(expected.writes);
+        } catch (e) {
+          batchWriteError = e;
```
schmidt-sebastian, line 1333: Do the tests not fail if we simply let the exception throw here? Ideally, the implementation should be able to handle if the `batchWrite` RPC throws an exception and propagate it as a failed BatchWrite.

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+            'root/doc',
+            select('__name__'),
+            allDescendants(/* kindless= */ true)
+          );
+          return stream();
+        },
+        // Include dummy response for the deleted docRef.
+        batchWrite: () => {
+          return response(successResponse(1));
+        },
```
schmidt-sebastian, line 1436:
```suggestion
        batchWrite: () => response(successResponse(1)),
```

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+      return firestore.recursiveDelete(firestore.doc('root/doc'));
+    });
+  });
+
+  describe('deletes', () => {
+    it('collection', async () => {
+      firestore = await instantiateInstance([
+        ['anna', 'bob', 'bob/children/charlie', 'bob/children/daniel'],
+      ]);
+      await firestore.recursiveDelete(firestore.collection('collectionId'));
```
schmidt-sebastian, line 1448: I am not sure I follow this test. The documents are not part of the collection, but they are still getting returned from the `RunQuery` request if I am not mistaken.

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+          throw new GoogleError('commit() failed in test');
+        },
+      };
+      firestore = await createInstance(overrides);
+      await expect(firestore.recursiveDelete(firestore.doc('root/doc'))).to
+        .eventually.be.rejected;
+    });
+
+    it.only('retries stream errors', async () => {
+      let attempts = 0;
```
schmidt-sebastian, line 1531: Please fix

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+      let requestCounter = 0;
+      const docIds = ['a', 'b', 'c'];
+      const streamItems = docIds.map(docId => [result(docId)]);
+      const expected = docIds.map(docId => createRequest([deleteOp(docId)]));
+      const responses = docIds.map(() => successResponse(1));
+
+      const overrides: ApiOverride = {
+        runQuery: () => {
+          return stream(...streamItems[requestCounter]);
+        },
```
schmidt-sebastian, line 1603:
```suggestion
        runQuery: () => stream(...streamItems[requestCounter]),
```

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+    });
+  });
+
+  describe('BulkWriter instance', () => {
+    it('uses custom BulkWriter instance if provided', async () => {
+      firestore = await instantiateInstance([['a', 'b', 'c']]);
+      let callbackCount = 0;
+      const bulkWriter = firestore.bulkWriter();
+      bulkWriter.onWriteResult(() => {
+        callbackCount++;
```
schmidt-sebastian, line 1651: Do we want to add a test that verifies that we return the expected list of DocumentReferences?

dev/test/index.ts
```diff
@@ -1273,3 +1296,389 @@ describe('getAll() method', () => {
...
+      });
+      await firestore.recursiveDelete(firestore.collection('foo'), bulkWriter);
+      expect(callbackCount).to.equal(3);
```
schmidt-sebastian, line 1654: We also don't seem to have tests that exercise the error handler.

dev/test/query.ts
```diff
@@ -282,27 +285,47 @@ function endAt(
...
+
 export function queryEquals(
   actual: api.IRunQueryRequest | undefined,
   ...protoComponents: api.IStructuredQuery[]
-) {
+): void {
+  queryEqualsHelper(actual, '', ...protoComponents);
+}
+
+function queryEqualsHelper(
```
schmidt-sebastian, line 303: I am not a huge fan of `Helpers`. Would it be possible to just make this `queryEqualsWithParent`?

types/firestore.d.ts
```diff
@@ -242,6 +242,30 @@ declare namespace FirebaseFirestore {
...
 
+    /**
+     * Recursively deletes all documents and subcollections at and under the
+     * specified level.
+     *
+     * If any deletes fail, the promise is rejected with an error message
+     * containing the number of failed deletes and the stack trace of the last
+     * failed delete. The provided reference is deleted regardless of whether
+     * all deletes succeeded, except when Firestore fails to fetch the provided
+     * reference's descendants.
```
schmidt-sebastian, line 253: Please update according to the suggestions above.

types/firestore.d.ts
```diff
@@ -242,6 +242,30 @@ declare namespace FirebaseFirestore {
...
+     * deletes. To customize throttling rates or add success/error callbacks,
+     * pass in a custom BulkWriter instance.
+     *
+     * @param ref The reference of a document or collection to delete.
+     * @param bulkWriter Custom BulkWriter instance used to perform the deletes.
+     * @return A promise that resolves when all deletes have been performed.
+     * The promise is rejected if any of the deletes fail.
+     */
+    recursiveDelete(
+      ref: CollectionReference | DocumentReference,
```
schmidt-sebastian, line 265: Please add `<unknown>`, otherwise this typing won't allow converters.

schmidt-sebastian (commented):
Feedback for the tests. Generally looks great!

dev/src/index.ts
thebrianchen: > Could we move this to BulkWriter (kind of like a singleton pattern)? This could then be a shared instance that could be used by other operations in the future as well.

Do you mean that calling `firestore.bulkWriter()` would return the same instance each time, or having a Firestore-internal `BulkWriter.bulkWriter()` getter that lazy-loads?

> I also wonder whether the default ramp up makes sense here. 

Yeah, that's something I was thinking about as well to implement as part of the RST_STREAM retries. Deletes aren't subject to 555 throttling, which creates two possibilities. 1) We can add an internal override on BulkWriter to turn off throttling when recursive delete is called. However, this would mean that writes being performed at the same time won't get throttled. 2) We can add logic for BulkWriter to automatically ignore throttling limits for deletes, but then we'd have to batch deletes separately from writes (or think of another solution that allows writes to still be throttled).

thebrianchen (commented):
Thanks for the detailed review! Addressed comments for the logic, going to work on test comments next.

thebrianchen (commented):
> One more note: Can you ping the backend chat to determine whether we can retry deletes that fail with RST_STREAM? (I see this is already part of your TODOs, so please pretend that you never saw this note).

Checked in with with backend team, and it's fine to retry deletes since they're idempotent.

dev/src/index.ts
schmidt-sebastian: Firestore internal getter that lazy-laods.

dev/src/index.ts
schmidt-sebastian: "private" is only hidden from TS developers. JS developers will still see it in code completion.

Underscore suffix is the same as underscore prefix, but from about 10 years ago. I am surprised this client uses underscore suffix.

dev/src/index.ts
```diff
@@ -1198,27 +1218,128 @@ export class Firestore implements firestore.Firestore {
...
+   * @return A promise that resolves when all deletes have been performed.
+   * The promise is rejected if any of the deletes fail.
+   *
+   * @example
+   * // Recursively delete a reference and log the references of failures.
+   * const bulkWriter = firestore.bulkWriter();
+   * bulkWriter
+   *   .onWriteError((error) => {
+   *     if (
+   *       error.code === GrpcStatus.UNAVAILABLE &&
```
schmidt-sebastian, line 1245: I think we should drop this line. Deletes are almost always retyrable, and we should show this here and do a blind retry up to MAX_RETRY_ATTEMPTS times.

Regardless of what the developer does, we should apply a sensible backoff on top of these attempts. This means that we should backoff longer for RESOURCE_EXHAUSTED.

dev/src/index.ts
```diff
@@ -1198,27 +1218,128 @@ export class Firestore implements firestore.Firestore {
...
+   *       error.failedAttempts < MAX_RETRY_ATTEMPTS
+   *     ) {
+   *       return true;
+   *     } else {
+   *       console.log('Failed write at document: ', error.documentRef);
```
schmidt-sebastian, line 1250: We should probably add a `toString()` to make this output sensible.

"That sounds like a good idea, but it seems unrelated to this PR."

I came up with an answer for you to copy & paste.

dev/src/index.ts
```diff
@@ -1198,27 +1218,128 @@ export class Firestore implements firestore.Firestore {
...
+          resultDeferred.reject(
+            lastError.stack ? wrapError(error, lastError.stack ?? '') : error
+          );
+        }
+      });
+    };
+
+    docStream
+      .on('error', err => {
+        err.code = Status.FAILED_PRECONDITION;
```
schmidt-sebastian, line 1307: Please take a look at https://developers.google.com/maps-booking/reference/grpc-api/status_codes

FAILED_PRECONDITION means that a simple retry won't work here. The developer has to change something about the state of the system before retrying the operation. This is likely not the case here. I would guess that we will have an error code here already when we hit this code path, but if we don't, I would suggest using UNAVAILABLE as the return code.

dev/src/index.ts
```diff
@@ -1198,27 +1218,128 @@ export class Firestore implements firestore.Firestore {
...
     // document containing that collection (or the database root, if it is a
     // root collection).
+
     let parentPath = ref._resourcePath;
     if (ref instanceof CollectionReference) {
       parentPath = parentPath.popLast();
     }
     const collectionId =
-      ref instanceof CollectionReference ? ref.id : ref.parent.id;
+      ref instanceof CollectionReference
```
schmidt-sebastian, line 1340: CollectionReference means `CollectionReference<DocumentData>`. If you change this to `ref instanceof CollectionReference<unknown>` you should be able to remove the `as DocumentReference` cast (which should also be `as DocumentReference<unknown>`).

dev/test/query.ts
schmidt-sebastian: You could do:

```
export function queryEquals(
  actual: api.IRunQueryRequest | undefined,
  parentOrProtoComponent: string|api.IStructuredQuery|undefined,
  ...protoComponents: api.IStructuredQuery[]
): void;
```

But that is not all that pretty.

I was actually thinking you keep both `queryEquals` and `queryEqualsWithParent`, but merge `queryEqualsWithParent` and `queryEqualsHelper`.

dev/system-test/firestore.ts
```diff
@@ -2629,69 +2629,95 @@ describe('BulkWriter class', () => {
...
+
+    it('recursiveDelete with custom BulkWriter instance', async () => {
+      const bulkWriter = firestore.bulkWriter();
+      let callbackCount = 0;
+      bulkWriter.onWriteResult(() => {
+        callbackCount++;
+      });
+      await firestore.recursiveDelete(randomCol, bulkWriter);
+      expect(callbackCount).to.equal(6);
+    });
```
schmidt-sebastian, line 2720: Can we also get a test that has "randomCollA/docA" and "randomCollB/docB" and only deletes randomCollA? There might be some nuances in how to delete parent collection (in other ports) and I want to make sure that all of our clients handle this case correctly.

dev/test/index.ts
schmidt-sebastian: No, I think we can keep the logic as is. Individual RPC failures versus individual operations failures are not different from a user perspective. I was trying to see if we could simply bubble up the assertion failure as an individual RPC failure, which would surface these assertions in `onWriteResult` or via failed promises rather than via `batchWriteError`.

schmidt-sebastian (commented):
Looking better and better. Some small suggestions remain.

dev/src/index.ts
schmidt-sebastian: Looks good.

schmidt-sebastian (approved):

types/firestore.d.ts
```diff
@@ -242,6 +242,46 @@ declare namespace FirebaseFirestore {
...
+     *       error.code === GrpcStatus.UNAVAILABLE &&
+     *       error.failedAttempts < MAX_RETRY_ATTEMPTS
+     *     ) {
+     *       return true;
+     *     } else {
+     *       console.log('Failed write at document: ', error.documentRef);
+     *       return false;
+     *     }
+     *   });
+     * await firestore.recursiveDelete(docRef, bulkWriter);
```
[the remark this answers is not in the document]
thebrianchen: thanks for catching.

dev/src/index.ts
thebrianchen: Added the `<unknown>` type to the `DocumentReference` cast.

types/firestore.d.ts
```diff
@@ -242,6 +242,46 @@ declare namespace FirebaseFirestore {
...
+     * containing the number of failed deletes and the stack trace of the last
+     * failed delete. The provided reference is deleted regardless of whether
+     * all deletes succeeded.
+     *
+     * `recursiveDelete()` uses a BulkWriter instance with default settings to
+     * perform the deletes. To customize throttling rates or add success/error
+     * callbacks, pass in a custom BulkWriter instance.
+     *
+     * @param ref The reference of a document or collection to delete.
+     * @param bulkWriter Custom BulkWriter instance used to perform the deletes.
```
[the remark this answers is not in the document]
thebrianchen: done.

dev/src/bulk-writer.ts
```diff
@@ -372,6 +374,20 @@ export class BulkWriter {
...
 
+  /**
+   * Returns a lazy-loaded BulkWriter instance.
+   *
+   * @param firestore Used to instantiate the BulkWriter instance the first
+   * time.
+   * @private
+   */
+  static _getInstance(firestore: Firestore): BulkWriter {
+    if (!this.instance || this.instance.firestore !== firestore) {
```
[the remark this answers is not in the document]
thebrianchen: I can see arguments either way, but at this point, I'm more in favor of leaving it on the Firestore instance, since trying to track the different firestore instances to pass into BulkWriter is more confusing than just having the BulkWriter instance available on the Firestore instance.

Reverted back to before.

dev/system-test/firestore.ts
```diff
@@ -2709,7 +2709,21 @@ describe('BulkWriter class', () => {
...
+
+      // Add other nested collection under 'bob' that shouldn't be deleted.
+      const collB = randomCol.doc('bob').collection('pets');
+      await collB.doc('doggo').set({name: 'goodboi'});
+
+      await firestore.recursiveDelete(collA);
+
+      expect(await countCollectionChildren(collA)).to.equal(0);
+      expect(await countCollectionChildren(randomCol)).to.equal(3);
+      expect(await countCollectionChildren(collB)).to.equal(1);
```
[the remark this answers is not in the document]
thebrianchen: Thanks for clarifying what you meant. Added a test for that.

thebrianchen (commented):
Thanks for the quick feedback cycles!

Issue #1422, opened 2021-02-12
Add type definitions for v1 and v1beta1 clients


merlinnot:
Type definitions for v1 and v1beta1 exports are currently `any`.

I intended to do it myself (#1421), but `protos` is outside of `src` and since it's imported, the build output changes from `build/src` to `build/dev/src`, which would obviously require a lot of changes.

The simples fix would be to move `protos` to `src`, this way all other files and references can remain unchanged. WDYT?

schmidt-sebastian:
Moving `dev/protos` to `src/dev/protos` sounds good to me. 

Just a heads up - you will have to change https://github.com/googleapis/nodejs-firestore/blob/master/synth.py. Let us know if you want us to do this first.

merlinnot:
Just a small update from my side, I'm still planning to pick this up, probably within the next two-three weeks. If you guys have some time sooner than that, go ahead! :slightly_smiling_face:

schmidt-sebastian:
@thebrianchen Is working on changes that are required to make our source generation work with the new Proto location. We might either leave it at that part, or take this issue to completion once that is done.

merlinnot:
I'll have time for it in two to three weeks, don't wait for me if you have time to do that sooner.

merlinnot:
Thank you, that will be very useful for me!

PR #1431, opened 2021-02-25
chore: move dev/protos to dev/src/protos

thebrianchen:
Moving protos as part of #1422.

I spent some time playing around with when to move the protos/ directory in. I originally wanted to just generate the protos in the correct directory, but the `node.compile_protos()` method requires `protos/` and `src/` to be in the same directory.

I tried adding in #1421, but doing so results in the the `compile` npm command failing. For some reason, adding the types causes the entire `dev/` directory to be copied into `build/`, creating a new directory structure that has folders in the wrong place.


dev/src/protos/protos.json
```diff
@@ -3973,8 +3973,7 @@
                 "REQUIRED": 2,
                 "OUTPUT_ONLY": 3,
                 "INPUT_ONLY": 4,
-                "IMMUTABLE": 5,
-                "UNORDERED_LIST": 6
```
schmidt-sebastian, line 3977: Can you verify that this change was meant to happen? I am surprised that there is no open PR from Synthtool if this is an intended change.

The source of truth might be here: https://source.corp.google.com/piper///depot/google3/google/api/field_behavior.proto;l=56;bpv=1

synth.py
```diff
@@ -152,18 +185,25 @@
...
 os.chdir("protos")
 os.unlink('protos.js')
 os.unlink('protos.d.ts')
 subprocess.run('./update.sh', shell=True)
+
+# Copy the protos directory into dev/src after compiling.
 os.chdir("../../")
+s.move("dev/protos", "dev/src/protos")
+os.system("rm -rf dev/protos")
+
```
schmidt-sebastian, line 206: I think it might make sense to update the proto location in the generated client here. Before you move them, the source is not valid, is it?

schmidt-sebastian (approved):

dev/src/protos/protos.json
thebrianchen: I'm not sure what I've done differently. When I run synthtool on a master, this file isn't changed. However, I've verified that running synthtool on this branch results in this change happening. Tests still pass, but I'm not sure if this is a sign of something else wrong.

synth.py
thebrianchen: Moved.

thebrianchen:
Done in #1433, which didn't require moving the `protos/` location.

PR #1442, opened 2021-03-05
fix: retry BulkWriter deletes that fail with RST_STREAM error

thebrianchen:


dev/src/bulk-writer.ts
```diff
@@ -316,10 +316,14 @@ export class BulkWriter {
    * @private
    */
   private _errorFn: (error: BulkWriterError) => boolean = error => {
+    const isRstStreamError =
+      (error.code as number) === Status.INTERNAL &&
```
schmidt-sebastian, line 320: 1) We should be able to retry Status.INTERNAL regardless of whether this is a RST_STREAM error.
2) We should only add this retry for deletes.

dev/src/bulk-writer.ts
```diff
@@ -316,10 +316,13 @@ export class BulkWriter {
    * @private
    */
   private _errorFn: (error: BulkWriterError) => boolean = error => {
+    const isRetryableDeleteError =
+      error.operationType === 'delete' &&
+      (error.code as number) === Status.INTERNAL;
     const retryCodes = getRetryCodes('batchWrite');
     return (
       error.code !== undefined &&
```
schmidt-sebastian, line 324: Do we actually need this undefined check? Would retryCodes.include() just return false?

schmidt-sebastian (approved):

dev/src/bulk-writer.ts
thebrianchen: nope! removed.

dev/src/bulk-writer.ts
jyane: Would you please explain why the retry should be only for deletes?
The RST could happen on any other methods, this limitation comes from idempotency?

For example, Spanner automatically retries transaction commits, regardless of the queries within the txn. I wonder if there is a firestore specific thing here.

PR #1447, opened 2021-03-10
fix: BulkWriter: add backoff on retries

thebrianchen:


dev/src/bulk-writer.ts
```diff
@@ -714,27 +739,39 @@ export class BulkWriter {
...
-    this._bulkCommitBatch = new BulkCommitBatch(this.firestore);
-
-    // Send the batch if it is under the rate limit, or schedule another
-    // attempt after the appropriate timeout.
-    const underRateLimit = this._rateLimiter.tryMakeRequest(
+    // Use the write with the highest failure count when determining backoff.
+    const highestFailureCount = pendingBatch.pendingOps.reduce((prev, cur) =>
+      prev.failedAttempts > cur.failedAttempts ? prev : cur
+    ).failedAttempts;
+    const resourceExhausted =
```
thebrianchen, line 746: Not sure if this is worth adding a getter for `pendingOps`, or if I should just add instance methods on `BulkCommitBatch` for these two usages.

dev/test/bulk-writer.ts
```diff
@@ -1053,6 +1065,43 @@ describe('BulkWriter', () => {
       });
     return bulkWriter.close().then(() => {
       verifyOpCount(1);
+      expect(timeoutHandlerCounter).to.equal(MAX_RETRY_ATTEMPTS - 1);
+    });
+  });
+
+  it('applies maximum backoff on retries for RESOURCE_EXHAUSTED', async () => {
```
schmidt-sebastian, line 1072: Let's also add a test that uses multiple different backoffs in a batch and a test that verifies that we send a second batch before the first batch if the first batch has a backoff.

thebrianchen: done.

dev/src/bulk-writer.ts
```diff
@@ -65,6 +71,12 @@ export const DEFAULT_INITIAL_OPS_PER_SECOND_LIMIT = 500;
  */
 export const DEFAULT_MAXIMUM_OPS_PER_SECOND_LIMIT = 10000;
 
+/*!
+ * The default jitter to apply. For example, a factor of 0.3 means a 30% jitter
+ * is applied.
+ */
```
schmidt-sebastian, line 77: This should mention that this relates to backoff. It needs some context in this module.

dev/src/bulk-writer.ts
```diff
@@ -83,7 +95,9 @@ const RATE_LIMITER_MULTIPLIER_MILLIS = 5 * 60 * 1000;
  */
 class BulkWriterOperation {
   private deferred = new Deferred<WriteResult>();
-  private failedAttempts = 0;
+  private _failedAttempts = 0;
+  private _lastError?: Status;
```
schmidt-sebastian, line 99: lastStatus?

dev/src/bulk-writer.ts
```diff
@@ -107,16 +121,24 @@ class BulkWriterOperation {
     return this.deferred.promise;
   }
 
+  get lastError(): Status | undefined {
```
schmidt-sebastian, line 124: I am surprised this need to be exported.

dev/src/bulk-writer.ts
```diff
@@ -714,27 +748,34 @@ export class BulkWriter {
...
+      assert(
+        this._rateLimiter.tryMakeRequest(pendingBatch._opCount),
+        'RateLimiter should allow request if delayMs === 0'
+      );
       delayedExecution.resolve();
     } else {
-      const delayMs = this._rateLimiter.getNextRequestDelayMs(
-        pendingBatch._opCount
-      );
+      const finalDelayMs = Math.max(backoffMsWithJitter, delayMs);
```
schmidt-sebastian, line 772: You can move out of the if statement and then use it in the `if(...)` check itself.

dev/src/bulk-writer.ts
```diff
@@ -714,27 +748,34 @@ export class BulkWriter {
     const tag = requestTag();
     const pendingBatch = this._bulkCommitBatch;
 
-    this._bulkCommitBatch = new BulkCommitBatch(this.firestore);
-
-    // Send the batch if it is under the rate limit, or schedule another
-    // attempt after the appropriate timeout.
-    const underRateLimit = this._rateLimiter.tryMakeRequest(
+    // Use the write with the highest failure count when determining backoff.
+    const highestBackoffDuration = pendingBatch.pendingOps.reduce((prev, cur) =>
```
schmidt-sebastian, line 752: I think this doesn't quite work. Let's say we are supposed to back off for 10 seconds, but the rate limiter tells us that we need to wait 20 seconds. Once 20 seconds passed, we should be able to send the request right away. If I read this correctly, we still apply the 10 second backoff here.

I think you need to use a "target time" rather than a duration.

dev/src/bulk-writer.ts
```diff
@@ -743,6 +784,24 @@ export class BulkWriter {
...
+   *
+   * @private
+   */
+  private static applyJitter(backoffMs: number): number {
+    if (backoffMs === 0) return 0;
+    // Random value in [-0.3, 0.3].
+    const jitter =
+      Math.random() *
+      DEFAULT_JITTER_FACTOR *
+      (Math.round(Math.random()) ? 1 : -1);
```
schmidt-sebastian, line 798: Math.random() is not cheap. You can remove one call: Math.random() * 2 * DEFAULT_JITTER_FACTOR - DEFAULT_JITTER_FACTOR.

dev/test/bulk-writer.ts
```diff
@@ -1053,7 +1069,123 @@ describe('BulkWriter', () => {
...
+      DEFAULT_BACKOFF_INITIAL_DELAY_MS * DEFAULT_BACKOFF_FACTOR,
+    ];
+    setTimeoutHandler((fn, timeout) => {
+      // 1st batch should have max backoff. 2nd batch should have 1 round
+      // of backoff applied.
+      expect(timeout).to.be.within(
+        (1 - DEFAULT_JITTER_FACTOR) * expected[timeoutHandlerCounter],
+        (1 + DEFAULT_JITTER_FACTOR) * expected[timeoutHandlerCounter]
+      );
+      timeoutHandlerCounter++;
```
schmidt-sebastian, line 1127: Nice

dev/src/bulk-writer.ts
thebrianchen: Done.

dev/src/bulk-writer.ts
thebrianchen: Done.

dev/src/bulk-writer.ts
thebrianchen: Me too, removed.

dev/src/bulk-writer.ts
thebrianchen: Done.

dev/src/bulk-writer.ts
thebrianchen: I think the backoffs are applied together. We take apply the `max()` of the backoff and ratelimiter delays, and call `bulkCommit()` on the batch immediately after. Where do we apply the 2nd backoff?

dev/src/bulk-writer.ts
thebrianchen: Ooh that's some nice mathing! Refactored it slightly to make it more understandable (at least to me).

dev/src/bulk-writer.ts
schmidt-sebastian: If I read this correctly, if a batch hits a rate limit, the function call to compute the next backoff is executed again. This once again looks at the rate limit backoff (which would be 0) and re-computes the retry backoff. It looks like this last step starts from scratch again - instead, it should deduct the backoff that has already passed. Let me know if I am confused and this works differently.

thebrianchen: Discussed it on gvc. The current implementation doesn't apply double backoff, and can be merged. However, I still need to add rate limiter checking before calling `bulkCommit()`, which I will do in a separate PR on top of this one.

dev/src/bulk-writer.ts
```diff
@@ -83,7 +95,9 @@ const RATE_LIMITER_MULTIPLIER_MILLIS = 5 * 60 * 1000;
  */
 class BulkWriterOperation {
   private deferred = new Deferred<WriteResult>();
-  private failedAttempts = 0;
+  private _failedAttempts = 0;
```
schmidt-sebastian, line 98: You can rename this back to its old name since there is no getter anymore.

dev/src/bulk-writer.ts
schmidt-sebastian: After discussing offline, it looks like my understanding of how this worked was slightly off. Can stay as is.

schmidt-sebastian (approved):

PR #1451, opened 2021-03-16
fix: BulkWriter: apply rate limiter before sending batch

thebrianchen:


dev/src/bulk-writer.ts
```diff
@@ -738,58 +738,62 @@ export class BulkWriter {
...
-      );
-      assert(
-        underRateLimit,
-        'RateLimiter should allow request if delayMs === 0'
-      );
-      delayedExecution.resolve();
+    const underRateLimit = this._rateLimiter.tryMakeRequest(batch._opCount);
+    if (underRateLimit) {
+      await batch.bulkCommit({requestTag: tag});
+      if (flush) this._scheduleCurrentBatch(flush);
```
schmidt-sebastian, line 779: This codeflow still confused me, but I now know why. This sound like it is trying to re-send the same batch again. Should we call this `_scheduleNextBatch()`? FWIW, I think this function could also just be called `_sendBatch()` since it takes the batch as an argument now.

schmidt-sebastian (approved):

dev/src/bulk-writer.ts
thebrianchen: Yeah, that makes more sense. Renamed.

PR #1453, opened 2021-03-22
fix: avoid adding duplicate orderBys in startAfter()

thebrianchen:
Fixing a bug in recursive delete where invalid queries would be created on stream errors due to adding two `orderBy('__name__', ASCENDING)` clauses when calling `startAfter()` ([location](https://github.com/googleapis/nodejs-firestore/blob/2cf01c2f3f1b2e439d0620a77afd292ec3707da4/dev/src/reference.ts#L2187)). The private method `createImplicitOrderBy` adds an implicit orderBy if there is is no documentId on the fieldOrder fields, but we didn't check the case where the documentId comes from the fieldPaths.

### DIFF
--- a/dev/test/query.ts
+++ b/dev/test/query.ts
@@ -49,6 +49,7 @@ import {
 } from './util/helpers';
 
 import api = google.firestore.v1;
+import protobuf = google.protobuf;
 
 const PROJECT_ID = 'test-project';
 const DATABASE_ROOT = `projects/${PROJECT_ID}/databases/(default)`;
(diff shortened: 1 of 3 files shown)


dev/test/query.ts
```diff
@@ -285,9 +286,24 @@ function endAt(
...
+/**
+ * Returns the timestamp value for the provided readTimes, or the default
+ * readTime value used in tests if no values are provided.
+ */
+export function readTime(
+  seconds?: number,
+  nanos?: number
+): protobuf.ITimestamp {
+  if (seconds === undefined && nanos === undefined) {
+    return {seconds: '5', nanos: 6};
```
thebrianchen, line 298: We have to cast to seconds since the `Timestamp` class always [converts it](https://github.com/googleapis/nodejs-firestore/blob/5f6683c2177aa7bd80ec81c3942443453b834529/dev/src/timestamp.ts#L262) to `string` for some reason. Why do we do this?

schmidt-sebastian: Proto3 JSON represents int64 as strings, and the seconds component of Timestamps is an int64.

dev/test/query.ts
```diff
@@ -326,7 +346,12 @@ export function queryEquals(
   actual: api.IRunQueryRequest | undefined,
   ...protoComponents: api.IStructuredQuery[]
 ): void {
-  queryEqualsWithParent(actual, '', ...protoComponents);
+  queryEqualsWithParentAndReadTime(
+    actual,
+    '',
```
schmidt-sebastian, line 351: What is ''? Can you add a comment?

schmidt-sebastian (approved):

PR #1462, opened 2021-03-31
fix: omit readTime in RunQueryRequest for recursive deletes

thebrianchen:
The RunQuery stream returns documents with the same readTime from when the query stream starts. Since the backend requires that the readTime be with 270 seconds, the query stream will error when it's restarted with an old readTime. Since we don't need the readTime when performing recursive deletes, we can omit the the field instead.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -1009,8 +1009,8 @@ export class QueryOptions<T> {
     readonly kindless = false,
     // Whether to require consistent documents when restarting the query. By
     // default, restarting the query uses the readTime offset of the original
-    // query.
-    readonly requireConsistency = false
+    // query to provide consistent results.
+    readonly requireConsistency = true
   ) {}
 
   /**
@@ -1057,7 +1057,8 @@ export class QueryOptions<T> {
    */
   static forKindlessAllDescendants<T = firestore.DocumentData>(
     parent: ResourcePath,
-    id: string
+    id: string,
+    requireConsistency = true
   ): QueryOptions<T> {
     let options = new QueryOptions<T>(
       parent,
@@ -1070,7 +1071,7 @@ export class QueryOptions<T> {
 
     options = options.with({
       kindless: true,
-      requireConsistency: true,
+      requireConsistency,
     });
     return options;
   }
@@ -2185,11 +2186,11 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
                 // `requestStream()` will backoff should the restart fail before
                 // delivering any results.
                 if (this._queryOptions.requireConsistency) {
-                  request = this.startAfter(lastReceivedDocument).toProto();
-                } else {
                   request = this.startAfter(lastReceivedDocument).toProto(
                     lastReceivedDocument.readTime
                   );
+                } else {
+                  request = this.startAfter(lastReceivedDocument).toProto();
                 }
               }
               streamActive.resolve(/* active= */ true);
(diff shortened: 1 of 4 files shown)


dev/src/reference.ts
```diff
@@ -2177,9 +2177,18 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
                 // query cursor. Note that we do not use backoff here. The call to
                 // `requestStream()` will backoff should the restart fail before
                 // delivering any results.
-                request = this.startAfter(lastReceivedDocument).toProto(
-                  lastReceivedDocument.readTime
-                );
+                if (this._queryOptions.kindless) {
```
thebrianchen, line 2180: I originally wanted to abstract away the recursive delete requirements from the Query class, but I can't think of another way to omit readTime elegantly. At some point, should I just rename the queryOptions to include recursive delete in the naming instead of "kindless"?

schmidt-sebastian: Can we just be explicit here and create a new setting that either mentions read time directly or addresses the concept that we want to change? Something like `enforceConsistency` or `requiresConsistency` might work.

thebrianchen: That makes sense. Changed to `requireConsistency`.

dev/src/reference.ts
```diff
@@ -1006,7 +1006,11 @@ export class QueryOptions<T> {
     readonly projection?: api.StructuredQuery.IProjection,
     // Whether to select all documents under `parentPath`. By default, only
     // collections that match `collectionId` are selected.
-    readonly kindless = false
+    readonly kindless = false,
+    // Whether to require consistent documents when restarting the query. By
+    // default, restarting the query uses the readTime offset of the original
+    // query.
+    readonly requireConsistency = false
```
schmidt-sebastian, line 1013: This should default to `true`.

dev/src/reference.ts
```diff
@@ -1066,6 +1070,7 @@ export class QueryOptions<T> {
 
     options = options.with({
       kindless: true,
+      requireConsistency: true,
```
schmidt-sebastian, line 1073: This should be false (and you should pass it in since there is not really a reason why a kindless all-descendants query should not require consistency by default)

dev/src/reference.ts
```diff
@@ -2177,12 +2184,7 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
                 // query cursor. Note that we do not use backoff here. The call to
                 // `requestStream()` will backoff should the restart fail before
                 // delivering any results.
-                if (this._queryOptions.kindless) {
-                  // Kindless queries are currently used exclusively for recursive
-                  // deletes. We don't need the readTime in for documents that are
-                  // going to be deleted. The backend also requires the readTime
-                  // to not be older than 270 seconds, so omitting the readTime
-                  // is easier than trying to keep it updated.
+                if (this._queryOptions.requireConsistency) {
```
schmidt-sebastian, line 2187: Again, flip this.

dev/src/reference.ts
thebrianchen: done.

dev/src/reference.ts
thebrianchen: done.

dev/src/reference.ts
thebrianchen: done.

schmidt-sebastian (approved):

PR #1464, opened 2021-04-06
fix: remove floating point rounding error in Timestamp.fromMillis()

### LINKED ISSUE #1463: Timestamp.fromMillis can't handle nanoseconds.
This is the client library issue.

no.

no.

#### Environment details

  - OS: macOS Big Sur 11.2.3
  - Node.js version: v14.16.0
  - npm version: 6.14.12
  - `@google-cloud/firestore` version: 3.8.6

#### Steps to reproduce

  1. `$ node -pe 'require("@google-cloud/firestore/build/src/timestamp").Timestamp.fromMillis(1000.1);'`
     
     More than 1000 milliseconds may be an abnormal case.
     
     ```
     ./node_modules/@google-cloud/firestore/build/src/validate.js:191
                 throw new Error(invalidArgumentMessage(arg, 'integer'));
                 ^
     
     Error: Value for argument "nanoseconds" is not a valid integer.
         at Object.validateInteger (./node_modules/@google-cloud/firestore/build/src/validate.js:191:19)
         at new Timestamp (./node_modules/@google-cloud/firestore/build/src/timestamp.js:74:20)
         at Function.fromMillis (./node_modules/@google-cloud/firestore/build/src/timestamp.js:126:16)
         at [eval]:1:66
         at Script.runInThisContext (vm.js:133:18)
         at Object.runInThisContext (vm.js:310:38)
         at internal/process/execution.js:77:19
         at [eval]-wrapper:6:22
         at evalScript (internal/process/execution.js:76:60)
         at internal/main/eval_string.js:23:3
     ```

     - `$ node -pe 'require("@google-cloud/firestore/build/src/timestamp").Timestamp.fromMillis(999.1);'`
     
       Less than 1000 milliseconds is always a normal case.
       
       ```
       Timestamp { _seconds: 0, _nanoseconds: 999100000 }
       ```

> Making sure to follow these steps will guarantee the quickest resolution possible.

https://github.com/googleapis/nodejs-firestore/blob/42ade788be6e9c1ba01b6e81779e5185d5cc75cd/dev/src/timestamp.ts#L110
The above line will be
`        const nanos = Math.round((milliseconds - seconds * 1000) * MS_TO_NANOS);`
or to avoid minus's round-off error
`        const nanos = Math.round(milliseconds * MS_TO_NANOS - seconds * 1000 * MS_TO_NANOS);`

thebrianchen:
Fixes #1463.

By multiplying `MS_TO_NANOS` separately, we avoid the floating point precision rounding errors.

### DIFF
--- a/dev/src/timestamp.ts
+++ b/dev/src/timestamp.ts
@@ -107,7 +107,7 @@ export class Timestamp implements firestore.Timestamp {
    */
   static fromMillis(milliseconds: number): Timestamp {
     const seconds = Math.floor(milliseconds / 1000);
-    const nanos = Math.round(
+    const nanos = Math.floor(
       milliseconds * MS_TO_NANOS - seconds * 1000 * MS_TO_NANOS
     );
     return new Timestamp(seconds, nanos);
--- a/dev/src/timestamp.ts
+++ b/dev/src/timestamp.ts
@@ -107,7 +107,9 @@ export class Timestamp implements firestore.Timestamp {
    */
   static fromMillis(milliseconds: number): Timestamp {
     const seconds = Math.floor(milliseconds / 1000);
-    const nanos = (milliseconds - seconds * 1000) * MS_TO_NANOS;
+    const nanos = Math.round(
+      milliseconds * MS_TO_NANOS - seconds * 1000 * MS_TO_NANOS
+    );
     return new Timestamp(seconds, nanos);
   }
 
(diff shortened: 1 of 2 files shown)


dev/src/timestamp.ts
```diff
@@ -107,7 +107,9 @@ export class Timestamp implements firestore.Timestamp {
    */
   static fromMillis(milliseconds: number): Timestamp {
     const seconds = Math.floor(milliseconds / 1000);
-    const nanos = (milliseconds - seconds * 1000) * MS_TO_NANOS;
+    const nanos = Math.round(
```
schmidt-sebastian, line 110: I think you need to round down here so that nanos is always less than "one second in nanos".

schmidt-sebastian (approved):

PR #1468, opened 2021-04-08
fix: use BigInt when calculating nanos in Timestamp.fromMillis()

thebrianchen:
After porting the original change to the web SDK, I noticed that the nanos calculation is subject to floating point precision loss. I used Math.floor() on the Web SDK, but using BigInt is more accurate, so I prefer to use it where we can.


thebrianchen:
Build is broken b/c of sinon [update](https://github.com/sinonjs/sinon/issues/2352), but will merge once it's fixed.

schmidt-sebastian (approved):

PR #1470, opened 2021-04-12
feat: add buffering layer to BulkWriter

thebrianchen:
This PR adds a buffer layer to BulkWriter to limit the number of outstanding RPCs that BulkWriter has. This helps prevent BulkWriter from OOMing when multiple RPCs start failing due to hotspotting or contention.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -861,9 +861,7 @@ export class BulkWriter {
 
     // Advance the `_lastOp` pointer. This ensures that `_lastOp` only resolves
     // when both the previous and the current write resolves.
-    this._lastOp = this._lastOp.then(() => {
-      return silencePromise(bulkWriterOp.promise);
-    });
+    this._lastOp = this._lastOp.then(() => silencePromise(bulkWriterOp.promise));
 
     // Schedule the operation if the BulkWriter has fewer than the maximum
     // number of allowed pending operations, or add the operation to the
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -861,7 +861,9 @@ export class BulkWriter {
 
     // Advance the `_lastOp` pointer. This ensures that `_lastOp` only resolves
     // when both the previous and the current write resolves.
-    this._lastOp = this._lastOp.then(() => silencePromise(bulkWriterOp.promise));
+    this._lastOp = this._lastOp.then(() =>
+      silencePromise(bulkWriterOp.promise)
+    );
 
     // Schedule the operation if the BulkWriter has fewer than the maximum
     // number of allowed pending operations, or add the operation to the
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -90,10 +90,11 @@ const RATE_LIMITER_MULTIPLIER_MILLIS = 5 * 60 * 1000;
 
 /*!
  * The default maximum number of pending operations that can be enqueued onto a
- * BulkWriter instance. BulkWriter buffers additional writes after this many
- * pending operations in order to avoiding going OOM.
+ * BulkWriter instance. An operation is considered pending if BulkWriter has
+ * sent it via RPC and is awaiting the result. BulkWriter buffers additional
+ * writes after this many pending operations in order to avoiding going OOM.
  */
-const DEFAULT_MAXIMUM_PENDING_OPERATIONS_COUNT = 10000;
+const DEFAULT_MAXIMUM_PENDING_OPERATIONS_COUNT = 500;
 
 /**
  * Represents a single write for BulkWriter, encapsulating operation dispatch
@@ -339,6 +340,8 @@ export class BulkWriter {
 
   /**
    * The number of pending operations enqueued on this BulkWriter instance.
+   * An operation is considered pending if BulkWriter has sent it via RPC and
+   * is awaiting the result.
    * @private
    */
   private _pendingOpsCount = 0;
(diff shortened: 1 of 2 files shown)


dev/src/bulk-writer.ts
```diff
@@ -817,7 +879,22 @@ export class BulkWriter {
...
       this._successFn.bind(this)
     );
-    this._sendFn(enqueueOnBatchCallback, bulkWriterOp);
+    if (this._pendingOpsCount < this._maxPendingOpCount) {
+      this._pendingOpsCount++;
+      this._sendFn(enqueueOnBatchCallback, bulkWriterOp);
+    } else {
+      // Increment the _lastOp pointer to ensure that the promise returned by
+      // flush() includes this operation.
+      this._lastOp = this._lastOp.then(() => {
```
schmidt-sebastian, line 888: I think we should do this in both cases and remove the duplication between here and _sendFn.

dev/src/bulk-writer.ts
```diff
@@ -817,7 +879,22 @@ export class BulkWriter {
...
+        // eslint-disable-next-line no-console
+        console.log('hacky lastOp increment complete');
+            return silencePromise(bulkWriterOp.promise);
+          }
+      );
+      this._bufferedOperations.push(() =>
+        this._sendFn(enqueueOnBatchCallback, bulkWriterOp)
```
schmidt-sebastian, line 895: Do we need to increment pendingOpsCount? (did not check)

dev/src/bulk-writer.ts
```diff
@@ -817,8 +855,54 @@ export class BulkWriter {
...
+    // enqueued before the next batch is scheduled in `_sendBatch()`.
+    return bulkWriterOp.promise
+      .then(res => {
+        this._processBufferedOps();
+        return res;
+      })
+      .catch(err => {
+        this._processBufferedOps();
+        throw err;
+      });
```
schmidt-sebastian, line 889: You can probably return this._lastOp if you move this code to either line 862 above or _sendFn.

dev/test/bulk-writer.ts
```diff
@@ -500,6 +500,50 @@ describe('BulkWriter', () => {
...
+          successResponse(3),
+        ]),
+      },
+      {
+        request: createRequest([setOp('doc4', 'bar'), setOp('doc5', 'bar')]),
+        response: mergeResponses([successResponse(4), successResponse(5)]),
+      },
+    ]);
+    // eslint-disable-next-line no-console
+    console.log('BCHEN start buffer test');
```
schmidt-sebastian, line 523: Remove

schmidt-sebastian (approved):
This looks great! Some optional suggestions.

dev/src/bulk-writer.ts
```diff
@@ -330,6 +337,37 @@ export class BulkWriter {
    */
   readonly _rateLimiter: RateLimiter;
 
+  /**
+   * The number of pending operations enqueued on this BulkWriter instance.
+   * @private
+   */
+  private _pendingOpsCount = 0;
```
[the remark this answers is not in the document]
thebrianchen: Couldn't think of a better term. It seemed like pending was the closest thing, but I added a separate comment to describe what it means.

dev/src/bulk-writer.ts
```diff
@@ -817,8 +855,54 @@ export class BulkWriter {
...
       this._successFn.bind(this)
     );
-    this._sendFn(enqueueOnBatchCallback, bulkWriterOp);
-    return bulkWriterOp.promise;
+
+    // Advance the `_lastOp` pointer. This ensures that `_lastOp` only resolves
+    // when both the previous and the current write resolves.
+    this._lastOp = this._lastOp.then(() => {
+      return silencePromise(bulkWriterOp.promise);
+    });
```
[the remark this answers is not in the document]
thebrianchen: done.

dev/test/bulk-writer.ts
thebrianchen: done.

dev/src/bulk-writer.ts
thebrianchen: I can't return lastOp since it's `Promise<void>`, but the return type is `Promise<WriteResult>`.

dev/src/bulk-writer.ts
```diff
@@ -817,8 +855,54 @@ export class BulkWriter {
...
+
+    // Schedule the operation if the BulkWriter has fewer than the maximum
+    // number of allowed pending operations, or add the operation to the
+    // buffer.
+    if (this._pendingOpsCount < this._maxPendingOpCount) {
+      this._pendingOpsCount++;
+      this._sendFn(enqueueOnBatchCallback, bulkWriterOp);
```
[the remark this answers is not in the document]
thebrianchen: I moved the incrementing call into sendFn. However, `_processedBufferedOps` has to be part of the returned promise in order for it to get run. Any promise off of lastOp doesn't get run by Node, which then creates the race condition with `bulkCommit()` in `_sendBatch`.

dev/src/bulk-writer.ts
```diff
@@ -88,6 +88,13 @@ const RATE_LIMITER_MULTIPLIER = 1.5;
  */
 const RATE_LIMITER_MULTIPLIER_MILLIS = 5 * 60 * 1000;
 
+/*!
+ * The default maximum number of pending operations that can be enqueued onto a
+ * BulkWriter instance. BulkWriter buffers additional writes after this many
+ * pending operations in order to avoiding going OOM.
+ */
+const DEFAULT_MAXIMUM_PENDING_OPERATIONS_COUNT = 10000;
```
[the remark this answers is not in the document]
thebrianchen: Changed to 500, after running some throughput tests. I don't see any drastic changes, but the throughput usually has pretty high variance unless I run tests for > 1 hour.

thebrianchen (commented):
Thanks!

dev/src/bulk-writer.ts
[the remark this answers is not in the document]
schmidt-sebastian: But is it? :)

schmidt-sebastian (approved):

PR #1479, opened 2021-04-16
feat: use limit queries on recursiveDelete()

thebrianchen:
This PR changes `recursiveDelete()` to use limit queries in order to prevent OOMing caused by the query stream from loading documents faster than BulkWriter can delete. I tested different query limit sizes and found that a `limit()` of 5000 seemed to have the fastest throughput that can keep up with BulkWriter on start.

A minor optimization I made was to start the next query when there are < 1000 pending BulkWriter promises. I found that it takes some time for a new query to start, so this should prevent the BulkWriter from idling while waiting for the query to start streaming results. The maximum number of BulkWriter promises would be 6000, and in my benchmarks, this resulted in total maximum memory usage peaking under 100 MB.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1251,7 +1251,7 @@ export class Firestore implements firestore.Firestore {
   ): Promise<void> {
     const writer = bulkWriter ?? this.getBulkWriter();
     const deleter = new RecursiveDelete(this, writer, ref);
-    return deleter.delete();
+    return deleter.run();
   }
 
   /**
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -16,14 +16,14 @@
 
 import * as firestore from '@google-cloud/firestore';
 
-import {CallOptions, GoogleError, grpc, RetryOptions, Status} from 'google-gax';
+import {CallOptions, grpc, RetryOptions, Status} from 'google-gax';
 import {Duplex, PassThrough, Transform} from 'stream';
 
 import {URL} from 'url';
 
 import {google} from '../protos/firestore_v1_proto_api';
 import {ExponentialBackoff, ExponentialBackoffSetting} from './backoff';
-import {BulkWriter, BulkWriterError} from './bulk-writer';
+import {BulkWriter} from './bulk-writer';
 import {BundleBuilder} from './bundle';
 import {fieldsFromJson, timestampFromJson} from './convert';
 import {
@@ -40,7 +40,7 @@ import {
   validateResourcePath,
 } from './path';
 import {ClientPool} from './pool';
-import {CollectionReference, Query, QueryOptions} from './reference';
+import {CollectionReference, Query} from './reference';
 import {DocumentReference} from './reference';
 import {Serializer} from './serializer';
 import {Timestamp} from './timestamp';
@@ -76,6 +76,7 @@ const serviceConfig = interfaces['google.firestore.v1.Firestore'];
 
 import api = google.firestore.v1;
 import {CollectionGroup} from './collection-group';
+import {RecursiveDelete} from './recursive-delete';
 
 export {
   CollectionReference,
@@ -155,34 +156,6 @@ const DEFAULT_MAX_IDLE_CHANNELS = 1;
  * multiplex all requests over a single channel.
  */
 const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
-
-/*!
- * The query limit used for recursiveDelete() when fetching all descendants of
- * the specified reference to delete. This is done to prevent the query stream
- * from streaming documents faster than Firestore can delete.
- */
-export const RECURSIVE_DELETE_QUERY_LIMIT = 5000;
-
-/**
- * The number of pending BulkWriter operations at which recursiveDelete()
- * starts the next limit query to fetch descendants. By starting the query
- * while there are pending operations, Firestore can improve BulkWriter
- * throughput. This helps prevent BulkWriter from idling while Firestore
- * fetches the next query.
- */
-const RECURSIVE_DELETE_PENDING_OP_LIMIT = 1000;
-
-/**
- * Datastore allowed numeric IDs where Firestore only allows strings. Numeric
- * IDs are exposed to Firestore as __idNUM__, so this is the lowest possible
- * negative numeric value expressed in that format.
- *
- * This constant is used to specify startAt/endAt values when querying for all
- * descendants in a single collection.
- *
- * @private
- */
-export const REFERENCE_NAME_MIN_ID = '__id-9223372036854775808__';
 
 /**
  * Document data (e.g. for use with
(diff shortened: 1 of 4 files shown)


dev/src/index.ts
```diff
@@ -1265,19 +1281,95 @@ export class Firestore implements firestore.Firestore {
...
+    // are met.
+    const deleteRef = (docRef: DocumentReference): void => {
+      pendingPromiseCount++;
+      writer
+        .delete(docRef)
+        .then(() => pendingPromiseCount--)
+        .catch(err => {
+          pendingPromiseCount--;
+          incrementErrorCount(err);
+        })
```
schmidt-sebastian, line 1321:
```suggestion
        .catch(err =>  incrementErrorCount(err); })
        .then(() => pendingPromiseCount--)
```

dev/src/index.ts
```diff
@@ -1265,19 +1281,95 @@ export class Firestore implements firestore.Firestore {
...
+          ) {
+            const nextDocStream = this._getAllDescendants(
+              ref instanceof CollectionReference
+                ? (ref as CollectionReference<unknown>)
+                : (ref as DocumentReference<unknown>),
+              /* startAfter= */ lastDocumentSnap
+            );
+            setupStream(nextDocStream);
+          }
+        });
```
schmidt-sebastian, line 1340: This might be easier to reason about if you move the logic out of the `then` block of the delete and instead invoked a separate function. This is about the query itself and not the individual delete anymore.

I also think that this logic is not complex enough that it likely warrants its own class, which would allow you to use proper methods.

dev/src/index.ts
```diff
@@ -1303,17 +1395,12 @@ export class Firestore implements firestore.Firestore {
...
-      })
-      .on('data', (snap: QueryDocumentSnapshot) => {
-        writer.delete(snap.ref).catch(err => incrementErrorCount(err));
-      })
-      .on('end', () => onStreamEnd());
+    const docStream = this._getAllDescendants(
+      ref instanceof CollectionReference
+        ? (ref as CollectionReference<unknown>)
+        : (ref as DocumentReference<unknown>)
+    );
```
schmidt-sebastian, line 1402: I think this could be done in `setupStream` to remove some duplication. I think we should also pass the limit here as it is otherwise not clear where it comes from when looking at the callsite.

dev/src/index.ts
```diff
@@ -156,6 +156,22 @@ const DEFAULT_MAX_IDLE_CHANNELS = 1;
  */
 const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
 
+/*!
+ * The query limit used for recursiveDelete() when fetching all descendants of
+ * the specified reference to delete. This is done to prevent the query stream
+ * from streaming documents faster than Firestore can delete.
+ */
+export const RECURSIVE_DELETE_QUERY_LIMIT = 5000;
```
schmidt-sebastian, line 164:
```suggestion
export const RECURSIVE_DELETE_MAX_PENDING_OPS = 5000;
```

dev/src/index.ts
```diff
@@ -156,6 +156,22 @@ const DEFAULT_MAX_IDLE_CHANNELS = 1;
...
+
+/**
+ * The number of pending BulkWriter operations at which recursiveDelete()
+ * starts the next limit query to fetch descendants. By starting the query
+ * while there are pending operations, Firestore can improve BulkWriter
+ * throughput. This helps prevent BulkWriter from idling while Firestore
+ * fetches the next query.
+ */
+const RECURSIVE_DELETE_PENDING_OP_LIMIT = 1000;
```
schmidt-sebastian, line 173:
```suggestion
export const RECURSIVE_DELETE_MIN_PENDING_OPS = 1000;
```

dev/src/index.ts
```diff
@@ -1265,19 +1281,95 @@ export class Firestore implements firestore.Firestore {
...
+    // run at a time.
+    let streamInProgress = false;
+
+    // The last document snapshot returned by the stream. Used to set the
+    // startAfter() field in the subsequent stream.
+    let lastDocumentSnap: QueryDocumentSnapshot;
+
+    // The number of pending BulkWriter operations. Used to determine when the
+    // next query can be run.
+    let pendingPromiseCount = 0;
```
schmidt-sebastian, line 1309:
```suggestion
    let pendingOpsCount = 0;
```

It it is not clear that the limit is the number of outstanding promises, so I suggest using `ops` instead.

dev/src/index.ts
thebrianchen: done.

dev/src/index.ts
thebrianchen: done.

dev/src/index.ts
thebrianchen: done.

dev/src/index.ts

dev/src/index.ts
thebrianchen: Moved the logic into a separate class.

dev/src/index.ts
thebrianchen: Done. I'm passing the limit in `getAllDescendants()` now.

dev/src/recursive-delete.ts
```diff
@@ -0,0 +1,317 @@
...
+      })
+      .then(() => {
+        this.pendingOpsCount--;
+        // We wait until the previous stream has ended in order to sure the
+        // startAfter document is correct. Starting the next stream while
+        // there are pending operations allows Firestore to maximize
+        // BulkWriter throughput.
+        if (
+          !this.allDescendantsFetched &&
+          !this.streamInProgress &&
```
schmidt-sebastian, line 305: I wonder if could ever hit this code path after all documents are done deleting but before the stream ended. That seems unlikely, but if it is possible it would mean that recursive delete stalls. Do we need to also call `setupStream` in `onStreamEnd()`? I am not sure but you might have better judgement here.

schmidt-sebastian (commented):
This is much cleaner. Some potential nits and potential problems, but this is super close.

dev/src/recursive-delete.ts
thebrianchen: After our conversation, I added a `setupStream()` call at the end.

dev/src/recursive-delete.ts
```diff
@@ -0,0 +1,317 @@
...
+   * @private
+   */
+  private lastError: GoogleError | BulkWriterError | undefined;
+
+  /**
+   * Whether all descendants for this query have been fetched, or if the stream
+   * has errored.
+   * @private
+   */
+  private allDescendantsFetched = false;
```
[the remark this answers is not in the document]
thebrianchen: I tried thinking but `documentsPending` was the best name. Updated the code comments though.

schmidt-sebastian (approved):

thebrianchen:
> Should we just use the same condition as below? (this.pendingOpsCount < MIN_PENDING_OPS)

I don't think so. This is just code to prevent the recursive delete operation from hanging in the error case we discussed. I'd rather have the main scheduling logic be done in `deleteRef()`.

PR #1490, opened 2021-04-29
Upgrade Typescript to 4.2.4

dingbat:

------

I was working on adding proper typing to the Firestore `update` methods (by making `UpdateData` generic), but it needed some newer Typescript features to work.

Sorry I didn't open an issue first, I figured these changes were pretty straightforward.

The only thing I wasn't sure about here was the `as any` for `pathValue` - this was to allow the test to `delete` the key later on (it needed to be optional, and I wasn't sure about the cleanest way to do that since there is no interface defined for `allSupportedTypesObject`).

### DIFF
--- a/dev/conformance/runner.ts
+++ b/dev/conformance/runner.ts
@@ -432,7 +432,7 @@ function runTest(spec: ConformanceProto) {
     };
 
     return createInstance(overrides).then(() => {
-      return new Promise((resolve, reject) => {
+      return new Promise<void>((resolve, reject) => {
         const unlisten = watchQuery().onSnapshot(
           actualSnap => {
             const expectedSnapshot = expectedSnapshots.shift();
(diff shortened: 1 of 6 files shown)


schmidt-sebastian:
@dingbat Thank you for this. I need to verify whether we can accept this PR or whether it is going to break our internal tooling.

@JustinBeckwith - Is this a PR we can take?

JustinBeckwith:
Most likely not.  Changing the version of the TypeScript compiler changes the `d.ts` in ways that can be backwards incompatible with older versions of the TypeScript compiler.  Landing this is a breaking change, and something we should pair with the dropping of Node.js 10.x in a few months.  Tagging in @bcoe so we can coordinate!

schmidt-sebastian:
FWIW We don't use auto-generated types in this repo, so we might be okay here. That being said, manually adding 4.x features to our manual typings is also a breaking change.

glock47 (changes requested):
Data/CmmtSet/924303.csv

bcoe:
we are now on a newer version of typescript 👍

PR #1494, opened 2021-05-04
feat: add recursive delete to Firestore class

### LINKED ISSUE #746: Add Recursive Delete on Collections
When trying to efficiently delete a large (100k+ docs) collection, functionality offered in firebase seems to outperform anything offered with Firestore.

To utilize firebase's delete functionality (https://github.com/firebase/firebase-tools/blob/master/src/firestore/delete.js), we are having to import firebase-tools, firebase-admin, and firebase-functions. Just curious, will some of this stuff be coming to the Firestore SDK soon?

thebrianchen:
Fixes #746.

This PR adds recursive delete functionality to the top level `Firestore` class in a method that recursively deletes all documents and subcollections at and under the specified level. If any delete fails, the promise is rejected with an error message containing the number of failed deletes and the stack trace of the last failed delete. The provided reference is deleted regardless of whether all deletes succeeded.

Example:
```js
// Recursively delete a reference and log the references of failures.
const bulkWriter = firestore.bulkWriter();
bulkWriter
  .onWriteError((error) => {
    if (
      error.failedAttempts < MAX_RETRY_ATTEMPTS
    ) {
      return true;
    } else {
      console.log('Failed write at document: ', error.documentRef.path);
      return false;
    }
  });
await firestore.recursiveDelete(docRef, bulkWriter);
```


schmidt-sebastian (approved):

jakeleventhal:
@thebrianchen how exactly does this work? What is docRef referring to? How would you delete an entire collection?

schmidt-sebastian:
You pass a DocumentReference which will delete the reference, plus any nested collections and documents whose path contains the DocumentReference. You can also pass a CollectionReference, which will delete all directly nested documents and their collections and documents.

jakeleventhal:
will this work with collections that have queries?
@schmidt-sebastian 
```javascript
const bulkWriter = firestore.bulkWriter();
bulkWriter.onWriteError(() => true);
await firestore.recursiveDelete(firestore.collection('Things').where('UserId', '==', id), bulkWriter);
```

schmidt-sebastian:
No, unfortunately it won't. The underlying implementation issues a "key only query" to recursively list all documents in the collection. This query is limited and does not support any additional constraints.

Issue #599, opened 2019-04-14
Documentation Request for firestore.collection.doc.delete


vajahath:
Please provide a short note/sample about the `precondition` argument the `delete()` function can accept in the doc page.
I was referring [firestore delete docs](https://firebase.google.com/docs/firestore/manage-data/delete-data) and couldn't found any.
It would be nice if you reflect the docs on the [definition file](https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L349) also (for better IDE support).

Relevant code:
I was using firestore via [firebase-admin](https://github.com/firebase/firebase-admin-node):
```ts
  await fireAdmin
    .firestore()
    .collection("projects")
    .doc("doc-id")
    .delete(/* precondition */);
```
![image](https://user-images.githubusercontent.com/12165822/56087843-f8900380-5e90-11e9-9dff-162df862c768.png)

schmidt-sebastian:
Hi! Thanks for reporting this issue. The `.d.ts` file should allow your IDE to autocomplete all possible Preconditions. In the meantime, I will modify the sample for https://cloud.google.com/nodejs/docs/reference/firestore/1.1.x/DocumentReference#delete to include the use of a Preondition.

By the way, it's unlikely that we will include this feature in our shared API documentation (the " firestore delete docs" you linked to) since Preconditions are only available via the Server SDKs.

lautarovc:
Hi! The docs and examples are also missing the exists precondition, that determines if a document should exist before deleting. This is defined in the [document.ts](https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/document.ts#L1012) file, Precondition class.

PR #1505, opened 2021-05-18
feat: add Precondition.exists to delete()

thebrianchen:
Adding `Preconditions.exists()` to documentation and TS autocomplete, as requested in [comment](https://github.com/googleapis/nodejs-firestore/issues/599#issuecomment-840093592).

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -244,7 +244,7 @@ const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
  *  enforces that the document was last updated at lastUpdateTime. Fails the
  *  operation if the document was last updated at a different time.
  * @property {boolean} exists If set, enforces that the target document must
- * or must not exist. This property is only valid with delete operations.
+ * or must not exist.
  * @typedef {Object} Precondition
  */
 
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -243,6 +243,9 @@ const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
  * @property {Timestamp} lastUpdateTime The update time to enforce. If set,
  *  enforces that the document was last updated at lastUpdateTime. Fails the
  *  operation if the document was last updated at a different time.
+ * @property {boolean} exists If set, enforces that the target document must
+ * exist. This property
+ * is only valid with delete operations.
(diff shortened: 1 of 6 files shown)


dev/src/index.ts
```diff
@@ -243,6 +243,9 @@ const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
  * @property {Timestamp} lastUpdateTime The update time to enforce. If set,
  *  enforces that the document was last updated at lastUpdateTime. Fails the
  *  operation if the document was last updated at a different time.
+ * @property {boolean} exists If set, enforces that the target document must
+ * exist. This property
+ * is only valid with delete operations.
```
schmidt-sebastian, line 248: This is not necessarily true. It can be used for sets with exists:false. There is also a difference in behavior for true and false and undefined.

dev/src/reference.ts
```diff
@@ -371,6 +371,8 @@ export class DocumentReference<T = firestore.DocumentData>
    * @param {Timestamp=} precondition.lastUpdateTime If set, enforces that the
    * document was last updated at lastUpdateTime. Fails the delete if the
    * document was last updated at a different time.
+   * @param {boolean=} precondition.exists If set, enforces that the target document
+   * must exist.
```
schmidt-sebastian, line 375: must or must not exist (or something like that)

thebrianchen: Done.

dev/src/index.ts
```diff
@@ -243,6 +243,8 @@ const MAX_CONCURRENT_REQUESTS_PER_CLIENT = 100;
  * @property {Timestamp} lastUpdateTime The update time to enforce. If set,
  *  enforces that the document was last updated at lastUpdateTime. Fails the
  *  operation if the document was last updated at a different time.
+ * @property {boolean} exists If set, enforces that the target document must
+ * or must not exist. This property is only valid with delete operations.
```
schmidt-sebastian, line 247: I would delete the last sentence. It can be used with delete(), even if it does not make a lot of sense. It could be used to trigger an error if the deleted document does not exist for example.

schmidt-sebastian (approved):

dev/src/index.ts

PR #1514, opened 2021-05-25
fix: recursive delete: backporting changes from Java

thebrianchen:
Backporting some changes from the Java [port](https://github.com/googleapis/java-firestore/pull/622)

### DIFF
--- a/dev/src/recursive-delete.ts
+++ b/dev/src/recursive-delete.ts
@@ -48,7 +48,7 @@ export const REFERENCE_NAME_MIN_ID = '__id-9223372036854775808__';
  * from streaming documents faster than Firestore can delete.
  */
 // Visible for testing.
-export const MAX_PENDING_OPS = 5000;
+export const RECURSIVE_DELETE_MAX_PENDING_OPS = 5000;
 
 /**
  * The number of pending BulkWriter operations at which RecursiveDelete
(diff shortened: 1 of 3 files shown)


dev/src/recursive-delete.ts
```diff
@@ -84,6 +84,21 @@ export class RecursiveDelete {
...
   private documentsPending = true;
 
+  /**
+   * Whether run() has been called.
+   * @private
+   */
+  private started = false;
+
+  /** Query limit to use when fetching all descendants. */
+  private maxPendingOps: number;
```
schmidt-sebastian, line 94: Does this need to be marked "private" (here and below)?

dev/test/recursive-delete.ts
```diff
@@ -235,8 +239,30 @@ describe('recursiveDelete() method:', () => {
...
+
+      // This deferred promise is used to delay the BatchWriteResponses from
+      // returning in order to create the situation where the number of pending
+      // operations is less than `minPendingOps`.
+      const bufferDeferred = new Deferred<void>();
+
+      // This deferred completes when the second query is run.
+      const secondQueryDeferred = new Deferred<void>();
+
+      const firstStream = Array.from(Array(maxPendingOps).keys()).map((_, i) =>
```
schmidt-sebastian, line 260: Can you create a helper variable for `Array.from(Array(maxPendingOps).keys())` and name it appropriately? It is used twice and not very pretty :)

schmidt-sebastian (approved):

dev/src/recursive-delete.ts
thebrianchen: Thanks for catching, done.

dev/test/recursive-delete.ts
thebrianchen: Done.

PR #1539, opened 2021-06-22
docs: expand documentation on transactions

### LINKED ISSUE #1221: Improve documentation about Transactions making emphasis on the pessimistic locks
The documentation regarding the Transactions is pretty basic regarding on how they work, especially on how the pessimistic transactions work, it just tells you that _Holds a pessimistic lock_, but doesn't mention:

- If a transaction that has multiple reads, blocks other reads/writes
- If a transaction that will perform a write, blocks the resources before executing the actual writing

I even found more information about how the transactions work on the github that what I found on the documentation, the next excerpt is from an issue raised about pessimistic locks here:

> If you are reading a document using the runTransaction() API via a Server SDK the backend will prevent other clients from modifying the document until the transaction is committed. This lock can be held for up to 270 seconds. We refer to this as "pessimistic locking" in our documentation and all Firestore Server SDK expose these semantics.
> 
> There are some rare cases when the backend aborts currently running transactions (e.g. if there is too much contention), in which case all SDKs will retry, independent of how locks were obtained.

I would expect to see at least the same level of detail of documentation from [Client Side Transactions](https://firebase.google.com/docs/firestore/manage-data/transactions#transactions)

schmidt-sebastian:
Fixes https://github.com/googleapis/nodejs-firestore/issues/1221

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -930,11 +930,12 @@ export class Firestore implements firestore.Firestore {
    * out or fails for any reason.
    *
    * Transactions are committed once 'updateFunction' resolves. If a transaction
-   * fails with contention, the transaction is retried five up to five times and
-   * its `updateFunction` is invoked again.
+   * fails with contention, the transaction is retried up to five times. The
+   * `updateFunction` is invoked once for each attempt.
    *
-   * Transaction time out after 270 seconds or 60 seconds of after the last
-   * document read.
+   * Transactions time out after 60 seconds if no documents are read.
+   * Transactions that are not committed within than 270 seconds are also
+   * aborted.
    *
    * @template T
    * @param {Firestore~updateFunction} updateFunction The user function to
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -920,8 +920,21 @@ export class Firestore implements firestore.Firestore {
    * the transaction.
    *
    * You can use the transaction object passed to 'updateFunction' to read and
-   * modify Firestore documents under lock. Transactions are committed once
-   * 'updateFunction' resolves and attempted up to five times on failure.
+   * modify Firestore documents under lock. You have to perform all reads before
+   * before you perform any write.
+   *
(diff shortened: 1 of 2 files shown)


dev/src/index.ts
```diff
@@ -920,8 +920,21 @@ export class Firestore implements firestore.Firestore {
...
+   * before you perform any write.
+   *
+   * Documents read during a transaction are locked pessimistically. A
+   * transaction's lock on a document blocks other transactions, batched
+   * writes, and other non-transactional writes from changing that document.
+   * A transaction releases its document locks at commit time or once it times
+   * out or fails for any reason.
+   *
+   * Transactions are committed once 'updateFunction' resolves. If a transaction
+   * fails with contention, the transaction is retried five up to five times and
```
thebrianchen, line 933: "transaction is retried up to five times"

Also, the current wording makes it seem that the `updateFunction` is only invoked after the retries are completed, which could be misleading.

dev/src/index.ts
```diff
@@ -920,8 +920,21 @@ export class Firestore implements firestore.Firestore {
...
+   * its `updateFunction` is invoked again.
+   *
+   * Transaction time out after 270 seconds or 60 seconds of after the last
```
thebrianchen, line 936: nits:
s/Transaction/Transactions
s/60 seconds of after/seconds after
"last document is read"

Also, when does the timeout change from 60s to 270s? If a document is read in a tx, is the timeout window reduced to 60s + write-only tx timeout have a 270s timeout?

thebrianchen (approved):

dev/src/index.ts
schmidt-sebastian: Changed the wording.

dev/src/index.ts
schmidt-sebastian: Changed to:

```
   * Transactions time out after 60 seconds if no documents are read.
   * Transactions that are not committed within than 270 seconds are also
   * aborted.
 ```

PR #1542, opened 2021-06-23
chore: move transaction default max retry attempts to a const

thebrianchen:
Moving it to a const for better visibility as part of b/147440261.

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -138,8 +138,7 @@ const CLOUD_RESOURCE_HEADER = 'google-cloud-resource-prefix';
 export const MAX_REQUEST_RETRIES = 5;
 
 /*!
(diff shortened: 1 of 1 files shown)


dev/src/index.ts
```diff
@@ -137,6 +137,12 @@ const CLOUD_RESOURCE_HEADER = 'google-cloud-resource-prefix';
  */
 export const MAX_REQUEST_RETRIES = 5;
 
+/*!
+ * The maximum number of times to attempt committing a transaction before
```
schmidt-sebastian, line 141: It's not just committing it, it is retrying the transaction as a whole. Please rephrase.

schmidt-sebastian (approved):

PR #1544, opened 2021-06-24
feat: retry BatchGetDocuments RPCs that fail with errors

### LINKED ISSUE #1387: Retry batchGet RESOURCE_EXHAUSTED
I've experienced the `RESOURCE_EXHAUSTED` error recently a lot. So, I wanted to dig into the error and understand why occurs and how it could be fixed.

I've found [this answer](https://stackoverflow.com/questions/60393891/cloud-firestore-bandwidth-exhausted-error) from a Google Cloud engineer suggesting that this is a quota error, however, I did not find any quota that is exhausted...

To verify that this is indeed a temporary issue (as this kept occurring for the same read document), I've validated that I could read the document outside of the script where it occurs.

Under that assumption, it is a temporary issue and a quota issue. I thought why doesn't it get retried?

I've also found this issue: #909 
Where @schmidt-sebastian confirms that some issues do get retried, but not [`RESOURCE_EXHAUSTED`](https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/v1/firestore_client_config.json#L10-L14) is there a reason why this error does not get retried?

From another issue for Node PubSub, it suggests that usually also a `RESOURCE_EXHAUSTED` error will be retried (at least by other libraries).
https://github.com/googleapis/google-cloud-node/issues/2661#issuecomment-335305754

So I guess we should also retry on `RESOURCE_EXHAUSTED` errors to make them occur less!

#### Environment details

  - OS: Compute Engine image (debian-9-stretch-v20200902)
  - Node.js version: 12.20.0
  - npm version: 6.14.8
  - `@google-cloud/firestore` version: 4.7.1

schmidt-sebastian:
This PR moves the existing logic for Firestore.getAll() to a new class and adds a new retry mechanism that is similar to the current Query RPC retry.

Fixes https://github.com/googleapis/nodejs-firestore/issues/1387

### DIFF
--- a/dev/src/document-reader.ts
+++ b/dev/src/document-reader.ts
@@ -107,8 +107,7 @@ export class DocumentReader<T> {
       request.mask = {fieldPaths};
     }
 
-    let resultCount = 0;
-
+    const responses: api.BatchGetDocumentsResponse[] = [];
     try {
       const stream = await this.firestore.requestStream(
         'batchGetDocuments',
@@ -117,7 +116,13 @@ export class DocumentReader<T> {
       );
       stream.resume();
 
+      // Gather results in a temporary array since Node12 does not handle
+      // exceptions thrown within a `for await` loop.
       for await (const response of stream) {
+        responses.push(response);
+      }
+
+      for (const response of responses) {
         let snapshot: DocumentSnapshot<DocumentData>;
 
         if (response.found) {
@@ -147,15 +152,15 @@ export class DocumentReader<T> {
         const path = snapshot.ref.formattedName;
         this.outstandingDocuments.delete(path);
         this.retrievedDocuments.set(path, snapshot);
-        ++resultCount;
       }
     } catch (error) {
       const shouldRetry =
         // Transactional reads are retried via the transaction runner.
         !this.transactionId &&
         // Only retry if we made progress.
-        resultCount > 0 &&
+        responses.length > 0 &&
         // Don't retry permanent errors.
+        error.code !== undefined &&
         !isPermanentRpcError(error, 'batchGetDocuments');
 
       logger(
@@ -175,7 +180,7 @@ export class DocumentReader<T> {
         'DocumentReader.fetchDocuments',
         requestTag,
         'Received %d results',
-        resultCount
+        responses.length
       );
     }
   }
--- a/dev/src/document-reader.ts
+++ b/dev/src/document-reader.ts
@@ -21,9 +21,8 @@ import {isPermanentRpcError} from './util';
 import {google} from '../protos/firestore_v1_proto_api';
 import {logger} from './logger';
 import {Firestore} from './index';
-
-import api = google.firestore.v1;
 import {DocumentData} from '@google-cloud/firestore';
+import api = google.firestore.v1;
 
 /**
  * A wrapper around BatchGetDocumentsRequest that retries request upon stream
@@ -63,7 +62,7 @@ export class DocumentReader<T> {
    * @param requestTag A unique client-assigned identifier for this request.
    */
   async get(requestTag: string): Promise<Array<DocumentSnapshot<T>>> {
-    await this.fetchAllDocuments(requestTag);
+    await this.fetchDocuments(requestTag);
 
     // BatchGetDocuments doesn't preserve document order. We use the request
     // order to sort the resulting documents.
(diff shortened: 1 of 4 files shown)


dev/src/document-reader.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  private remainingDocuments = new Set<string>();
+  private retrievedDocuments = new Map<string, DocumentSnapshot>();
+
+  /**
+   * Internal method to retrieve multiple documents from Firestore, optionally
+   * as part of a transaction.
+   *
+   * @param firestore The Firestore instance to use.
+   * @param allDocuments The documents to receive.
```
thebrianchen, line 47: documents to *get?

dev/src/document-reader.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @private
+ */
+export class DocumentReader<T> {
+  /** An optional field mask to apply to this read. */
+  fieldMask?: FieldPath[];
+  /** An optional transaction ID to use for this read. */
+  transactionId?: Uint8Array;
+
+  private remainingDocuments = new Set<string>();
```
thebrianchen, line 39: nit: something more specific like unfetchedDocuments or requestedDocuments?

dev/src/document-reader.ts
```diff
@@ -0,0 +1,193 @@
...
+      this.remainingDocuments.add(docRef.formattedName);
+    }
+  }
+
+  /**
+   * Invokes the BatchGetDocuments RPC and returns the results.
+   *
+   * @param requestTag A unique client-assigned identifier for this request.
+   */
+  async get(requestTag: string): Promise<Array<DocumentSnapshot<T>>> {
```
thebrianchen, line 64: If get() is only meant to be run once, the method should throw an error if it is run again

dev/src/document-reader.ts
```diff
@@ -0,0 +1,193 @@
...
+                    response.missing!
+                  );
+                  document = this.firestore.snapshot_(
+                    response.missing!,
+                    response.readTime!
+                  );
+                }
+
+                const path = document.ref.formattedName;
+                this.remainingDocuments.delete(path);
```
thebrianchen, line 173: optional: add assert() to check that path is in the set.

dev/src/document-reader.ts
```diff
@@ -0,0 +1,193 @@
...
+        ) {
+          logger(
+            'DocumentReader.fetchAllDocuments',
+            requestTag,
+            'BatchGetDocuments failed with non-retryable stream error:',
+            err
+          );
+          throw err;
+        } else {
+          logger(
```
thebrianchen, line 111: Is the maximum number of retries allowed capped?

thebrianchen (approved):

dev/src/document-reader.ts
schmidt-sebastian: Renamed to outstandingDocuments. Pretty outstanding, me thinks.

dev/src/document-reader.ts
schmidt-sebastian: Done

dev/src/document-reader.ts
schmidt-sebastian: It is only meant to be run once, but it should work if run multiple times. It will return the same result without doing any work.

dev/src/document-reader.ts
schmidt-sebastian: No, I thought it would be better not to. But then I changed my mind. I rewrote this logic to only fetch more results if we made some progress, which should ensure that we never end up in a loop with just errors. This also removes the need to add backoff, which I otherwise should have done.

FYI - This rewrite was pretty complicated, so I ended up changing the implementation to use async iterators. I did this in steps in case you want to see more reasonable diffs.

dev/src/document-reader.ts
schmidt-sebastian: Skipped for now.

dev/src/document-reader.ts
```diff
@@ -0,0 +1,187 @@
...
+  /** An optional field mask to apply to this read. */
+  fieldMask?: FieldPath[];
+  /** An optional transaction ID to use for this read. */
+  transactionId?: Uint8Array;
+
+  private outstandingDocuments = new Set<string>();
+  private retrievedDocuments = new Map<string, DocumentSnapshot>();
+
+  /**
+   * Internal method to retrieve multiple documents from Firestore, optionally
```
thebrianchen, line 43: nit: it seems weird to call the constructor an "internal method". Move this comment elsewhere?

thebrianchen (approved):
lgtm

dev/src/document-reader.ts
schmidt-sebastian: Oh, that's an old comment. Updated.

PR #1548, opened 2021-06-29
feat: add support for optimistic transactions

### LINKED ISSUE #7489: Support optimistic transactions as an option

I found that for some usage patterns optimistic transactions would work better than pessimistic.

I'd like to pass an option to `runTransaction` which would allow me to choose if pessimistic or optimistic transactions should be used.

There are workarounds for it, like manually fetching documents, checking preconditions, then starting a transaction and reading the documents again, but I find it unnecessarily verbose and having an optimistic transaction would make my code simpler.

 **Additional context**

None.

schmidt-sebastian:
This PR adds support for optimistic transactions. A `verify` write is added to the commit for all documents read during an optimistic transaction.
 
This PR relies on an backend feature that is not yet public. We need to figure out what our strategy here will be.

The first commit just re-arranges tests. It is probably best to only review the second commit.

Fixes https://github.com/googleapis/google-cloud-node/issues/7489

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -959,10 +959,9 @@ export class Firestore implements firestore.Firestore {
    * transaction.
    * @param {boolean=} optimisticLocking Whether to use optimistic locking.
    * By default, transactions lock documents pessimistically, which increases
-   * the chance that transactions succeed but blocks other clients from
-   * modifying the documents read during the transaction. When
-   * `optimisticLocking` is enabled, writes are not blocked. When other clients
-   * modify a document, the transaction is instead retried. Optimistic
+   * the chance that transactions succeed but blocks other clients`
+   * modifications. With `optimisticLocking`, writes are not blocked. When other
+   * clients modify a document, the transaction is instead retried. Optimistic
    * transactions do not support queries.
    * @param {number=} maxAttempts The maximum number of attempts for this
    * transaction. Defaults to five.
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -957,6 +957,13 @@ export class Firestore implements firestore.Firestore {
    * @template T
    * @param {false=} readOnly Set to false or omit to indicate a read-write
    * transaction.
+   * @param {boolean=} optimisticLocking Whether to use optimistic locking.
+   * By default, transactions lock documents pessimistically, which increases
+   * the chance that transactions succeed but blocks other clients from
+   * modifying the documents read during the transaction. When
+   * `optimisticLocking` is enabled, writes are not blocked. When other clients
+   * modify a document, the transaction is instead retried. Optimistic
+   * transactions do not support queries.
    * @param {number=} maxAttempts The maximum number of attempts for this
    * transaction. Defaults to five.
    */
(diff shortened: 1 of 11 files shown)


dev/src/transaction.ts
```diff
@@ -187,8 +191,29 @@ export class Transaction implements firestore.Transaction {
...
     const documentReader = new DocumentReader(this._firestore, documents);
     documentReader.fieldMask = fieldMask || undefined;
-    documentReader.transactionId = this._transactionId;
-    return documentReader.get(this._requestTag);
+
+    if (!this._optimisticLocking) {
+      documentReader.transactionId = this._transactionId;
+    }
+
+    return documentReader.get(this._requestTag).then(docs => {
```
thebrianchen, line 199: Do we have to handle the error case here? If we fail to fetch a document inside a tx, but the error is caught in the transaction callback, the final `commit()` won't have the preconditions set.

I think we need to chain the for loop after a `silencePromise()`.

dev/src/transaction.ts
```diff
@@ -187,8 +191,29 @@ export class Transaction implements firestore.Transaction {
...
+      for (const doc of docs) {
+        if (doc.exists) {
+          this._locks.set(
+            doc.ref.formattedName,
+            new Precondition({
+              lastUpdateTime: doc.updateTime,
+            })
+          );
+        } else {
+          this._locks.set(
```
thebrianchen, line 209: What's the expected behavior if an optimistic transaction reads the same document twice, but gets a different value each time?

My intuition says the tx should fail since the document value changed within the scope of the transaction, and if that's the intended behavior, we should check that the document isn't in the map.

dev/src/transaction.ts
```diff
@@ -609,6 +671,8 @@ function isRetryableTransactionError(error: GoogleError): boolean {
         // IDs that have expired. While INVALID_ARGUMENT is generally not
         // retryable, we retry this specific case.
         return !!error.message.match(/transaction has expired/);
+      case StatusCode.FAILED_PRECONDITION:
```
thebrianchen, line 674: question: why do we retry FAILED_PRECONDITION for optimistic locking?

dev/src/write-batch.ts
```diff
@@ -568,16 +568,17 @@ export class WriteBatch implements firestore.WriteBatch {
...
+   * @param commitOptions.preprocessor A function that can be used to mutate
+   * the request before the commit.
    * @returns  A Promise that resolves when this batch completes.
    */
   async _commit<Req extends api.ICommitRequest, Resp>(commitOptions?: {
-    transactionId?: Uint8Array;
     requestTag?: string;
     retryCodes?: number[];
     methodName?: FirestoreUnaryMethod;
+    preproccessor?: (request: api.ICommitRequest) => api.ICommitRequest;
```
thebrianchen, line 581: Now that I've grasped the PR, the preprocessor implementation is so clean ✨

dev/test/transaction.ts
```diff
@@ -1042,6 +1073,177 @@ describe('failed transactions', () => {
...
+          transaction.set(docRef, {});
+        });
+      },
+      begin(),
+      getDocument(/* transaction= */ ''),
+      commit(undefined, [set])
+    );
+  });
+
+  it('does not combines precondition if already set', () => {
```
thebrianchen, line 1149: ultranit: s/combines/combine :)

dev/test/transaction.ts
```diff
@@ -1042,6 +1073,177 @@ describe('failed transactions', () => {
...
+            {'a.b': 'c'},
+            {lastUpdateTime: new Timestamp(1, 2)}
+          );
+        });
+      },
+      begin(),
+      getDocument(/* transaction= */ ''),
+      commit(undefined, [set, verify])
+    );
+  });
```
thebrianchen, line 1200: optional (based on my above comments in transaction.ts): add test to check precondition if the same doc is read twice, add test to check that preconditions hold even if the document get() fails.

dev/src/index.ts
```diff
@@ -972,10 +978,16 @@ export class Firestore implements firestore.Firestore {
    * Transactions can be performed as read-only or read-write transactions. By
    * default, transactions are executed in read-write mode.
    *
-   * A read-write transaction obtains a pessimistic lock on all documents that
-   * are read during the transaction. These locks block other transactions,
-   * batched writes, and other non-transactional writes from changing that
-   * document. Any writes in a read-write transactions are committed once
+   * By default, read-write transactions obtain a pessimistic locks on all
```
thebrianchen, line 981: I'll create a google doc to rewrite this section and add you.

thebrianchen (commented):
thanks for taking the time to explain this to me!

dev/test/transaction.ts
schmidt-sebastian: Test added

thebrianchen (approved):
lgtm, just add the updated comments.

dev/src/transaction.ts
schmidt-sebastian: There should already be a comment (in the last diff).

schmidt-sebastian:
Changed this PR to specify a consistent read time once one has been obtained.

thebrianchen (approved):

PR #1549, opened 2021-06-30
fix: lower batch size on BulkWriter retry

thebrianchen:
Lowering batch size to 10 when retrying failed write in order to guarantee the batch is under 10MiB, as part of the plan outlined in go/bulkwriter-throughput. Will confirm with batchik@ about 9 vs 10 batch size before merging.

Storing `maxBatchSize` in `BulkCommitBatch` for easier tracking.


schmidt-sebastian (approved):
This is much simpler than anticipated. Thanks!

PR #1571, opened 2021-07-28
chore: log warning to console if BulkWriter operation fails

thebrianchen:


thebrianchen:
This PR is no longer needed since the unhandled failures throw an `UnhandledPromiseRejection` error.

PR #1572, opened 2021-07-28
feat: allow UnhandledPromiseRejection errors in BulkWriter if no error handler is specified

thebrianchen:
BulkWriter should throw an error if an operation promise is rejected without a custom error handler. Otherwise, failures are silently swallowed without any indication to the end-developer. If an error handler is specified with `onWriteError()`, BulkWriter assumes that the developer is handling the error and will swallow failed promises so that the developer does not have to add a dangling `catch()` statement for every promise.

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -416,6 +416,16 @@ export class BulkWriter {
    */
   private _bufferedOperations: Array<BufferedOperation> = [];
 
+  /**
+   * Whether a custom error handler has been set. BulkWriter only swallows
+   * errors if an error handler is set. Otherwise, an UnhandledPromiseRejection
+   * is thrown by Node if an operation promise is rejected without being
+   * handled.
+   * @private
+   * @internal
+   */
+  private _errorHandlerSet = false;
+
   // Visible for testing.
   _getBufferedOperationsCount(): number {
     return this._bufferedOperations.length;
@@ -558,7 +568,7 @@ export class BulkWriter {
     const op = this._enqueue(documentRef, 'create', bulkCommitBatch =>
       bulkCommitBatch.create(documentRef, data)
     );
-    silencePromise(op);
+    this._silencePromiseIfErrorHandlerSet(op);
     return op;
   }
 
@@ -598,7 +608,7 @@ export class BulkWriter {
     const op = this._enqueue(documentRef, 'delete', bulkCommitBatch =>
       bulkCommitBatch.delete(documentRef, precondition)
     );
-    silencePromise(op);
+    this._silencePromiseIfErrorHandlerSet(op);
     return op;
   }
 
@@ -655,7 +665,7 @@ export class BulkWriter {
     const op = this._enqueue(documentRef, 'set', bulkCommitBatch =>
       bulkCommitBatch.set(documentRef, data, options)
     );
-    silencePromise(op);
+    this._silencePromiseIfErrorHandlerSet(op);
     return op;
   }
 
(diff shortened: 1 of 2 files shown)


dev/src/bulk-writer.ts
```diff
@@ -1028,6 +1039,19 @@ export class BulkWriter {
...
 
+  /**
+   * Silences the promise if an error handler has been passed in. This prevents
+   * UnhandledPromiseRejections from being thrown if a floating BulkWriter
+   * operation promise fails.
+   * @private
+   * @internal
+   */
+  private _silencePromiseIfErrorHandlerSet(promise: Promise<unknown>): void {
+    if (this._errorHandlerSet) {
```
schmidt-sebastian, line 1050: Can we do this slightly later and make this decision when the operation returns? This would allow the following:

```
bulkWriter.set();
bulkWriter.set();
bulkWriter.onError(...);
```

Note that this error handler will always be attached before any result is returned.

thebrianchen: What's the use case for changing the error handler after an operation has been enqueued but before it resolves? To me, it seems like an anti-pattern, or an accidental bug.

schmidt-sebastian: It's guaranteed that this code executes before the request is processed, so I don't think we should forbid this. It is not possible for any network task to resolve before all synchronous code finished executing. I do think it is cleaner to attach the error handler first, but we do not enforce this, so I think we should allow this even if we believe this to be an anti-pattern.

dev/src/bulk-writer.ts
```diff
@@ -975,10 +978,24 @@ export class BulkWriter {
       this._successFn.bind(this)
     );
 
+    // Swallow the error if the developer has set an error listener. This
+    // prevents UnhandledPromiseRejections from being thrown if a floating
+    // BulkWriter operation promise fails when an error handler is specified.
+    //
+    // This is done here in order to chain the caught promise onto `lastOp`,
+    // which ensures that flush() resolves after the operation promise.
+    const returnedBulkWriterPromise = bulkWriterOp.promise.catch(err => {
```
schmidt-sebastian, line 987: This should probably use "op" instead of BulkWriter since it only affects a single operation.  

I would have also used the term "user" in this variable, but that may be the lack of oxygen at my current elevation (`userPromise`).

dev/test/bulk-writer.ts
```diff
@@ -54,6 +54,7 @@ import {
   verifyInstance,
 } from './util/helpers';
 import api = proto.google.firestore.v1;
+import {Deferred} from '../src/util';
```
schmidt-sebastian, line 57: Can you move this up to the other src imports?

dev/src/bulk-writer.ts
```diff
@@ -416,6 +416,16 @@ export class BulkWriter {
...
 
+  /**
+   * Whether a custom error handler has been set. BulkWriter only swallows
+   * errors if an error handler is set. Otherwise, an UnhandledPromiseRejection
+   * is thrown by Node if an operation promise is rejected without being
+   * handled.
+   * @private
+   * @internal
+   */
+  private _errorHandlerSet = false;
```
schmidt-sebastian, line 427: I don't think you need this. You can just do just do:

```
const DEFAULT_ERROR_FUNCTION = error => {
    const isRetryableDeleteError =
      error.operationType === 'delete' &&
      (error.code as number) === StatusCode.INTERNAL;
    const retryCodes = getRetryCodes('batchWrite');
    return (
      (retryCodes.includes(error.code) || isRetryableDeleteError) &&
      error.failedAttempts < MAX_RETRY_ATTEMPTS
    );
  };

private _errorFn = DEFAULT_ERROR_FUNCTION;

if (this._errorFn !== DEFAULT_ERROR_FUNCTION) {...}
```

Up to you.

schmidt-sebastian (approved):
Thank you. Some final words but this is basically good to go.

PR #1575, opened 2021-08-02
fix: avoid destructuring undefined timestamps

### LINKED ISSUE #1574: Trigger fails due to destructuring undefined
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.

  - Search the issues already opened: https://github.com/googleapis/nodejs-firestore/issues
  - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: macOS 11.4
  - Node.js version: v14.17.0
  - npm version: v6.14.13
  - `@google-cloud/firestore` version: 4.14.0

#### Steps to reproduce

Sadly we are not sure how to reproduce as we cannot get the document id that causes this issue. Here is a provided stack trace of the issue.
```javascript
TypeError: Cannot read property 'seconds' of undefined at Function.fromProto (/workspace/node_modules/@google-cloud/firestore/build/src/timestamp.js:136:47)
at Firestore.snapshot_ (/workspace/node_modules/@google-cloud/firestore/build/src/index.js:715:57)
at beforeSnapshotConstructor (/workspace/node_modules/firebase-functions/lib/providers/firestore.js:139:30)
at changeConstructor (/workspace/node_modules/firebase-functions/lib/providers/firestore.js:143:49)
at cloudFunction (/workspace/node_modules/firebase-functions/lib/cloud-functions.js:133:34)
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

Sahtair:
In `https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/index.ts#L911` we call from proto with a potential undefined instead of an object. This causes an error in `https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/timestamp.ts#L122` where it tries to destructure an undefined value. This problem causes triggers to fail before entering user defined handler.

Fixes #1574 🦕

### DIFF
--- a/dev/src/convert.ts
+++ b/dev/src/convert.ts
@@ -49,10 +49,7 @@ export function timestampFromJson(
   timestampValue?: string | google.protobuf.ITimestamp,
   argumentName?: string
 ): google.protobuf.ITimestamp | undefined {
-  let timestampProto: google.protobuf.ITimestamp = {
-      seconds: undefined,
-      nanos: undefined
-  };
+  let timestampProto: google.protobuf.ITimestamp = {};
 
   if (typeof timestampValue === 'string') {
     const date = new Date(timestampValue);
--- a/dev/src/convert.ts
+++ b/dev/src/convert.ts
@@ -49,7 +49,10 @@ export function timestampFromJson(
   timestampValue?: string | google.protobuf.ITimestamp,
   argumentName?: string
 ): google.protobuf.ITimestamp | undefined {
-  let timestampProto: google.protobuf.ITimestamp | undefined;
+  let timestampProto: google.protobuf.ITimestamp = {
+      seconds: undefined,
+      nanos: undefined
+  };
 
   if (typeof timestampValue === 'string') {
     const date = new Date(timestampValue);


Sahtair:
@googlebot I signed it!

dev/src/convert.ts
```diff
@@ -49,7 +49,10 @@ export function timestampFromJson(
   timestampValue?: string | google.protobuf.ITimestamp,
   argumentName?: string
 ): google.protobuf.ITimestamp | undefined {
-  let timestampProto: google.protobuf.ITimestamp | undefined;
+  let timestampProto: google.protobuf.ITimestamp = {
+      seconds: undefined,
+      nanos: undefined
+  };
```
schmidt-sebastian, line 55:
```suggestion
  let timestampProto: google.protobuf.ITimestamp = {};
```

schmidt-sebastian (approved):
Thank you for sending this. We might be able to simplify this if you are keen :)

Sahtair:
@schmidt-sebastian Yes I would agree we can simplify. I was not sure what the convention was so I did the verbose way :) Thank you for the review!

PR #1582, opened 2021-08-05
temporary types exploration

thebrianchen:
TODO:
- Support `WithFieldValue` typings in nested `TypedUpdateData` fields
- Only allow `FieldValue.delete()` in optional fields.

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -2473,14 +2473,48 @@ describe('WriteBatch class', () => {
       {address: {nestedStringArray: FieldValue.arrayUnion('baz')}},
       {merge: true}
     );
+    await ref.set(
+      {address: {nestedStringArray: FieldValue.arrayRemove('baz')}},
+      {merge: true}
+    );
 
     // These should error.
     await ref.set({numberArray: ['not-number']}, {merge: true});
     await ref.set({number: 'not-number'}, {merge: true});
     await ref.set({number: null}, {merge: true});
     await ref.set({address: {nestedStringArray: 'not-array'}}, {merge: true});
+    await ref.set(
+      {address: {nestedStringArray: FieldValue.serverTimestamp()}},
+      {merge: true}
+    );
 
-    const doc = await ref.get();
+    // NumericFieldValue Allowed
+    await ref.set({number: FieldValue.increment(1)}, {merge: true});
+    await ref.set({number: FieldValue.delete()}, {merge: true});
+
+    // NumericFieldValue Error
+    await ref.set({number: FieldValue.arrayUnion(4)}, {merge: true});
+    await ref.set({number: FieldValue.arrayRemove(4)}, {merge: true});
+    await ref.set({number: FieldValue.serverTimestamp()}, {merge: true});
+
+    // TimestampFieldValue Allowed
+    await ref.set({timestamp: FieldValue.serverTimestamp()}, {merge: true});
+    await ref.set({timestamp: FieldValue.delete()}, {merge: true});
+
+    // TimestampFieldValue Error
+    await ref.set({timestamp: FieldValue.increment(1)}, {merge: true});
+    await ref.set({timestamp: FieldValue.arrayUnion(4)}, {merge: true});
+    await ref.set({timestamp: FieldValue.arrayRemove(4)}, {merge: true});
+    await ref.set({timestamp: 3}, {merge: true});
+
+    // ArrayFieldValue Allowed
+    await ref.set({numberArray: FieldValue.arrayUnion(123)}, {merge: true});
+    await ref.set({numberArray: FieldValue.arrayRemove(12345)}, {merge: true});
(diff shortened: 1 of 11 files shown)


types/firestore.d.ts
```diff
@@ -50,6 +50,37 @@ declare namespace FirebaseFirestore {
...
+
+  /** Like Partial but recursive */
+  export type NestedPartial<T> = T extends Builtin
+    ? T
+    : T extends Map<infer K, infer V>
+    ? Map<NestedPartial<K>, NestedPartial<V>>
+    : T extends Array<infer U>
+    ? Array<NestedPartial<U>>
+    : T extends {}
+    ? {[K in keyof T]?: NestedPartial<T[K] | WithFieldValue<T[K]>>}
```
Feiyang1, line 81: `WithFieldValue<T[K]>` includes `T[K]` already.
```suggestion
    ? {[K in keyof T]?: NestedPartial<WithFieldValue<T[K]>>}
```
Do you want to accept partial `FieldValue`s? It feels a little weird, but probably doesn't matter because no one will pass partial `FieldValue`s in practice. You could add a branch for `FieldValue`s and return them as is like the `Builtin`s. Up to you.

Feiyang1 (commented):
Looks good. I don't think any of the changes are breaking.

types/firestore.d.ts
```diff
@@ -42,6 +42,40 @@ declare namespace FirebaseFirestore {
...
+  // Represents an update object to Firestore document data, which can contain either fields like {a: 2}
+  // or dot-separated paths such as {"a.b" : 2} (which updates the nested property "b" in map field "a").
+  export type TypedUpdateData<T extends Record<string, any>> = {
+    [K in keyof T]?: WithFieldValue<T[K]>;
+  } &
+    NestedUpdateFields<T>;
+
+  // For each field (e.g. "bar"), calculate its nested keys (e.g. {"bar.baz": T1, "bar.quax": T2}), and then
+  // intersect them together to make one giant map containing all possible keys (all marked as optional).
+  export type NestedUpdateFields<T extends Record<string, any>> =
```
schmidt-sebastian, line 54: Does this support custom types (via converter)?

dev/system-test/firestore.ts
```diff
@@ -2447,6 +2458,89 @@ describe('WriteBatch class', () => {
...
+
+    // Non-special type Allowed
+    await ref.set({name: FieldValue.delete()}, {merge: true});
+
+    // Non-special type Error
+    await ref.set({name: FieldValue.increment(12345)}, {merge: true});
+    await ref.set({name: FieldValue.serverTimestamp()}, {merge: true});
+    await ref.set({name: FieldValue.arrayUnion(123)}, {merge: true});
+    await ref.set({name: FieldValue.arrayRemove(12345)}, {merge: true});
+  });
```
schmidt-sebastian, line 2542: You should add some tests that fail validation with your new code - https://dev.to/oliverjumpertz/making-good-use-of-ts-expect-error-in-typescript-1f41

types/firestore.d.ts
```diff
@@ -50,6 +84,35 @@ declare namespace FirebaseFirestore {
...
+    | undefined
+    | null;
+  export type Builtin = Primitive | Function;
+
+  /** Like Partial but recursive */
+  export type NestedPartial<T> = T extends Builtin
+    ? T
+    : T extends Map<infer K, infer V>
+    ? Map<NestedPartial<K>, NestedPartial<V>>
+    : T extends Array<infer U>
```
schmidt-sebastian, line 110: We don't support updating sub-arrays. Can we remove this?

types/firestore.d.ts
```diff
@@ -50,6 +84,35 @@ declare namespace FirebaseFirestore {
...
+    ? T | ArrayFieldValue | DeleteFieldValue
+    : T | DeleteFieldValue;
+
+  export type Primitive =
+    | string
+    | number
+    | boolean
+    | bigint
+    | symbol
```
schmidt-sebastian, line 100: Do we need `symbol` for Firestore?

types/firestore.d.ts
```diff
@@ -97,7 +160,10 @@ declare namespace FirebaseFirestore {
      * toFirestore() must be defined with `Partial<T>`.
      */
     toFirestore(modelObject: T): DocumentData;
```
schmidt-sebastian, line 162: This has to support FieldValue as well.

types/firestore.d.ts
```diff
@@ -97,7 +160,10 @@ declare namespace FirebaseFirestore {
...
-    toFirestore(modelObject: Partial<T>, options: SetOptions): DocumentData;
+    toFirestore(
+      modelObject: NestedPartial<T>,
```
schmidt-sebastian, line 164: I think this is also a breaking change since you are widening the type of a function that has to be implemented by the user.

schmidt-sebastian (commented):
This looks promising, but I would need to see a lot of different tests to validate this. Thanks for taking it this far!

PR #1584, opened 2021-08-05
chore: migrate to Owl Bot

SurferJeffAtGoogle:
Attempt 3.


bcoe:
@SurferJeffAtGoogle this looks really healthy, I think the problem is just the replace rules are perhaps interfering with linting? Do we only want to perform the replace when new code is added?

PR #1592, opened 2021-08-20
chore!: upgrade typescript to v4.1.5

thebrianchen:


dev/system-test/firestore.ts
```diff
@@ -444,7 +444,8 @@ describe('DocumentReference class', () => {
   });
 
   it('has set() method', () => {
-    const allSupportedTypesObject = {
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    const allSupportedTypesObject: {[field: string]: any} = {
```
schmidt-sebastian, line 448: This can probably be `unknown`.

dev/test/index.ts
```diff
@@ -262,7 +262,8 @@ const allSupportedTypesInput = {
   bytesValue: Buffer.from([0x1, 0x2]),
 };
 
-const allSupportedTypesOutput = {
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const allSupportedTypesOutput: {[field: string]: any} = {
```
schmidt-sebastian, line 266: This can probably be `unknown`.

schmidt-sebastian (commented):
I wonder if this needs to be `chore!` just to play it safe. This would trigger a major version bump.

dev/test/index.ts
thebrianchen: We need `any` here since we access `pathValue.formattedName` on L680, which would would error since `pathValue` has `unknown` type.

dev/system-test/firestore.ts
thebrianchen: We need `any` since we access `allSupportedTypesObject.pathValue.path` on L478.

dev/test/index.ts
schmidt-sebastian: You can cast there if it happens just one. `unknown as X` is more type safe than `any`.

thebrianchen: Good point, done for both.

dev/test/index.ts
```diff
@@ -676,7 +676,8 @@ describe('snapshot_() method', () => {
     // Deep Equal doesn't support matching instances of DocumentRefs, so we
     // compare them manually and remove them from the resulting object.
     expect(actualObject.get('pathValue').formattedName).to.equal(
-      expected.pathValue.formattedName
+      /* eslint-disable @typescript-eslint/no-explicit-any */
+      (expected.pathValue as any).formattedName
```
schmidt-sebastian, line 680: `as DocumentReference`

dev/system-test/firestore.ts
```diff
@@ -474,7 +474,8 @@ describe('DocumentReference class', () => {
       .then(doc => {
         const data = doc.data()!;
         expect(data.pathValue.path).to.equal(
-          allSupportedTypesObject.pathValue.path
+          /* eslint-disable @typescript-eslint/no-explicit-any */
+          (allSupportedTypesObject.pathValue as any).path
```
schmidt-sebastian, line 478: `as IDontKnowWhatTypeThisIsButWeCanFindOut`

dev/test/index.ts
thebrianchen: Sorry for being hasty 😨

schmidt-sebastian (approved):
LGTM.

Let's not merge this until the follow-up PR is ready and we know how to make patch releases for prior versions.

thebrianchen:
> LGTM.
> 
> Let's not merge this until the follow-up PR is ready and we know how to make patch releases for prior versions.

Yep, I'm merging this into a 4.1 feature branch that I'll add the other changes to.

PR #1595, opened 2021-08-24
feat!: add QoL improvements for converters

### LINKED ISSUE #1448: Make UpdateData type Generic to fix lack of type safety when using withConverter()
This package works great with Typescript when using the `.withConverter` method and enforces correct types when performing a `get()` or `set()` or `onSnapshot()`, but fails to enforce types on the `update()` method.

With TS v4.1 we can now safely enforce that dot-notation strings as object keys are correct, so we can do better than:

```ts
export type UpdateData = {[fieldPath: string]: any};
```

Check it out...

```ts
import { firestore } from "firebase-admin"

interface User {
  name: string
  email: string
  address: {
    line1: string
    line2: string
    postcode: string
    verified: boolean
    timeAtAddress: {
      days: string
      months: string
      hours: string
    }
  }
}

// This helper function pipes my types through the converter
const converter = <T>() => ({
  toFirestore: (data: Partial<T>) => data,
  fromFirestore: (snap: FirebaseFirestore.QueryDocumentSnapshot) => snap.data() as T
})

const getUser = async () => {
  const userDoc = await firestore().collection('user').withConverter(converter<User>()).doc('1234').get()
  if (!userDoc.exists) return null
  const userData = userDoc.data()
  console.log(userData.email) // nice!
  console.log(userData.unknownKey)
  // TS Error: Property 'unknownKey' does not exist on type 'User'.
  // Great!
}

const setUser = async () => {
  await firestore().collection('user').withConverter(converter<User>()).doc('1234').set({
    email: '',
    name: ''
  })
  // TS Error: Property 'address' is missing in type '{ email: string; name: string; }' but required in type 'User'
  // Great!
}

const setUserMerged = async () => {
  await firestore().collection('user').withConverter(converter<User>()).doc('1234').set({
    email: '',
    name: '',
    unknownKey: ''
  }, { merge: true })
  // TS Error: Object literal may only specify known properties, and 'unknownKey' does not exist in type 'Partial<User>'.
  // Great!
}

const updateUserUnsafe = async () => {
  await firestore().collection('user').withConverter(converter<User>()).doc('1234').update({
    unknownKey: '', // no error here
    'anything.I.like': true // no type error
  })
}

// Here's how we could do an update with type safety:

type PathImpl<T, K extends keyof T> =
  K extends string
  ? T[K] extends Record<string, any>
  ? T[K] extends ArrayLike<any>
  ? K | `${K}.${PathImpl<T[K], Exclude<keyof T[K], keyof any[]>>}`
  : K | `${K}.${PathImpl<T[K], keyof T[K]>}`
  : K
  : never

type Path<T> = PathImpl<T, keyof T> | keyof T

type PathValue<T, P extends Path<T>> =
  P extends `${infer K}.${infer Rest}`
  ? K extends keyof T
  ? Rest extends Path<T[K]>
  ? PathValue<T[K], Rest>
  : never
  : never
  : P extends keyof T
  ? T[P]
  : never

type CustomUpdateData<T extends object> = Partial<{
  [TKey in Path<T>]: PathValue<T, TKey>
}>

const updateUserSafe = async () => {
  // Here I make the object first which provides some safety
  const updatesObject: CustomUpdateData<User> = {
    name: '',
    'address.timeAtAddress.days': '',
    'some.unknown.path': false,
    // Object literal may only specify known properties, and ''some.unknown.path'' does not exist in type 'Partial<CustomUpdateData<User>>'
    'address.verified': 'Oops! Wrong value type!'
    // Type 'string' is not assignable to type 'boolean'
  }

  await firestore().collection('user').withConverter(converter<User>()).doc('1234').update(updatesObject)
}
```

I propose that the helper types for creating a type that is able to handle object paths should be added to `FirebaseFirestore.UpdateData`. I think that a solution could look like this:

```ts
// @google-cloud/firestore/types/firestore.d.ts

type PathImpl<T, K extends keyof T> =
  K extends string
  ? T[K] extends Record<string, any>
  ? T[K] extends ArrayLike<any>
  ? K | `${K}.${PathImpl<T[K], Exclude<keyof T[K], keyof any[]>>}`
  : K | `${K}.${PathImpl<T[K], keyof T[K]>}`
  : K
  : never

type Path<T> = PathImpl<T, keyof T> | keyof T

type PathValue<T, P extends Path<T>> =
  P extends `${infer K}.${infer Rest}`
  ? K extends keyof T
  ? Rest extends Path<T[K]>
  ? PathValue<T[K], Rest>
  : never
  : never
  : P extends keyof T
  ? T[P]
  : never
  
declare namespace FirebaseFirestore {
  export type DocumentData = { [field: string]: any }

  export type UpdateData<T = DocumentData> = Partial<{
    [TKey in Path<T>]: PathValue<T, TKey>
  }>

  // ...
  
  export class DocumentReference<T = DocumentData> {
    // ...
    update(data: UpdateData<T>, precondition?: Precondition): Promise<WriteResult>
  }
}

```

Which would allow type safety like this:

```ts
const updateSafe = async () => {
  await firestore().collection('user').withConverter(converter<User>()).doc('1234').update({
    name: '',
    'address.line1': '',
    'some.unknown.path': false,
    // Object literal may only specify known properties, and ''some.unknown.path'' does not exist in type 'Partial<CustomUpdateData<User>>'
    'address.verified': 'Oops! Wrong value type!'
    // Type 'string' is not assignable to type 'boolean'
  })
}
```

 **Additional context**
Proposed solution in action:
Screenshot 2021-03-12 at 12 42 17

thebrianchen:
Porting from [web](https://github.com/firebase/firebase-js-sdk/pull/5268).

Fixes #1448.

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -3159,9 +3159,9 @@ describe('Types test', () => {
       // Check top-level fields.
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           outerString: 3,
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           outerArr: null,
         },
         {merge: true}
@@ -3172,10 +3172,10 @@ describe('Types test', () => {
         {
           nested: {
             innerNested: {
-              // @ts-expect-error Should fail to compile.
+              // @ts-expect-error Should fail to transpile.
               innerNestedNum: 'string',
             },
-            // @ts-expect-error Should fail to compile.
+            // @ts-expect-error Should fail to transpile.
             innerArr: null,
           },
         },
@@ -3183,7 +3183,7 @@ describe('Types test', () => {
       );
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           nested: 3,
         },
         {merge: true}
@@ -3195,7 +3195,7 @@ describe('Types test', () => {
       // Top-level property.
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           nonexistent: 'foo',
         },
         {merge: true}
@@ -3205,7 +3205,7 @@ describe('Types test', () => {
       await ref.set(
         {
           nested: {
-            // @ts-expect-error Should fail to compile.
+            // @ts-expect-error Should fail to transpile.
             nonexistent: 'foo',
           },
         },
@@ -3266,9 +3266,9 @@ describe('Types test', () => {
       // Check top-level fields.
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           outerString: 3,
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           outerArr: null,
         },
         {merge: true}
@@ -3279,10 +3279,10 @@ describe('Types test', () => {
         {
           nested: {
             innerNested: {
-              // @ts-expect-error Should fail to compile.
+              // @ts-expect-error Should fail to transpile.
               innerNestedNum: 'string',
             },
-            // @ts-expect-error Should fail to compile.
+            // @ts-expect-error Should fail to transpile.
             innerArr: null,
           },
         },
@@ -3290,7 +3290,7 @@ describe('Types test', () => {
       );
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           nested: 3,
         },
         {merge: true}
@@ -3302,7 +3302,7 @@ describe('Types test', () => {
       // Top-level property.
       await ref.set(
         {
-          // @ts-expect-error Should fail to compile.
+          // @ts-expect-error Should fail to transpile.
           nonexistent: 'foo',
         },
         {merge: true}
@@ -3312,7 +3312,7 @@ describe('Types test', () => {
       await ref.set(
         {
           nested: {
-            // @ts-expect-error Should fail to compile.
+            // @ts-expect-error Should fail to transpile.
             nonexistent: 'foo',
           },
         },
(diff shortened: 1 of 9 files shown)


dev/system-test/firestore.ts
```diff
@@ -3045,3 +3055,559 @@ describe('Bundle building', () => {
...
+          },
+          innerArr: FieldValue.arrayUnion(3),
+        },
+      });
+    });
+
+    it('validates inner and outer fields', async () => {
+      const ref = doc.withConverter(testConverter);
+      await ref.update({
+        // @ts-expect-error Should fail to compile.
```
schmidt-sebastian, line 3430: s/compile/transpile :)

types/firestore.d.ts
```diff
@@ -27,10 +27,31 @@ declare namespace FirebaseFirestore {
...
+    : T extends {}
+    ? {[K in keyof T]: WithFieldValue<T[K]> | FieldValue}
+    : Partial<T>;
+
   /**
    * Update data (for use with [update]{@link DocumentReference#update})
    * that contains paths mapped to values. Fields that contain dots reference
-   * nested fields within the document.
+   * nested fields within the document. FieldValues can be passed in
+   * as property values.
```
schmidt-sebastian, line 54: Thinking out loud: Is "property" the right term and does it make it harder to understand? Should we drop it and just call it "values"?

schmidt-sebastian (approved):
Thanks for doing this!

types/firestore.d.ts
thebrianchen: I wanted to specify "property values" to convey that FieldValues can be set as the value of an object property, but not as the entire object value.

Jeandcc:
Apologies for pinging here, I'm just a bit confused by all the issues that reference this same "need". Has this been published to any version of the "firebase-admin" package, or is it still in the pipeline to be released?

Thanks!

schmidt-sebastian:
This has not been published. We are still working on some details for the implementation that we published for the `@firebase/firestore` Web SDK and will publish the Node version once these changes have been finalized.

oxc:
I'm a bit confused by the changes implemented for the update() method signature. The update() method does, to the best of my knowledge, **not** call the converter. Its parameters will not be converted, and must match the "storage format". So it should in fact not be an `UpdateData<T>`, but an `UpdateData<D>` where `D` is the theoretical type that reflects the stored data (not the converted one).

Am I missing something here?

PR #1599, opened 2021-08-26
fix: Handles identical document ids from different collections.

### LINKED ISSUE #1598: Bundles use id as unique identifier however id unique-ness is not enforced across all collections
When adding documents to a bundle the documents id (doc.id) is used to identify the document preventing duplicates being added. However id unique-ness is not enforced across all collections, eg both collection1/doc and collection2/doc share the same id. 

The code for this can be found here https://github.com/googleapis/nodejs-firestore/blob/2522715d4eafd535ec0855000fb21a89e8396041/dev/src/bundle.ts#L85

Shouldnt the path be used to test for the same document being added multiple times?

wu-hui:
Fixes https://github.com/googleapis/nodejs-firestore/issues/1598


schmidt-sebastian (approved):
Thanks for the quick fix!

willcalderbank:
Thanks for the quick fix! Great work!

PR #1604, opened 2021-09-10
chore: upgrade typescript to v4.4.2

thebrianchen:
Leaving the typescript at `^4.1.5` causes compiler errors when v4.4.2 is installed due to v.4.4's breaking change where errors in catch blocks default to `unknown` type ([documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#using-unknown-in-catch-variables)). This can cause `npm install` to fail if a 4.4.2 is the installed.

As a quick fix, I've added a compiler option to remove this functionality. There are several changes that would need to be made in order to remove the flag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "proxyquire": "^2.1.3",
     "sinon": "^11.0.0",
     "ts-node": "^10.0.0",
-    "typescript": "^4.1.5",
+    "typescript": "^4.4.3",
     "through2": "^4.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


schmidt-sebastian (approved):
How much work would a fix for these catch statements be? We probably just need to add a few ":Error" statements.

thebrianchen:
> How much work would a fix for these catch statements be? We probably just need to add a few ":Error" statements.

It's a bit more involved since TS change forces usage of `catch (e)` to be explicitly typed with `catch(e: unknown)` or `catch(e: any)`. We have multiple parts in the code where we assume the thrown error is an `Error` or `GoogleError` object and try to access `error.code` or `err.message`. directly. Probably won't take too long, but I'll look into a fix for this later.

PR #1618, opened 2021-10-14
fix: make SetOptions a union type

thebrianchen:
Making SetOptions match the web SDK.
- Remove some old validation logic
- Add test to check that validation is being run on individual field path strings.
- Kept the check against specifying `merge` and `mergeFields` in case there are JS users who somehow specify both (open to deleting that if you want)


schmidt-sebastian (approved):
Thanks

PR #1631, opened 2021-11-10
fix(cloud-rad): move comments for TSDoc

fhinkel:
As we move our ref docs to cloud.google.com, we rely on TSDoc rather JSDoc.

TSDoc expects comments before the first overloaded function, we
currently have those on the last function. Those comments don't make
it into the d.ts files. We need to move comments to the first
overloaded function rather than the last one.

Internally b/190631834

Script used: https://github.com/fhinkel/cloud-rad-script/blob/main/moveComments.js

### DIFF
--- a/dev/src/bundle.ts
+++ b/dev/src/bundle.ts
@@ -42,8 +42,6 @@ export class BundleBuilder {
 
   constructor(readonly bundleId: string) {}
 
-  add(documentSnapshot: DocumentSnapshot): BundleBuilder;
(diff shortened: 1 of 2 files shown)


meredithslota:
This had some unintended consequences in nodejs-bigquery, see: https://github.com/googleapis/nodejs-bigquery/issues/1068 — I'm not sure if we need to tweak the approach or not, but please check the generated docs after merging to ensure we're not accidentally removing things.

PR #1639, opened 2021-12-01
fix: save negative zero as doubleValue

### LINKED ISSUE #1638: Firestore negative zero (-0) is stored as integerValue instead of doubleValue
This issue is about an inconsistency between the firestore JS SDK and the nodejs SDK. Firestore JS SDK stores negative zero (-0) has a double value with the correct sign. The NodeJS SDK does not do so. This leads to loss of data (the number's sign) as well as some odd behavior when trying to query documents saved from the JS SDK based on -0.

#### Environment details

  - OS: macOS
  - Node.js version: 16.13
  - npm version: 8.1
  - `@google-cloud/firestore` version: 4.15.1

#### Steps to reproduce

  1. write negative zero to a document

```js
await firestore.collection('test').doc('test').set({ number: -0 })
```

  2. read the document and ensure the value is negative zero

```js
const doc = await firestore.collection('test').doc('test').get();
const expectedToBeNegativeZero = doc.data().number;
console.assert(Object.is(expectedToBeNegativeZero, -0))
```

Firebase Web SDK: ✅ Works as expected
Firebase Node.JS SDK: ❌ Value is 0 instead of -0

Yonom:
# Why?

Javascript only has a number type while firestore differentiates between int64 and double. Both the Web SDK and the nodejs SDK identify integers to store them appropriately. However, the algorithms between the Web SDK and the nodejs SDK differ in one particular case: the negative zero double value.

Relevant lines of code from the JS SDK:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/number_serializer.ts#L56
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/util/types.ts#L44

This PR aligns the nodejs SDK behavior with the one in the Web SDK.

# Template

- [ ] ~~Appropriate docs were updated (if necessary)~~

Fixes #1638 🦕

### DIFF
--- a/dev/src/serializer.ts
+++ b/dev/src/serializer.ts
@@ -118,7 +118,7 @@ export class Serializer {
     }
 
     if (typeof val === 'number') {
-      const isNegativeZero = value === 0 && 1 / value === -Infinity;
+      const isNegativeZero = val === 0 && 1 / val === -Infinity;
       if (isNegativeZero) {
         return {
           doubleValue: -0,
--- a/dev/src/serializer.ts
+++ b/dev/src/serializer.ts
@@ -118,7 +118,12 @@ export class Serializer {
     }
 
     if (typeof val === 'number') {
-      if (Number.isSafeInteger(val)) {
+      const isNegativeZero = value === 0 && 1 / value === -Infinity;
+      if (isNegativeZero) {
+        return {
+          doubleValue: -0,
+        };
+      } else if (Number.isSafeInteger(val)) {
         return {
           integerValue: val as number,
         };
--- a/dev/src/serializer.ts
+++ b/dev/src/serializer.ts
@@ -118,12 +118,8 @@ export class Serializer {
     }
 
     if (typeof val === 'number') {
-      const isNegativeZero = val === 0 && 1 / val === -Infinity;
-      if (isNegativeZero) {
-        return {
-          doubleValue: -0,
-        };
-      } else if (Number.isSafeInteger(val)) {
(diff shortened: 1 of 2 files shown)


dconeybe:
Thanks for the PR! Could you (a) sign the CLA (see `CONTRIBUTING.md`) and (b) add a test for this case to `dev/test/document.ts`?

dconeybe:
You can ignore the "Auto-approve.yml check" failed check.

This looks good to me, but I'm going to get my colleague with more experience in this particular GitHub repository to provide formal approval.

Thanks for taking the time to submit this fix!

dconeybe:
@schmidt-sebastian Can you take a look at this PR? It LGTM but wanted a 2nd opinion.

Yonom:
Awesome. Thank you for the quick review! 😊

dev/src/serializer.ts
```diff
@@ -118,7 +118,12 @@ export class Serializer {
     }
 
     if (typeof val === 'number') {
-      if (Number.isSafeInteger(val)) {
+      const isNegativeZero = val === 0 && 1 / val === -Infinity;
```
schmidt-sebastian, line 121: Just for our future sanity, can we use the exact same check that we use on the Web SDK? https://github.com/firebase/firebase-js-sdk/blob/254137540a0e73d357a9a8cd27c5826108679f7b/packages/firestore/src/util/types.ts#L37

Otherwise this LGTM.

Yonom: Done :)

schmidt-sebastian (approved):

PR #1648, opened 2021-12-21
docs: improve documentation for merge behavior

### LINKED ISSUE #987: [Firestore] API Feedback – Empty Object Merge Behavior
### Environment

  * Operating System version: macOS Catalina 10.15.2
  * Firebase SDK version: 7.17.2
  * Firebase Product: Firestore
  * Node.js version: 12.14.1
  * NPM version: 6.14.7

### Problem

It's been discussed in more than a few other places, but the current behavior of `firestore.set('any/document'. { property: {} }, { merge: true })` is unintuitive and dangerous!

ref: https://github.com/firebase/firebase-js-sdk/issues/1371#issuecomment-445066500
ref:  https://github.com/firebase/firebase-admin-node/issues/365
ref: https://github.com/mesqueeb/vuex-easy-firestore/issues/73
ref: https://github.com/firebase/firebase-js-sdk/issues/1371
ref: https://github.com/firebase/firebase-js-sdk/issues/1168

We nearly had a production error because of this database API quirk – we caught it through sheer luck. I wanted to flag it again and make the case for changing the behavior in a future major release.

## The Good
```javascript
firestore.set('any/document'. { property: { foo: 'bar' } }, { merge: true });
firestore.set('any/document'. { property: { biz: 'baz' } }, { merge: true });
console.log(await firestore.doc('any/document').get()).data());
// Expected Log: { foo: 'bar', biz: 'baz' }  👍 
// Actual Log: { foo: 'bar', biz: 'baz' }    😎 
```
## The Bad
```javascript
firestore.set('any/document'. { property: { foo: 'bar' } }, { merge: true });
firestore.set('any/document'. { property: {} }, { merge: true });
console.log(await firestore.doc('any/document').get()).data());
// Expected Log: { foo: 'bar' } 👍 
// Actual Log: { } 😱 
```

This is a very easy mistake to make, especially since it it not clearly called out in the docs, and depending on the setup a system may remove keys to patch update based on user input. This behavior forces extra (expensive!) input validation on deep objects that should not be necessary.

I would expect that **any** destructive operations with `{ merge: true }` must be explicitly set using `FieldValue.delete()`.

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-admin-node/issues/987

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -645,10 +645,12 @@ export class BulkWriter {
    * @param {SetOptions=} options An object to configure the set behavior.
    * @param {boolean=} options.merge - If true, set() merges the values
    * specified in its data argument. Fields omitted from this set() call remain
-   * untouched.
+   * untouched. If your input sets any field to an empty map, all nested fields
+   * at this location are overwritten.
    * @param {Array.<string|FieldPath>=} options.mergeFields - If provided, set()
    * only replaces the specified field paths. Any field path that is not
-   * specified is ignored and remains untouched.
+   * specified is ignored and remains untouched. If your input sets any field to
+   * an empty map, all nested fields at this location are overwritten.
    * @returns {Promise<WriteResult>} A promise that resolves with the result of
    * the write. If the write fails, the promise is rejected with a
    * [BulkWriterError]{@link BulkWriterError}.
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -646,11 +646,11 @@ export class BulkWriter {
    * @param {boolean=} options.merge - If true, set() merges the values
    * specified in its data argument. Fields omitted from this set() call remain
    * untouched. If your input sets any field to an empty map, all nested fields
-   * at this location are overwritten.
+   * are overwritten.
    * @param {Array.<string|FieldPath>=} options.mergeFields - If provided, set()
    * only replaces the specified field paths. Any field path that is not
    * specified is ignored and remains untouched. If your input sets any field to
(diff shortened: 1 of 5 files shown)


dev/src/bulk-writer.ts
```diff
@@ -645,10 +645,12 @@ export class BulkWriter {
...
    * @param {boolean=} options.merge - If true, set() merges the values
    * specified in its data argument. Fields omitted from this set() call remain
-   * untouched.
+   * untouched. If your input sets any field to an empty map, all nested fields
+   * at this location are overwritten.
    * @param {Array.<string|FieldPath>=} options.mergeFields - If provided, set()
    * only replaces the specified field paths. Any field path that is not
-   * specified is ignored and remains untouched.
+   * specified is ignored and remains untouched. If your input sets any field to
+   * an empty map, all nested fields at this location are overwritten.
```
thebrianchen, line 653: I was initially confused since I inferred "location" to mean the document being set. The ambiguous interpretations:
- all nested fields in the document location are overwritten with empty map, but top level fields remain
- if the specified field has a nested field, it's deleted, but if the specified field maps to a value, it's not.

alternative suggestion: "if your input sets any field to an empty map, the specified field is (overwritten/set to) an empty map.

thebrianchen (approved):

dev/src/bulk-writer.ts
schmidt-sebastian: You are right, this is a bit confusing. I don't like how your example uses "empty map" twice, but I think we can achieve the same by just removing "at this location". What do you think?

thebrianchen: Yeah, that works!

PR #1654, opened 2022-01-05
fix: remove serializer check from Query.isEqual()

### LINKED ISSUE #1652: isEqual() not working correctly for Queries
#### Environment details

  - OS: macOS 11.6 (Big Sur)
  - Node.js version: 14.18.0
  - yarn version: 1.22.15
  - `firebase-admin` version: 10.0.1 (encountered the issue whilst using Firebase Admin)
  - `@google-cloud/firestore` version: 5.0.1 (tested this version to verify that the issue was from @google-cloud/firestore)

#### Steps to reproduce

Calling `isEqual` for two queries with the same options should return `true`, but it currently returns `false`.

```ts
import { Firestore } from "@google-cloud/firestore";

const queryA = firestore.collection("docs").where("field", ">", 1);
const queryB = firestore.collection("docs").where("field", ">", 1);
console.log(queryA.isEqual(queryB));
// Expected output: true
// Actual output: false
```

This does **not** happen for collection or document references - the `isEqual` method for those work fine.

```ts
const docA = firestore.doc("doc/1");
const docB = firestore.doc("doc/1");
console.log("docA == docB", docA.isEqual(docB));
// Output: docA == docB true

const colA = firestore.collection("docs");
const colB = firestore.collection("docs");
console.log("colA == colB", colA.isEqual(colB));
// Output: colA == colB true
```

#### Investigation

Looking into the source, I believe the issue is caused by the deep equality check on the `fieldFilters` field [here](https://github.com/googleapis/nodejs-firestore/blob/d5a33f6c88424bb602e0e4ae1128b109b82da38c/dev/src/reference.ts#L1211) in the `QueryOptions` class. Run this script using `ts-node` to verify.

```ts
import { Firestore } from "@google-cloud/firestore";
import { QueryOptions } from "@google-cloud/firestore/build/src/reference";
import deepEqual from "fast-deep-equal";

const queryA = firestore.collection("docs").where("field", ">", 1);
const queryB = firestore.collection("docs").where("field", ">", 1);

const aOptions = (queryA as any)._queryOptions as QueryOptions<unknown>;
const bOptions = (queryB as any)._queryOptions as QueryOptions<unknown>;

console.log("A instanceof QueryOptions", aOptions instanceof QueryOptions); // true
console.log("B instanceof QueryOptions", bOptions instanceof QueryOptions); // true
console.log("parentPath", aOptions.parentPath.isEqual(bOptions.parentPath)); // true
console.log("collectionId", aOptions.collectionId === bOptions.collectionId); // true
console.log("converter", aOptions.converter === bOptions.converter); // true
console.log("allDescendants", aOptions.allDescendants === bOptions.allDescendants); // true
console.log("limit", aOptions.limit === bOptions.limit); // true

console.log("fieldFilters", deepEqual(aOptions.fieldFilters, bOptions.fieldFilters)); // false <-- HERE
console.log("fieldOrders", deepEqual(aOptions.fieldOrders, bOptions.fieldOrders)); // true
console.log("startAt", deepEqual(aOptions.startAt, bOptions.startAt)); // true
console.log("endAt", deepEqual(aOptions.endAt, bOptions.endAt)); // true
console.log("projection", deepEqual(aOptions.projection, bOptions.projection)); // true

console.log("offset", aOptions.offset === bOptions.offset); // true
console.log("kindless", aOptions.kindless === bOptions.kindless); // true
console.log("requireConsistency", aOptions.requireConsistency === bOptions.requireConsistency); // true
```

And digging deeper, it seems that the private property `serializer` ([see here](https://github.com/googleapis/nodejs-firestore/blob/d5a33f6c88424bb602e0e4ae1128b109b82da38c/dev/src/reference.ts#L714)) in the `FieldFilter` class is failing the deep equality check due to the arrow function properties in the [Serializer](https://github.com/googleapis/nodejs-firestore/blob/d5a33f6c88424bb602e0e4ae1128b109b82da38c/dev/src/serializer.ts#L58) class. I'm not sure if this technically means that the two queries are not the same, but it's definitely not expected behaviour based on what's exposed through the public API.

```ts
// ... continued from above

const a = aOptions.fieldFilters[0].serializer;
const b = bOptions.fieldFilters[0].serializer;

console.log(a, b);
console.log("serializer.createReference", deepEqual(a.createReference, b.createReference)); // false <-- HERE
console.log("serializer.createInteger", deepEqual(a.createInteger, b.createInteger)); // false <-- HERE
console.log("serializer.allowUndefined", deepEqual(a.allowUndefined, b.allowUndefined)); // true
```

thebrianchen:
Fixes #1652.

The existing `Query.isEqual()` method incorrectly returns `false` when two different collection objects pointing to the same collection id with the same filters are compared. This was because of the underlying deep-equals implementation checking for arrow functions in the `FieldFilter` class.

This change adds an `isEqual()` implementation for `FieldFilter` that does not check the serializer.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -768,6 +768,14 @@ class FieldFilter {
       },
     };
   }
+
+  isEqual(other: FieldFilter): boolean {
+    return (
+      this.field.isEqual(other.field) &&
+      this.op === other.op &&
+      deepEqual(this.value, other.value)
+    );
+  }
 }
 
 /**
@@ -1197,12 +1205,12 @@ export class QueryOptions<T> {
     return (
       other instanceof QueryOptions &&
       this.parentPath.isEqual(other.parentPath) &&
+      this.fieldFiltersEqual(other.fieldFilters) &&
       this.collectionId === other.collectionId &&
       this.converter === other.converter &&
       this.allDescendants === other.allDescendants &&
       this.limit === other.limit &&
       this.offset === other.offset &&
-      deepEqual(this.fieldFilters, other.fieldFilters) &&
       deepEqual(this.fieldOrders, other.fieldOrders) &&
       deepEqual(this.startAt, other.startAt) &&
       deepEqual(this.endAt, other.endAt) &&
@@ -1210,6 +1218,19 @@ export class QueryOptions<T> {
       this.kindless === other.kindless &&
       this.requireConsistency === other.requireConsistency
     );
+  }
+
+  private fieldFiltersEqual(other: FieldFilter[]): boolean {
+    if (this.fieldFilters.length !== other.length) {
+      return false;
+    }
+
+    for (let i = 0; i < other.length; i++) {
+      if (!this.fieldFilters[i].isEqual(other[i])) {
+        return false;
+      }
+    }
+    return true;
   }
 }
 
--- a/dev/test/query.ts
+++ b/dev/test/query.ts
@@ -417,10 +417,12 @@ describe('query interface', () => {
     );
 
     queryEquals(
-        [queryA.where('a', '==', '1').where('b', '==',2), queryB.where('a', '==', '1',).where('b','==',2)],
-        []
+      [
+        queryA.where('a', '==', '1').where('b', '==', 2),
+        queryB.where('a', '==', '1').where('b', '==', 2),
+      ],
+      []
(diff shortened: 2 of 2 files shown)


schmidt-sebastian (approved):
Thanks!

PR #1664, opened 2022-01-20
fix: split v1 and v1beta1 protos to improve startup time

schmidt-sebastian:


thebrianchen (approved):
lgtm, can you run synthtool or the owl bot to preview that the new version still compiles after this change?

PR #1673, opened 2022-01-27
tests: relax error message check for REST client

schmidt-sebastian:


summer-ji-eng (approved):
Thanks for quickly response.

PR #1685, opened 2022-02-11
feat: #1684 set collection method as parametrized

### LINKED ISSUE #1684: [TypeScript] set function as parametrized
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

When I run the following code 

```
const query = await admin
    .firestore()
    .collection("shops")
    .where("name", "==", context.params.name)
    .get();

  if (!query.empty) {
    const snapshot = query.docs[0];
    const shop = snapshot.data();
    ....
 }
```

My `shop`'s type is `DocumentData` where I would like my collection to returns `Shop` objects.

Being able to set type in collection : 

```
const query = await admin
    .firestore()
    .collection<Shop>("shops")
    .where("name", "==", context.params.name)
    .get();
```

Cast object

```

  type Shope = {
     name: string;
  }

  const shop = snapshot.data() as Shop;
```

I will make a PR for this

mlecoq:

Fixes #1684 🦕

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -725,7 +725,9 @@ export class Firestore implements firestore.Firestore {
    * });
    * ```
    */
-  collection(collectionPath: string): CollectionReference {
+  collection<T extends firestore.DocumentData>(
+    collectionPath: string
+  ): CollectionReference<T> {
     validateResourcePath('collectionPath', collectionPath);
 
     const path = ResourcePath.EMPTY.append(collectionPath);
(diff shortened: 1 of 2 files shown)


schmidt-sebastian:
We cannot take this change as it breaks some invariants as explained in https://github.com/googleapis/nodejs-firestore/issues/1684

PR #1698, opened 2022-03-14
feat: Use REST

schmidt-sebastian:


bcoe:
@schmidt-sebastian this is cool, is your intention to default to gRPC, and document REST?

schmidt-sebastian:
@bcoe I haven't decided yet. We need to fix one more issue with queries and then compare performance data to see what mode should be the default.

alexander-fenster:
@MarkDuckworth It works now! When I `throw new Error('...')` in `node_modules/@grpc/grpc-js/build/src/index.js` and run `npm run system-test:rest`, only a handful of tests that actually require gRPC fail, most tests pass.

dev/src/bulk-writer.ts
```diff
@@ -285,9 +285,10 @@ class BulkCommitBatch extends WriteBatch {
         );
         this.pendingOps[i].onSuccess(new WriteResult(updateTime));
       } else {
-        const error = new (require('google-gax').GoogleError)(
-          status.message || undefined
-        );
+        const error =
+          new (require('google-gax/build/src/fallback').GoogleError)(
```
MarkDuckworth, line 289: I'm wondering if instead of using this path to fallback, we can use a subpath export to the fallback? https://nodejs.org/api/packages.html#subpath-exports

Maybe it's already set up that way, I didn't look at gax source yet. But the path indicates maybe it is not.

MarkDuckworth (approved):
Looks great! One minor suggestion but I don't feel very strongly about it.

dev/src/bulk-writer.ts
alexander-fenster: That's a great idea (and "today I learned" as well).

We will be able to safely start using it when we drop Node 12 support. The documentation says subpath exports were added in v12.7.0 but we have `"engines.node"` set to `">=12"` now, so we cannot possibly start using it without making a semver major. But for the next major I definitely want to do it. I filed https://github.com/googleapis/gax-nodejs/issues/1337 so that we don't forget.

hspak:
Hello! Sorry for reviving an old pull request, but was this data published anywhere?
> We need to fix one more issue with queries and then compare performance data to see what mode should be the default.

MarkDuckworth:
Performance data was not published. From what I have seen, internal tests and customer reports indicate a speedup when using preferRest in the cold start of functions. However, I cannot give specific numbers. 

We did not change the default mode, which is still gRPC.

PR #1741, opened 2022-06-23
feat: Support Logical Termination on RunQueryResponse

cherylEnkidu:
Logical Termination: RunQueryResponse will emit "end" event when the "data" event proto which has "done" field set to true.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2012,6 +2012,7 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
 
     return new Promise((resolve, reject) => {
       let readTime: Timestamp;
+      let hasResolve = false;
 
       this._stream(transactionId)
         .on('error', err => {
@@ -2024,6 +2025,9 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
           }
         })
         .on('end', () => {
+          if (hasResolve) return;
+          hasResolve = true;
+
           if (this._queryOptions.limitType === LimitType.Last) {
             // The results for limitToLast queries need to be flipped since
             // we reversed the ordering constraints before sending the query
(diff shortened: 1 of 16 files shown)


dev/src/reference.ts
```diff
@@ -2024,6 +2025,9 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
           }
         })
         .on('end', () => {
+          if (hasResolve) return;
+          hasResolve = true;
```
schmidt-sebastian, line 2029: hasResolved

Also, did you check whether you can just invoke resolve() again?

dev/test/query.ts
```diff
@@ -719,6 +755,34 @@ describe('query interface', () => {
...
+
+      query
+        .stream()
+        .on('data', doc => {
+          expect(doc).to.be.an.instanceOf(DocumentSnapshot);
+          ++received;
+        })
+        .on('end', () => {
+          expect(received).to.equal(2);
+          expect(++counter).to.equal(1);
```
schmidt-sebastian, line 780: You should test that you only receive a single 'end' callback. One think you can do is change this to:

let endCounter = 0;
query
        .stream()
        .on('data', doc => {
          expect(doc).to.be.an.instanceOf(DocumentSnapshot);
          ++receivedCounter;
        })
        .on('end', () => {
          expect(received).to.equal(2);
         ++endCounter;
          setImmediate(() => {
             expect(endCounter).to.equal(1);
             callback();
          });
        });
    });
  });
```

This way, you ensure that once all test processing is done, there is an additional check that verifies that 'end' is only called once. Your test right now does the same thing, but once you call `callback` test is considered done and it won't fail even if any expectations fail after.

wu-hui (approved):

PR #1766, opened 2022-09-07
fix: Version 7 of protobufjs requires update.sh script to be fixed.

tom-andersen:


alexander-fenster (approved):
LGTM. If possible, please make the PR title fit in 50 characters so that it looks pretty in the release notes :)

PR #1767, opened 2022-09-07
fix: Update with generated code from protobufjs version 7.0.0

tom-andersen:
This is the result of running update.sh script .... I am very confused as to purpose of this script, since it appears to overlap googleapis-gen.

### DIFF
--- a/dev/protos/update.sh
+++ b/dev/protos/update.sh
@@ -38,7 +38,7 @@ PBTS="$(npm bin)/pbts"
 pushd "$WORK_DIR"
 
 # Clone necessary git repos.
-git clone --depth 1 https://github.com/googleapis/googleapis.git
+git clone -b preview --depth 1 https://github.com/googleapis/googleapis.git
 git clone --depth 1 https://github.com/google/protobuf.git
 
 # Copy necessary protos.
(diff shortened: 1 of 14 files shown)


dconeybe (approved):

dconeybe (changes requested):
I'm withdrawing my approval because we don't want stuff from the googleapis preview branch going into the main branch of this repo.

alexander-fenster:
To give you some context, the initial idea of `update.sh` was to split protos into smaller files, and load them separately as needed, which should improve cold start time. This is a unique thing, no other client libraries do this. I haven't looked at why the preview features get included though, the script might need some fixing. Feel free to ping me if you want extra 👀 to look at it.

dev/protos/update.sh
```diff
@@ -38,7 +38,7 @@ PBTS="$(npm bin)/pbts"
 pushd "$WORK_DIR"
 
 # Clone necessary git repos.
-git clone --depth 1 https://github.com/googleapis/googleapis.git
+git clone -b preview --depth 1 https://github.com/googleapis/googleapis.git
```
alexander-fenster, line 41: Why `-b preview`?

alexander-fenster:
In #1765 I updated protos with the latest updates from googleapis. I'm assuming this PR could be closed now.

PR #1772, opened 2022-09-14
fix: Tests will now verify asynchronous termination of underlying steam, and fix related bug.

tom-andersen:


dev/test/query.ts
```diff
@@ -517,7 +517,8 @@ describe('query interface', () => {
       },
     };
 
-    return createInstance(overrides).then(firestore => {
+    return createInstance(overrides).then(firestoreInstance => {
+      firestore = firestoreInstance;
```
cherylEnkidu, line 521: Could you please share with me what's the difference between this change and the old implementation?

tom-andersen: There are the `beforeEach` and `afterEach` functions that respectively assign vanilla firestore instance to `firestore` variable declared outside of `it` test, and then calling `verifyInstance` on `firestore` variable after test has run.

There were 2 problems:

1. The `firestore` variable was never set as part of test when creating a instance with overrides.
2. The `verifyInstance` is an asynchronous function, and the `afterEach` function did not wait for it to complete, therefore any error would be lost.

It is important to assign value to `firestore` variable outside of test in order for cleanup in `afterEach` to verify executors are finished in firestore state.

dev/src/reference.ts
```diff
@@ -2285,7 +2285,9 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
           if (proto.done) {
             logger('Query._stream', tag, 'Trigger Logical Termination.');
             backendStream.unpipe(stream);
-            stream.emit('end');
+            backendStream.resume();
```
ehsannas, line 2288: Why do we need to `resume` and then `end` ? by reading the [documentation](https://sourcegraph.com/npm/types/node@a90b5bfc0c1a1602b1dc1bf6fffdb44a889c1a46/-/blob/stream.d.ts?L305&subtree=true) it seems that `unpipe` followed by `end` should be a reasonable thing to do?

ehsannas (commented):
Generally LGTM. Thanks for the fix.

dev/src/reference.ts
tom-andersen: There is some funky logic that requires the 'end' event to emit.

https://github.com/googleapis/nodejs-firestore/blob/main/dev/src/index.ts#L1623-L1623

For some reason, this doesn't happen unless the backendStream is resumed.

I don't see where in your referenced documentation it shows 'end' doesn't require 'resume'. As far as I can tell, documentation only shows calling 'end' on the downstream, not upstream.

ehsannas: 👍

ehsannas (approved):

PR #1774, opened 2022-09-15
feat: COUNT Queries

tom-andersen:
Add support for "count" queries; that is, a query that gets the number of documents in the result set without actually downloading the documents.


dev/src/reference.ts
```diff
@@ -2832,6 +2854,285 @@ export class CollectionReference<T = firestore.DocumentData>
...
+      transform: (proto: api.IRunAggregationQueryResponse, enc, callback) => {
+        if (proto.result) {
+          const readTime = Timestamp.fromProto(proto.readTime!);
+          const data = this.decodeResult(proto.result);
+          callback(
+            undefined,
+            new AggregateQuerySnapshot<T>(this, readTime, data)
+          );
+        } else {
+          callback(Error('unexpected'));
```
dconeybe, line 2941: Please improve this error message.

dev/src/reference.ts
```diff
@@ -2832,6 +2854,285 @@ export class CollectionReference<T = firestore.DocumentData>
...
+          });
+          backendStream.on('error', err => {
+            backendStream.unpipe(stream);
+            // If a non-transactional query failed, attempt to restart.
+            // Transactional queries are retried via the transaction runner.
+            if (!transactionId && !isPermanentRpcError(err, 'runQuery')) {
+              logger(
+                'Query._stream',
+                tag,
+                'Query failed with retryable stream error:',
```
dconeybe, line 2975: Do you need to put a space after the ':' character? Also on line 2983 below.

dev/src/reference.ts
```diff
@@ -2832,6 +2854,285 @@ export class CollectionReference<T = firestore.DocumentData>
...
+
+  get query(): firestore.AggregateQuery<T> {
+    return this._query;
+  }
+
+  get readTime(): firestore.Timestamp {
+    return this._readTime;
+  }
+
+  getCount(): number {
```
dconeybe, line 3096: Remove getCount(); the users should just get if from data().

dconeybe (changes requested):
I've reviewed the non-test code so far. I'm not done my review, but wanted to send my feedback now because I probably won't get back to it until later today.

dev/src/reference.ts
tom-andersen: This is copied code from the Query class _stream method.

I investigated, and under the covers it uses the `util.format` function that formats correctly.

dev/src/reference.ts
tom-andersen: Removed

types/firestore.d.ts
```diff
@@ -1692,6 +1706,21 @@ declare namespace FirebaseFirestore {
...
+     */
+    count(): AggregateQuery<{count: AggregateField<number>}>;
+
+    /**
+     * Returns `AggregateQuery` for given aggregates based on this `Query`.
+     *
+     * @param aggregates Specify aliases with aggregate functions.
+     * @return An AggregateQuery that contains given aggregates.
+     */
+    aggregate<T extends AggregateSpec>(aggregates: T): AggregateQuery<T>;
```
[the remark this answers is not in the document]
tom-andersen: Removed

types/firestore.d.ts
```diff
@@ -2012,6 +2041,101 @@ declare namespace FirebaseFirestore {
...
+  export type AggregateSpecData<T extends AggregateSpec> = {
+    [P in keyof T]: T[P] extends AggregateField<infer U> ? U : never;
+  };
+
+  export class AggregateQuery<T extends AggregateSpec> {
+    private constructor();
+
+    readonly query: Query<unknown>;
+
+    readonly aggregates: T;
```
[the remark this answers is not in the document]
tom-andersen: Removed

dev/src/reference.ts
```diff
@@ -1615,7 +1615,7 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
    * @param aggregates the aggregations to perform.
    * @return an `AggregateQuery` that performs the given aggregations.
    */
-  aggregate<T extends firestore.AggregateSpec>(
+  _aggregate<T extends firestore.AggregateSpec>(
```
dconeybe, line 1618: Explicitly change the visibility of ` _aggregate` to `private`, using the `private` keyword.

dconeybe (dismissed):
Thank you for addressing my comments, Tom. I just have one minor follow-up comment. I'm going to pass off this review to @ehsannas or @wu-hui to also review the test code.

ehsannas:
Can you fix the CI test failures? (`Property 'getCount' does not exist on type 'AggregateQuerySnapshot<{ count: AggregateField<number>; }>'`)

dev/src/reference.ts
```diff
@@ -2832,6 +2854,273 @@ export class CollectionReference<T = firestore.DocumentData>
...
+        }
+        data[prop] = serializer.decodeValue(fields[prop]);
+      }
+    }
+    return data;
+  }
+
+  /**
+   * Internal method for serializing a query to its RunAggregationQuery proto
+   * representation with an optional transaction id or read time.
```
ehsannas, line 3022: `or read time` ?

dev/system-test/firestore.ts
```diff
@@ -2140,6 +2140,82 @@ describe('Query class', () => {
...
+    await randomCol.doc('doc8').set({foo2: 'bar8'});
+
+    const count1 = randomCol.orderBy('foo2').count();
+    const res1 = await firestore.runTransaction(f => f.get(count1));
+    expect(res1.data().count).to.equal(3);
+
+    const count2 = randomCol.orderBy('foo3').count();
+    const res2 = await firestore.runTransaction(f => f.get(count2));
+    expect(res2.data().count).to.equal(0);
+  });
```
ehsannas, line 2216: A few more tests could be added (e.g. with `startAt` and `endAt`, with `offset`, with collection groups, and in the context of transactions). Java SDK [examples](https://github.com/googleapis/java-firestore/blob/ef39a25263383303a1255a6447048e02d15da0d8/google-cloud-firestore/src/test/java/com/google/cloud/firestore/ITQueryCountTest.java).

dev/src/reference.ts
```diff
@@ -2832,6 +2854,273 @@ export class CollectionReference<T = firestore.DocumentData>
...
 
+export class AggregateField<T> implements firestore.AggregateField<T> {
+  private constructor() {}
+
+  static count(): AggregateField<number> {
+    return new AggregateField<number>();
+  }
+
+  isEqual(other: firestore.AggregateField<T>): boolean {
+    return this === other || other instanceof AggregateField;
```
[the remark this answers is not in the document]
dconeybe: (drive-by comment) I *think* there is still a precedent for doing "unnecessary" instanceof checks in TypeScript to make the code behave well when being used from JavaScript, which doesn't enforce type checking. I know we do this in the web sdk, but not so sure about in the node sdk.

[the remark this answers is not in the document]
tom-andersen: This actually has more to do with inheritance. Notice that we are dealing with `AggregateField<T>` and `firestore.AggregateField<T>`. Any other class that implements the supertype will not be considered equal. When more aggregate types are introduced in the future, there will also be additional comparison of internals that requires we first do `instanceof` to safely access properties.

dev/src/reference.ts
tom-andersen: Copy pasta! Removed.

dev/system-test/firestore.ts
tom-andersen: Will do.

dev/src/reference.ts
tom-andersen: Done

dev/src/reference.ts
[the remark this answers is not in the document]
ehsannas: Makes sense. Thanks

ehsannas (approved):
LGTM. I've added the "do not merge" label. Since the PR exposes the API publicly we should wait until allowed to merge the PR.

dconeybe (approved):

PR #1783, opened 2022-09-30
docs: Write javadocs for COUNT API

dconeybe:
The wording is (mostly) copied from https://github.com/firebase/firebase-android-sdk/pull/4143, and also partially from https://github.com/firebase/firebase-js-sdk/pull/6632.

There were also some last-minute API and implementation changes:
1. `AggregateField.count()` and `AggregateField.isEqual()` were deleted, since they have no use case until other aggregations are supported (the same change was made in the firebase-js-sdk).
1. `Query._aggregate()` was removed, since it was so much simpler to just inline it into `Query.count()`.
1. Some lint warnings were suppressed with `eslint-disable-next-line` comments.
1. `AggregateQuery.isEqual()` and `AggregateQuerySnapshot.isEqual()` were tweaked to accommodate the deletion of `AggregateField.isEqual()`.

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -56,7 +56,6 @@ import {
 import {DocumentWatch, QueryWatch} from './watch';
 import {validateDocumentData, WriteBatch, WriteResult} from './write-batch';
 import api = protos.google.firestore.v1;
-import {Primitive} from '@google-cloud/firestore';
 
 /**
  * The direction of a `Query.orderBy()` clause is specified as 'desc' or 'asc'
@@ -3314,7 +3313,7 @@ function isArrayEqual<T extends {isEqual: (t: T) => boolean}>(
  * @param right Array of primitives.
  * @return True if arrays are equal.
  */
-function isPrimitiveArrayEqual<T extends Primitive>(
+function isPrimitiveArrayEqual<T>(
   left: T[],
   right: T[]
 ): boolean {
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -3061,23 +3061,28 @@ export class AggregateQuery<T extends firestore.AggregateSpec>
     if (this === other) {
       return true;
     }
-    if (other instanceof AggregateQuery) {
-      if (!this._query.isEqual(other._query)) {
-        return false;
-      }
-
-      const thisAggregates: [string, AggregateField<unknown>][] =
-        Object.entries(this._aggregates);
-      const otherAggregates = other._aggregates;
-
-      return (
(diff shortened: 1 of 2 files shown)


dev/src/reference.ts
```diff
@@ -3051,76 +3052,127 @@ export class AggregateQuery<T extends firestore.AggregateSpec>
...
+      return false;
+    }
 
-      return (
-        thisAggregates.length === Object.keys(otherAggregates).length &&
-        thisAggregates.every(([alias, field]) =>
-          field.isEqual(otherAggregates[alias])
-        )
-      );
+    // Verify that this object has the same aggregation keys as `other`.
```
tom-andersen, line 3077: Can we do this without sort? Maybe there exists or we can create a helper method?

The consequence is likely small since we are likely only dealing with maps of one element. So, maybe we can just have a TODO to revisit this when other aggregate types are implement.

https://stackoverflow.com/a/35951373

dev/src/reference.ts
```diff
@@ -3051,76 +3052,127 @@ export class AggregateQuery<T extends firestore.AggregateSpec>
...
-      if (!this._query.isEqual(other._query)) {
-        return false;
-      }
-
-      const thisData = this._data;
-      const thisDataKeys: string[] = Object.keys(thisData);
+    if (!(other instanceof AggregateQuerySnapshot)) {
+      return false;
+    }
+    if (!this.readTime.isEqual(other.readTime)) {
```
tom-andersen, line 3156: The isEquals method for QuerySnapshot does not include timestamp. Are you sure this is correct behavior?

dev/src/reference.ts
```diff
@@ -3051,76 +3052,127 @@ export class AggregateQuery<T extends firestore.AggregateSpec>
...
+      return false;
+    }
+    if (!this.query.isEqual(other.query)) {
+      return false;
+    }
 
-      const otherData = other._data;
-      const otherDataKeys: string[] = Object.keys(otherData);
+    // Verify that this object has the same aggregation keys as `other`.
```
tom-andersen, line 3163: How is this better than previous map comparison?

tom-andersen (changes requested):

dev/src/reference.ts
dconeybe: I looked a little harder (I must have missed this while coding late last night!) and there already exists a `deepEqual()` function that employs the logic you suggested. So I've modified `AggregateQuery.isEqual()` and `AggregateQuerySnapshot.isEqual()` to just use `deepEqual()`.

tom-andersen (approved):

PR #1788, opened 2022-10-14
fix: Force http in the GAX module when using the GAX fallback and connecting to the emulator

MarkDuckworth:
fix: Force http in the GAX module when using the GAX fallback and connecting to the emulator

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -581,7 +581,7 @@ export class Firestore implements firestore.Firestore {
           const sslCreds = grpcModule.credentials.createInsecure();
 
           // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          const settings: any = {
+          const settings: firestore.Settings = {
             sslCreds,
             ...this._settings,
             fallback: useFallback,
@@ -591,7 +591,7 @@ export class Firestore implements firestore.Firestore {
           // also set the `protocol` option for GAX fallback to
           // force http
           if (useFallback) {
-            settings['protocol'] = 'http';
+            settings.protocol = 'http';
           }
 
           client = new module.exports.v1(settings, gax);
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -580,14 +580,21 @@ export class Firestore implements firestore.Firestore {
           const grpcModule = this._settings.grpc ?? require('google-gax').grpc;
           const sslCreds = grpcModule.credentials.createInsecure();
 
-          client = new module.exports.v1(
-            {
-              sslCreds,
-              ...this._settings,
-              fallback: useFallback,
-            },
-            gax
-          );
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const settings: any = {
+            sslCreds,
+            ...this._settings,
+            fallback: useFallback,
+          };
+
+          // If `ssl === false` and we're using the GAX fallback,
+          // also set the `protocol` option for GAX fallback to
+          // force http
+          if (useFallback) {
+            settings['protocol'] = 'http';
+          }
+
+          client = new module.exports.v1(settings, gax);
         } else {
           client = new module.exports.v1(
             {
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -580,16 +580,14 @@ export class Firestore implements firestore.Firestore {
           const grpcModule = this._settings.grpc ?? require('google-gax').grpc;
           const sslCreds = grpcModule.credentials.createInsecure();
 
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
           const settings: ClientOptions = {
             sslCreds,
             ...this._settings,
             fallback: useFallback,
           };
 
-          // If `ssl === false` and we're using the GAX fallback,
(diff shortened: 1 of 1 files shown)


dconeybe:
Can you think about how you might write a unit or integration test for this? My first, probably dumb, idea is to spin up a local http server and try to connect to it and make sure the connection is done via http (and not https). Just something to think about, if you can find a practical way to test this. Maybe there is somewhere else you can "hook in" to make sure the correct protocol is used.

dev/src/index.ts
```diff
@@ -580,14 +580,21 @@ export class Firestore implements firestore.Firestore {
...
-          client = new module.exports.v1(
-            {
-              sslCreds,
-              ...this._settings,
-              fallback: useFallback,
-            },
-            gax
-          );
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          const settings: any = {
```
alexander-fenster, line 584: `settings` probably has type `ClientOptions | firestore.Settings`. The `ClientOptions` type comes from `google-gax`, it can be added to the import on line 19.

dev/src/index.ts
```diff
@@ -580,14 +580,21 @@ export class Firestore implements firestore.Firestore {
...
+            sslCreds,
+            ...this._settings,
+            fallback: useFallback,
+          };
+
+          // If `ssl === false` and we're using the GAX fallback,
+          // also set the `protocol` option for GAX fallback to
+          // force http
+          if (useFallback) {
+            settings['protocol'] = 'http';
```
alexander-fenster, line 594: If `settings` has a proper type, this should be `settings.protocol`.

alexander-fenster (approved):
LGTM but would prefer having a well defined type if possible.

dev/src/index.ts
MarkDuckworth: Thanks for calling this out. ClientOptions is indeed the correct type for this constructor.

dev/src/index.ts
MarkDuckworth: Fixed thanks.

MarkDuckworth:
> Can you think about how you might write a unit or integration test for this? My first, probably dumb, idea is to spin up a local http server and try to connect to it and make sure the connection is done via http (and not https). Just something to think about, if you can find a practical way to test this. Maybe there is somewhere else you can "hook in" to make sure the correct protocol is used.

@dconeybe I think the best solution will be to configure all integration tests to run against the emulator. I'll touch base with the team on this and align with strategy for running against emulator vs production database.

dev/src/index.ts
```diff
@@ -580,14 +580,21 @@ export class Firestore implements firestore.Firestore {
...
 
-          client = new module.exports.v1(
-            {
-              sslCreds,
-              ...this._settings,
-              fallback: useFallback,
-            },
-            gax
-          );
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dconeybe, line 583: This `eslint-disable-next-line` comment can be removed, since you're no longer using `any`.

dev/src/index.ts
```diff
@@ -580,14 +580,21 @@ export class Firestore implements firestore.Firestore {
...
+          const settings: ClientOptions = {
+            sslCreds,
+            ...this._settings,
+            fallback: useFallback,
+          };
+
+          // If `ssl === false` and we're using the GAX fallback,
```
dconeybe, line 590: nit: Can this comment be collapsed into 2 lines (instead of 3) without exceeding the max line length?

dconeybe, line 590: nit: Consider rewording

```
If `ssl === false` and we're using the GAX fallback
```

to

```
Since `ssl === false`, if we're using the GAX fallback
```

The way it is currently worded implied (to me, at least) that the "if" block below should be checking `ssl === false`; however, that check was already done above.

dconeybe (changes requested):

dconeybe:
> > Can you think about how you might write a unit or integration test for this? My first, probably dumb, idea is to spin up a local http server and try to connect to it and make sure the connection is done via http (and not https). Just something to think about, if you can find a practical way to test this. Maybe there is somewhere else you can "hook in" to make sure the correct protocol is used.
> 
> @dconeybe I think the best solution will be to configure all integration tests to run against the emulator. I'll touch base with the team on this and align with strategy for running against emulator vs production database.

But then the scenario where the emulator is _not_ used is left untested. I definitely wouldn't block merging this PR on switching the tests to run against the Firestore emulator. But is there some way that you can write unit tests for the settings that get passed into `new module.exports.v1(settings, gax)`? It looks like it may be somewhat challenging, but it would be good to add such tests to avoid backsliding. Maybe just try it out to see if it's easy and timebox the work. If it ends up being unwieldy, then don't worry about it.

dev/src/index.ts
MarkDuckworth: Yes, thank you.

dev/src/index.ts
MarkDuckworth: Right. Good call on this.

MarkDuckworth: Thanks.

dconeybe (approved):

PR #1807, opened 2022-12-20
patch: ensure that the client pool consistently uses grpc clients after transitioning from rest

MarkDuckworth:
patch: Ensure that the client pool consistently uses gRPC clients after transitioning from REST. Previously there were some edge cases where the client pool may create a REST client after it had transitioned to gRPC. Additionally, API docs were added for preferRest.

BEGIN_COMMIT_OVERRIDE
fix: Ensure that the client pool consistently uses gRPC clients after transitioning from REST
END_COMMIT_OVERRIDE

### DIFF
--- a/dev/test/pool.ts
+++ b/dev/test/pool.ts
@@ -34,6 +34,26 @@ function deferredPromises(count: number): Array<Deferred<void>> {
   return deferred;
 }
 
+function assertOpCount<T>(
+  pool: ClientPool<T>,
+  grpcClientOpCount: number,
+  restClientOpCount: number
+): void {
+  let actualGrpcClientOpCount = 0;
+  let actualRestClientOpCount = 0;
+
+  pool._activeClients.forEach(clientConfig => {
+    if (clientConfig.grpcEnabled) {
+      actualGrpcClientOpCount += clientConfig.activeRequestCount;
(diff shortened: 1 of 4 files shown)


ehsannas (approved):
Thanks for the test coverage!

dev/test/pool.ts
```diff
@@ -156,9 +177,134 @@ describe('Client pool', () => {
...
+
+    // Assert client pool status
+    expect(clientPool.size).to.equal(1);
+    assertOpCount(clientPool, 2, 0);
+
+    operationPromises[1].resolve();
+    operationPromises[2].resolve();
+  });
+
+  it('does not re-use rest instance after beginning the transition to grpc - grpc client full', async () => {
```
ehsannas, line 255: Can we also add another test in which REST client is full and results in a second REST client, followed by a grpc operation?

e.g. 15 rest ops + 5 grpc ops (with limit of 10) --> client 1: 10 rest ops, client 2: 5 rest ops, client 3: 5 grpc ops

MarkDuckworth: Thanks for the suggestion. I added this.

ehsannas: 👍

PR #1812, opened 2023-01-17
fix: Set emulator credentials on rest fallback

### LINKED ISSUE #1811: Unable to connect to Firestore Emulator with preferRest: true
When REST fallback is enabled by setting `preferRest: true`, the fallback client obtains an authenticated `google-auth-library` client to communicate with the Firestore BE services.

If the emulator mode is enabled by setting `FIRESTORE_EMULATOR_HOST` env var, the rest client will look for application default credentials in the environment to initialize credentials to obtain an authenticated `google-auth-library` client. This step fails if there are no credentials set in the environment. Firestore Client should not use service account credentials to connect to the emulator even if the real service credentials are available, instead we should explicitly set emulator credentials on emulator mode.

We already set emulator credentials on gRPC mode by setting custom headers at: https://github.com/googleapis/nodejs-firestore/blob/653a8e996b6e6be01825f030e81d0379b4318c8a/dev/src/index.ts#L1453 but for some reason this is not effective on REST fallback.

### Steps to reproduce:

- Set`export FIRESTORE_EMULATOR_HOST="localhost:8080"`
- Remove any application default credentials from your environment:
  - `unset GOOGLE_APPLICATION_CREDENTIALS`
  - `gcloud auth application-default revoke`

```js
import { Firestore } from '@google-cloud/firestore'

// Create a new client
const firestore = new Firestore({ preferRest: true })

async function quickstart() {
  // Obtain a document reference.
  const document = firestore.doc('posts/intro-to-firestore')

  // Enter new data into the document.
  await document.set({
    title: 'Welcome to Firestore',
    body: 'Hello World',
  })
  console.log('Entered new data into the document')
}
quickstart()
```

Expected:
- A new entry should be added to the Firestore Emulator

Actual:
- Error log:
```
(node:62230) UnhandledPromiseRejectionWarning: Error: Could not load the default credentials. Browse to 
https://cloud.google.com/docs/authentication/getting-started for more information.
    at GoogleAuth.getApplicationDefaultAsync (nodejs-firestore/node_modules/google-auth-
library/build/src/auth/googleauth.js:206:19)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async GoogleAuth.getClient (nodejs-firestore/node_modules/google-auth-library/build/src/auth/googleauth.js:638:17)
    at async GrpcClient.createStub (nodejs-firestore/node_modules/google-gax/build/src/fallback.js:211:35)
Caused by: Error
    at WriteBatch.commit (nodejs-firestore/node_modules/@google-cloud/firestore/build/src/write-batch.js:433:23)
    at DocumentReference.set (nodejs-firestore/node_modules/@google-cloud/firestore/build/src/reference.js:392:27)
    at quickstart (file://nodejs-firestore/node-js-firestore-repro.js:11:18)
    at file://nodejs-firestore/node-js-firestore-repro.js:17:1
    at ModuleJob.run (internal/modules/esm/module_job.js:152:23)
    at async Loader.import (internal/modules/esm/loader.js:166:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

lahirumaramba:
Setting Emulator credentials when REST fallback is enabled by setting `preferRest: true`.

I am not familiar with the codebase and looking for review from the experts :)
We already set emulator credentials on gRPC mode by setting custom headers at: https://github.com/googleapis/nodejs-firestore/blob/653a8e996b6e6be01825f030e81d0379b4318c8a/dev/src/index.ts#L1453 but for some reason this is not effective on REST fallback.

**Test coverage:**
Tried to add unit tests with `sinon.spy` on `modules.exports.v1` and `FirestoreClient`, but couldn't get it to work. Appreciate any thoughts on adding test coverage to this.

Fixes googleapis/gax-nodejs#1811 🦕

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -597,6 +597,23 @@ export class Firestore implements firestore.Firestore {
           // also set the `protocol` option for GAX fallback to force http
           if (useFallback) {
             settings.protocol = 'http';
+
+            // If the emulator mode is enabled, we set emulator credentials
+            // to prevent the gRPC-fallback client from trying to obtain
+            // service account credentials from the environment.
+            if (process.env.FIRESTORE_EMULATOR_HOST) {
+              const emulatorCredentials = {
+                access_token: 'owner',
+              };
+              const emulatorClient =
+                // eslint-disable-next-line node/no-extraneous-require
+                new (require('google-auth-library').OAuth2Client)();
+              emulatorClient.setCredentials(emulatorCredentials);
+              const emulatorAuth = new (require('google-gax').GoogleAuth)({
+                authClient: emulatorClient,
+              });
+              settings.auth = emulatorAuth;
+            }
           }
 
           client = new module.exports.v1(settings, gax);
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -609,6 +609,8 @@ export class Firestore implements firestore.Firestore {
                 // eslint-disable-next-line node/no-extraneous-require
                 new (require('google-auth-library').OAuth2Client)();
               emulatorClient.setCredentials(emulatorCredentials);
+              emulatorClient.projectId =
+                settings.projectId ?? process.env.GCLOUD_PROJECT;
               const emulatorAuth = new (require('google-gax').GoogleAuth)({
                 authClient: emulatorClient,
               });


dev/src/index.ts
```diff
@@ -597,6 +597,23 @@ export class Firestore implements firestore.Firestore {
...
+            // If the emulator mode is enabled, we set emulator credentials
+            // to prevent the gRPC-fallback client from trying to obtain
+            // service account credentials from the environment.
+            if (process.env.FIRESTORE_EMULATOR_HOST) {
+              const emulatorCredentials = {
+                access_token: 'owner',
+              };
+              const emulatorClient =
+                // eslint-disable-next-line node/no-extraneous-require
+                new (require('google-auth-library').OAuth2Client)();
```
lahirumaramba, line 610: Since `google-auth-library` is not a direct dependency (it is a transient dependency of `google-gax`) We get `extraneous-require` lint error here. I think disabling it for this line is okay... WDYT?

MarkDuckworth: If we added 'google-auth-library' to nodejs-firestore dependencies, would that also resolve the issue? Or am I missing some other context?

MarkDuckworth (approved):
LGTM. I am curious about the 'google-auth-library' require.

dev/src/index.ts
lahirumaramba: Yep, adding `google-auth-library` to `package.json` would resolve this, but I am not sure if that is necessary here as `google-gax` should pull in `google-auth-library` as a dependency anyway.

Reference: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md

> If a `require()`'s target is extraneous (it's not written in `package.json`), the program works in local, but will not work after dependencies are re-installed. It will cause troubles to your team/contributors. This rule disallows `require()` of extraneous modules.

MarkDuckworth (changes requested):
I [updated the system tests to run against the emulator](https://github.com/googleapis/nodejs-firestore/pull/1813). Unfortunately, they are failing with this change, but passing without this change. Need to rescind the approval until we can understand the issue.

dev/src/index.ts
```diff
@@ -597,6 +597,23 @@ export class Firestore implements firestore.Firestore {
...
             settings.protocol = 'http';
+
+            // If the emulator mode is enabled, we set emulator credentials
+            // to prevent the gRPC-fallback client from trying to obtain
+            // service account credentials from the environment.
+            if (process.env.FIRESTORE_EMULATOR_HOST) {
+              const emulatorCredentials = {
+                access_token: 'owner',
+              };
+              const emulatorClient =
```
MarkDuckworth, line 608: Root cause for failing tests. The emulatorClient (cached credentials) needs to specify the project ID. If not specified here and not specified in the firestore settings, then the project ID will not be available, causing request the Firestore instance to behave incorrectly.

lahirumaramba: I see! How does the system tests obtain the project id? We can use the firestore settings and `GCLOUD_PROJECT` env var here, if that makes sense.

```js
 emulatorClient.projectId = settings.projectId ?? process.env.GCLOUD_PROJECT;
```
WDYT?

MarkDuckworth:
I'm closing this PR for now to clean things up. We're waiting on a change in the [gax-fallback](https://github.com/googleapis/gax-nodejs/issues/1409), so that there will be consistent implementation of credentials used for the emulator regardless of using rest or grpc. Thanks for creating this @lahirumaramba, we can re-open it later if the other approach fails.

lahirumaramba:
Makes sense! Thanks @MarkDuckworth !

PR #1813, opened 2023-01-17
fix: emulator support for system tests

MarkDuckworth:
fix: emulator support for system tests. Run system tests against the emulator using: `yarn system-test:grpc:emulator` or `yarn system-test:rest:emulator`

BEGIN_COMMIT_OVERRIDE
fix: emulator support for system tests. Run system tests against the emulator using: `yarn system-test:grpc:emulator` or `yarn system-test:rest:emulator`
END_COMMIT_OVERRIDE

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -208,123 +208,127 @@ describe('Firestore class', () => {
   });
 });
 
-describe('CollectionGroup class', () => {
-  const desiredPartitionCount = 3;
-  const documentCount = 2 * 128 + 127; // Minimum partition size is 128.
-
-  let firestore: Firestore;
-  let randomColl: CollectionReference;
-  let collectionGroup: CollectionGroup;
-
-  before(async () => {
-    randomColl = getTestRoot();
-    firestore = randomColl.firestore;
-    collectionGroup = firestore.collectionGroup(randomColl.id);
-
-    const batch = firestore.batch();
-    for (let i = 0; i < documentCount; ++i) {
-      batch.create(randomColl.doc(), {title: 'post', author: 'author'});
-    }
-    await batch.commit();
-  });
-
-  async function getPartitions<T>(
-    collectionGroup: CollectionGroup<T>,
-    desiredPartitionsCount: number
-  ): Promise<QueryPartition<T>[]> {
-    const partitions: QueryPartition<T>[] = [];
-    for await (const partition of collectionGroup.getPartitions(
-      desiredPartitionsCount
-    )) {
-      partitions.push(partition);
-    }
-    return partitions;
-  }
-
-  async function verifyPartitions<T>(
-    partitions: QueryPartition<T>[]
-  ): Promise<QueryDocumentSnapshot<T>[]> {
-    expect(partitions.length).to.not.be.greaterThan(desiredPartitionCount);
-
-    expect(partitions[0].startAt).to.be.undefined;
-    for (let i = 0; i < partitions.length - 1; ++i) {
-      // The cursor value is a single DocumentReference
-      expect(
-        (partitions[i].endBefore![0] as DocumentReference<T>).isEqual(
-          partitions[i + 1].startAt![0] as DocumentReference<T>
-        )
-      ).to.be.true;
-    }
-    expect(partitions[partitions.length - 1].endBefore).to.be.undefined;
-
-    // Validate that we can use the partitions to read the original documents.
-    const documents: QueryDocumentSnapshot<T>[] = [];
-    for (const partition of partitions) {
-      documents.push(...(await partition.toQuery().get()).docs);
-    }
-    expect(documents.length).to.equal(documentCount);
-
-    return documents;
-  }
-
-  it('partition query', async () => {
-    const partitions = await getPartitions(
-      collectionGroup,
-      desiredPartitionCount
-    );
-    await verifyPartitions(partitions);
-  });
-
-  it('partition query with manual cursors', async () => {
-    const partitions = await getPartitions(
-      collectionGroup,
-      desiredPartitionCount
-    );
-
-    const documents: QueryDocumentSnapshot<DocumentData>[] = [];
-    for (const partition of partitions) {
-      let partitionedQuery: Query = collectionGroup;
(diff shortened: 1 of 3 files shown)


MarkDuckworth:
Run system tests against the emulator using:
`yarn system-test:grpc:emulator` or `yarn system-test:rest:emulator`

dev/system-test/firestore.ts
```diff
@@ -2972,91 +3001,100 @@ describe('BulkWriter class', () => {
...
-      ['DocumentReference.create()', randomColl => randomColl.doc().create({})],
-      ['DocumentReference.set()', randomColl => randomColl.doc().set({})],
-      [
-        'DocumentReference.update()',
-        async randomColl => {
-          const update = randomColl.doc().update('foo', 'bar');
+      string,
+      (coll: CollectionReference) => Promise<unknown>,
+      /* skip */ boolean?
+    ]
```
dconeybe, line 3009: Optional: Consider using an `interface` instead of a tuple, to improve readability. It has become a little unwieldy and using a "normal object" instead of a tuple could improve readability.

dev/system-test/firestore.ts
```diff
@@ -208,123 +208,127 @@ describe('Firestore class', () => {
   });
 });
 
-describe('CollectionGroup class', () => {
-  const desiredPartitionCount = 3;
-  const documentCount = 2 * 128 + 127; // Minimum partition size is 128.
+// Skip partition query tests when running against the emulator
```
dconeybe, line 211: As a general comment to this PR, please ensure that each special case for the emulator is documented with an inline comment explaining _why_. As it is currently written, this comment merely repeats what the next line is clearly doing and, therefore, does not add value. As an outside observer, it is not clear to me _why_ this test suite needs to be skipped when running against the emulator and that rationale is worthy of an inline comment. In other places, it's not clear to me _why_ the expected behavior is different on the emulator. Please update/add comments like this.

dconeybe (changes requested):

dconeybe:
> Run system tests against the emulator using: `yarn system-test:grpc:emulator` or `yarn system-test:rest:emulator`

Are you able to add these to the github actions CI?

dev/system-test/firestore.ts
MarkDuckworth: Agreed.

dev/system-test/firestore.ts
```diff
@@ -208,123 +208,127 @@ describe('Firestore class', () => {
...
+(!process.env.FIRESTORE_EMULATOR_HOST ? describe : describe.skip)(
```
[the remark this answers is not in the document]
MarkDuckworth: Removed the double negative.

dev/system-test/firestore.ts
```diff
@@ -2472,51 +2487,61 @@ describe('Transaction class', () => {
...
-        ++attempts;
-        transaction.update(ref, {foo: 'b'});
-      })
-    ).to.eventually.be.rejectedWith('No document to update');
+    const promise = firestore.runTransaction(async transaction => {
+      ++attempts;
+      transaction.update(ref, {foo: 'b'});
+    });
+
+    if (!process.env.FIRESTORE_EMULATOR_HOST) {
```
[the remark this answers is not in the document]
MarkDuckworth: Removed the double negative.

dev/system-test/firestore.ts
MarkDuckworth: I hate to say no to an idea like this, it would improve readability. The impact is very limited, only to this test, and I'm just trying to keep focused on the goal of running tests against the emulator.

MarkDuckworth:
@dconeybe I created a separate task to modify the CI for running system tests against the emulator. See, b/266735404

dev/system-test/firestore.ts
```diff
@@ -711,13 +716,20 @@ describe('DocumentReference class', () => {
...
+      randomCol
+        .doc()
+        .update({foo: 'b'})
+        .catch(err => {
+          // The emulator generates a different error message, do not validate.
+          if (process.env.FIRESTORE_EMULATOR_HOST === undefined) {
+            expect(err.message).to.match(/No document to update/);
+          }
+
+          throw err;
```
dconeybe, line 730: Is this test going to unconditionally pass, since the `catch` block always throws and the test is verifying that the promise is rejected?

Perhaps a more robust way to write this test would be to save the promise to a local variable. Then, if running against the Firestore emulator just verify that the promise is rejected without validating its error message. Otherwise, if running against production, verify that it is rejected with the specified message. This is basically what is being done on lines 2498-2506.

This comment also applies to the test `will fail to delete document with exists: true if doc does not exist` below.

dev/system-test/firestore.ts
```diff
@@ -2540,24 +2571,28 @@ describe('Transaction class', () => {
     expect(snapshot.get('foo')).to.equal(1);
   });
 
-  it('fails read-only with writes', async () => {
-    let attempts = 0;
+  // This is not supported when working with the emulator. Do not run tests against emulator.
```
dconeybe, line 2574: Please define "This".

dconeybe (changes requested):

dev/system-test/firestore.ts
MarkDuckworth: Nice catch. I wasn't thinking through how chai assertions work. Both instances have been updated. PTAL.

dev/system-test/firestore.ts
MarkDuckworth: Updated for clarity and consistency. PTAL.

dconeybe (approved):
LGTM. Thanks for adding emulator support! It was annoying when I was working on COUNT (in the java-firestore sdk) that some tests would fail against the emulator. I just ended up ignoring those failures but it's better to have each such failure explicitly marked as not being supported in the firestore emulator. This will definitely be a benefit to future work on this sdk.

dconeybe:
> Run system tests against the emulator using: `yarn system-test:grpc:emulator` or `yarn system-test:rest:emulator`

Can you add this information to the PR's description? That will make it much more discoverable.

PR #1815, opened 2023-01-25
fix: Update the depth validation used when writing documents

MarkDuckworth:
BEGIN_COMMIT_OVERRIDE
fix: Update the depth validation used when writing documents, so that it matches the validation of the Firestore backend.
END_COMMIT_OVERRIDE

### DIFF
--- a/dev/src/serializer.ts
+++ b/dev/src/serializer.ts
@@ -312,7 +312,7 @@ export function validateUserInput(
   level?: number,
   inArray?: boolean
 ): void {
-  if (path && path.size > MAX_DEPTH) {
+  if (path && path.size - 1 > MAX_DEPTH) {
     throw new Error(
       `${invalidArgumentMessage(
         arg,
--- a/dev/test/serializer.ts
+++ b/dev/test/serializer.ts
@@ -60,7 +60,8 @@ describe('validateUserInput', () => {
                                           uiData: {
                                             // depth 19
                                             choicesFactors: {
-                                              rarely: 1, // depth 20
+                                              // depth 20
+                                              rarely: 1,
                                             },
                                           },
                                         },
@@ -135,7 +136,8 @@ describe('validateUserInput', () => {
                                             choicesFactors: {
                                               // depth 20
                                               rarely: {
-                                                cat: true, // depth 21
+                                                // depth 21
+                                                cat: true,
                                               },
                                             },
                                           },
@@ -166,7 +168,7 @@ describe('validateUserInput', () => {
         allowUndefined: false,
       })
     ).to.throw(
-      'Value for argument "nestedObject" is not a valid Firestore Object. Input object is deeper than 20 levels or contains a cycle.'
+      /Input object is deeper than 20 levels/i
     );
   });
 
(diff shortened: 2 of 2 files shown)


MarkDuckworth:
The behavior in question: is path "foo.bar.baz" 3 levels deep or 2? Admin SDK says it's 3 but DBE says 2.

The answer is not important since the backend already has customer data that is "21" levels deep. This PR modifies the depth validation in this SDK to compute the depth the same way.

dev/test/serializer.ts
```diff
@@ -0,0 +1,249 @@
...
+    };
+
+    expect(() =>
+      validateUserInput('nestedObject', nestedObject, 'Firestore Object', {
+        allowDeletes: 'none',
+        allowTransforms: false,
+        allowUndefined: false,
+      })
+    ).to.throw(
+      'Value for argument "nestedObject" is not a valid Firestore Object. Input object is deeper than 20 levels or contains a cycle.'
```
dconeybe, line 169: Consider using a regular expression here instead of an exact match on the exception's message (also below).

go/breaking-tests#appendix-common-types-of-brittleness-and-suggested-resolutions

dconeybe (approved):
One minor comment, but LGTM otherwise.

dev/test/serializer.ts
MarkDuckworth: Updated these to regular expressions that will be a little more robust. Thanks!

dev/src/serializer.ts
```diff
@@ -312,7 +312,7 @@ export function validateUserInput(
   level?: number,
   inArray?: boolean
 ): void {
-  if (path && path.size > MAX_DEPTH) {
+  if (path && path.size - 1 > MAX_DEPTH) {
```
xwkuang5, line 315: this may be the simplest workaround for the issue but may not be the best fix. 

IMO, L348 is the one that's problematic, it adds the path of a top-level map property to the accumulating path, creating the off-by-one error.  "foo" in `{"foo": {"bar": "baz"}}` have depth 0 but this implementation currently calculates it as depth 1. 

Currently, `path` serves two purposes: 

1. to be used in the fieldPathMessage (as part of error message)
2. to be used for counting the depth

I think we should separate this responsibility out. 

Separately, I think `validateUserInput` may be exposing too wide of an API. Currently, it can be used to validate an array property or validate a Firestore value as a whole. The ambiguity really comes from this. This also lead to incorrect usage, e.g., in https://github.com/googleapis/nodejs-firestore/blob/ce9abfd8cd65d0d2be4dd7f6a56fb679a100296a/dev/src/field-value.ts#L571, the path should include the array index in order for the depth calculation to be correct. 

I don't have a strong opinion on this though. This change is technically relaxing the requirement on the client side. If anything, we always have the backend side as fallback so the change as is may be fine.

dconeybe (approved):

dev/src/serializer.ts
MarkDuckworth: Thanks for the feedback.

Regarding L348 being problematic. If a path of `['foo', 'bar']` is considered to have a depth of `1` and  `['foo']` is considered to have a depth of `0`, then yes we could remove the top level map property or we could compute the depth as `path.size - 1`. It's the same IMO and allows the re-use of path.

I agree with your statement "I think validateUserInput may be exposing too wide of an API", but in consideration of our other discussion of removing all of this validation. I'm going to avoid any refactor for now.

PR #1817, opened 2023-01-26
WIP: Stub out the API for aggregates sum and average.

MarkDuckworth:

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import {
+  AggregateField,
   DocumentData,
   PartialWithFieldValue,
   QuerySnapshot,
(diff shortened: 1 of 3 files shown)


dev/system-test/firestore.ts
```diff
@@ -2265,6 +2266,70 @@ describe('Aggregates', () => {
...
+
+    const aggregateQuery = randomCol.aggregate({
+      barAverage: AggregateField.average('bar'),
+      fooSum: AggregateField.sum('foo'),
+      count: AggregateField.count(),
+    });
+    const res = await aggregateQuery.get();
+
+    expect(res.data().barAverage).to.equal(15);
+    expect(res.data().fooSum).to.equal(3);
```
IchordeDionysos, line 2330: Wouldn't this be 6 or 12 here? @MarkDuckworth 
```suggestion
    expect(res.data().fooSum).to.equal(6);
```

IchordeDionysos: Or where is the 3 coming from?

dev/system-test/firestore.ts
```diff
@@ -2265,6 +2266,70 @@ describe('Aggregates', () => {
...
+  it('runs multiple aggregates over the multiple fields', async () => {
+    await randomCol.doc('doc1').set({foo: 2, bar: 10});
+    await randomCol.doc('doc2').set({foo: 4, bar: 20});
+    await randomCol.doc('doc3').set({foo: 6});
+    await randomCol.doc('doc4').set({notfoo: 8});
+
+    const aggregateQuery = randomCol.aggregate({
+      barAverage: AggregateField.average('bar'),
+      fooSum: AggregateField.sum('foo'),
+      count: AggregateField.count(),
```
IchordeDionysos, line 2325: I wonder if `count` could be used to check how many documents a query would return and how many documents have a particular field.

e.g. 
```ts
const aggregateQuery = randomCol.aggregate({
  count: AggregateField.count('foo'),
});
const res = await aggregateQuery.get();
expect(res.data().count).to.equal(3);
```

If no field is given all documents would be counted.

IchordeDionysos: And I bet it may be technically more complex, but I think I'd expect that in this case here:

`fooSum` would still be 12, while `barAverage` be 15, as all documents containing the field `bar` average to 15 and all documents containing the field `foo` sum to 12.

And for count maybe I'd expect 3 here as only 3 documents have either the field `foo` or `bar`.

IchordeDionysos (commented):
Just a few comments regarding my personal expectations ☺️ 
Hope they are helpful

dev/system-test/firestore.ts
MarkDuckworth: Agreed. Looks like this should be `6`. Maybe I was thinking average when I did the math.

PR #1831, opened 2023-03-14
feat: Sum and Average Aggregations.

ehsannas:


dev/system-test/firestore.ts
```diff
@@ -2284,6 +2285,1554 @@ describe('Aggregates', () => {
...
+      await addTestDocs(testDocs);
+      const snapshot = await col
+        .aggregate({
+          totalRating: AggregateField.sum('rating'),
+        })
+        .get();
+      expect(snapshot.data().totalRating).to.equal(13);
+    });
+
+    it('performs sum that overflows max int', async () => {
```
MarkDuckworth, line 3063: I have to fix this test in Web. There will be another PR that will have to be ported for node. I can do this.

MarkDuckworth (commented):
Looking good. Just a few comments. I can probably finish this PR if you are too busy.

dev/src/aggregate.ts
```diff
@@ -29,7 +30,7 @@ export class Aggregate {
   constructor(
     readonly alias: AggregateAlias,
     readonly aggregateType: firestore.AggregateType,
-    readonly fieldPath?: string
+    readonly fieldPath?: string | firestore.FieldPath
```
MarkDuckworth, line 33: FWIW. This argument is only type `FieldPath` in Web. Forcing the API layer to convert string to `FieldPath` earlier. This could be more consistent, but it's not a big deal.

MarkDuckworth: That would simplify other code consuming the field path that is calling `FieldPath.fromArgument(this.fieldPath!)`

MarkDuckworth (approved):

PR #1848, opened 2023-05-10
feat: Add support for environment variable FIRESTORE_PREFER_REST

MarkDuckworth:

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -537,8 +537,10 @@ export class Firestore implements firestore.Firestore {
    * `DocumentReference<T>.onSnapshot()`, `CollectionReference<T>.onSnapshot()`, or
    * `Query<T>.onSnapshot()`. If specified, this setting value will take precedent over the
    * environment variable `FIRESTORE_PREFER_REST`. If not specified, the
-   * SDK will use the value specified in environment variable `FIRESTORE_PREFER_REST=true|false`. And if the
-   * environment variable is also not set, then the setting will default to `false`.
+   * SDK will use the value specified in the environment variable `FIRESTORE_PREFER_REST`.
+   * Valid values of `FIRESTORE_PREFER_REST` are `true` ('1') or `false` (`0`). Values are
+   * not case-sensitive. Any other value for the environment variable will be ignored and
+   * a warning will be logged to the console.
    */
   constructor(settings?: firestore.Settings) {
     const libraryHeader = {
@@ -669,9 +671,8 @@ export class Firestore implements firestore.Firestore {
 
     // If preferRest is not specified in settings, but is set as environment variable,
     // then use the environment variable value.
-    const [preferRestEnvIsSet, preferRestEnvValue] =
-      tryGetPreferRestEnvironmentVariable();
-    if (settings.preferRest === undefined && preferRestEnvIsSet) {
+    const preferRestEnvValue = tryGetPreferRestEnvironmentVariable();
+    if (settings.preferRest === undefined && preferRestEnvValue !== undefined) {
       settings = {
         ...settings,
         preferRest: preferRestEnvValue,
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -60,6 +60,7 @@ import {
   isPermanentRpcError,
   requestTag,
   wrapError,
+  tryGetPreferRestEnvironmentVariable,
 } from './util';
 import {
   validateBoolean,
@@ -534,7 +535,10 @@ export class Firestore implements firestore.Firestore {
    * for communication from that point forward. Currently the only operation
    * that requires gRPC is creating a snapshot listener with the method
    * `DocumentReference<T>.onSnapshot()`, `CollectionReference<T>.onSnapshot()`, or
-   * `Query<T>.onSnapshot()`.
+   * `Query<T>.onSnapshot()`. If specified, this setting value will take precedent over the
+   * environment variable `FIRESTORE_PREFER_REST`. If not specified, the
+   * SDK will use the value specified in environment variable `FIRESTORE_PREFER_REST=true|false`. And if the
+   * environment variable is also not set, then the setting will default to `false`.
    */
   constructor(settings?: firestore.Settings) {
     const libraryHeader = {
@@ -662,6 +666,17 @@ export class Firestore implements firestore.Firestore {
     }
 
     let url: URL | null = null;
+
+    // If preferRest is not specified in settings, but is set as environment variable,
+    // then use the environment variable value.
+    const [preferRestEnvIsSet, preferRestEnvValue] =
+      tryGetPreferRestEnvironmentVariable();
+    if (settings.preferRest === undefined && preferRestEnvIsSet) {
+      settings = {
+        ...settings,
+        preferRest: preferRestEnvValue,
+      };
+    }
 
     // If the environment variable is set, it should always take precedence
     // over any user passed in settings.
--- a/dev/src/util.ts
+++ b/dev/src/util.ts
@@ -219,27 +219,28 @@ export function wrapError(err: Error, stack: string): Error {
 }
 
 /**
- * Returns a tuple indicating if FIRESTORE_PREFER_REST is set to a valid value,
- * and if so, whether the value enables preferRest or disables preferRest.
+ * Parses the value of the environment variable FIRESTORE_PREFER_REST, and returns
+ * a value indicating if the environment variable enables or disables preferRest.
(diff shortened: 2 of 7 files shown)


dev/test/util.ts
```diff
@@ -33,4 +34,80 @@ describe('isPlainObject()', () => {
...
+    });
+
+    it('ignores case', async () => {
+      process.env.FIRESTORE_PREFER_REST = 'True';
+      const [prevSet, prevValue] = tryGetPreferRestEnvironmentVariable();
+      expect(prevSet).to.be.true;
+      expect(prevValue).to.be.true;
+    });
+
+    it('indicates when the environment variable is not set', async () => {
```
dconeybe, line 94: "indicates" maybe isn't the right word here. I interpreted it to mean that the test is verifying that there is some "indication" that the environment variable is not set; however, there is no such "indication".

dev/test/util.ts
```diff
@@ -33,4 +34,80 @@ describe('isPlainObject()', () => {
...
+    });
+
+    it('reads 0', async () => {
+      process.env.FIRESTORE_PREFER_REST = '0';
+      const [prevSet, prevValue] = tryGetPreferRestEnvironmentVariable();
+      expect(prevSet).to.be.true;
+      expect(prevValue).to.be.false;
+    });
+
+    it('ignores case', async () => {
```
dconeybe, line 87: Add a test for "strips leading and trailing whitespace"

dconeybe (changes requested):
Overall looks fantastic! Great test coverage and well designed. Thanks for making this improvement.

dev/src/index.ts
```diff
@@ -534,7 +535,10 @@ export class Firestore implements firestore.Firestore {
    * for communication from that point forward. Currently the only operation
    * that requires gRPC is creating a snapshot listener with the method
    * `DocumentReference<T>.onSnapshot()`, `CollectionReference<T>.onSnapshot()`, or
-   * `Query<T>.onSnapshot()`.
+   * `Query<T>.onSnapshot()`. If specified, this setting value will take precedent over the
+   * environment variable `FIRESTORE_PREFER_REST`. If not specified, the
+   * SDK will use the value specified in environment variable `FIRESTORE_PREFER_REST=true|false`. And if the
```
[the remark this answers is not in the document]
MarkDuckworth: IMO this isn't a great place to go into detail about the behavior of the environment variable, but after looking around, I don't think we have a better place to document environment variables used by the SDK. PTAL at the changes here and let me know what you think.

dev/src/util.ts
```diff
@@ -217,3 +217,29 @@ export function wrapError(err: Error, stack: string): Error {
...
+
+/**
+ * Returns a tuple indicating if FIRESTORE_PREFER_REST is set to a valid value,
+ * and if so, whether the value enables preferRest or disables preferRest.
+ *
+ * @return A tuple where the first value indicates if the environment variable is set to a valid value,
+ * and the second value indicates if the environment variable enables preferRest or disables preferRest.
+ */
+export function tryGetPreferRestEnvironmentVariable():
+  | [false, undefined]
```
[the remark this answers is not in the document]
MarkDuckworth: Hmm. This design is an artifact of an older implementation. It no longer makes sense. Thanks for calling it out.

dev/test/index.ts
```diff
@@ -639,6 +639,58 @@ describe('instantiation', () => {
...
+        delete process.env.FIRESTORE_PREFER_REST;
+      } else {
+        process.env.FIRESTORE_PREFER_REST = originalValue;
+      }
+    });
+
+    it('preferRest is disabled (falsy) by default', async () => {
+      delete process.env.FIRESTORE_PREFER_REST;
+      const firestore = new Firestore.Firestore({});
+      expect(!!firestore['_settings'].preferRest).to.be.false;
```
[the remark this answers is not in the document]
MarkDuckworth: Added a comment to clarify. Curious what you think about alternatively initializing this setting value to `false` (instead of leaving it as undefined) in the case where the setting is not set by the consumer?

dev/test/util.ts
MarkDuckworth: added. thanks.

dev/test/util.ts
MarkDuckworth: let me know if the new test name clarifies it for you. thanks

dev/src/util.ts
```diff
@@ -219,27 +219,28 @@ export function wrapError(err: Error, stack: string): Error {
...
 
 /**
- * Returns a tuple indicating if FIRESTORE_PREFER_REST is set to a valid value,
- * and if so, whether the value enables preferRest or disables preferRest.
+ * Parses the value of the environment variable FIRESTORE_PREFER_REST, and returns
+ * a value indicating if the environment variable enables or disables preferRest.
  *
- * @return A tuple where the first value indicates if the environment variable is set to a valid value,
- * and the second value indicates if the environment variable enables preferRest or disables preferRest.
+ * This method will warn to the console if the environment variable is set to an unsupported value.
```
dconeybe, line 225: ultranit: "method" -> "function" (a "method" is a member of a class, which this function is not).

dev/src/util.ts
```diff
@@ -219,27 +219,28 @@ export function wrapError(err: Error, stack: string): Error {
...
+ *
+ * @return {boolean | undefined} `true` if the environment variable enables `preferRest`,
```
dconeybe, line 227: nit: I'd recommend to remove the definition of the return type from the documentation since it's already specified formally in the function's return type declaration. By repeating it here you risk it getting out of sync with the actual return type, which is confusing.

dev/src/util.ts
```diff
@@ -219,27 +219,28 @@ export function wrapError(err: Error, stack: string): Error {
...
+ * `false` if the environment variable disable `preferRest`, or `undefined` if the
```
dconeybe, line 228: ultranit: "disable" -> "disables"

dev/test/index.ts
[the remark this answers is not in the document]
dconeybe: IMO leaving it as `undefined` is preferred because it more clearly conveys that the setting is optional. Thanks for adding the comment; that is helpful (to me!).

dev/test/util.ts
```diff
@@ -58,51 +58,50 @@ describe('isPlainObject()', () => {
 
     it('reads true', async () => {
       process.env.FIRESTORE_PREFER_REST = 'true';
-      const [prevSet, prevValue] = tryGetPreferRestEnvironmentVariable();
-      expect(prevSet).to.be.true;
+      const prevValue = tryGetPreferRestEnvironmentVariable();
```
dconeybe, line 61: nit: Can you rename `prevValue` to avoid the abbreviation? I don't know what "prev" is an abbreviation for. Is it "previous"? Oh, is it an acronym for "prefer rest environment variable"? Consider a name like "returnValue", "result" or "actual", or even just "value". Even better, get rid of the local variable and merge it into the next line, like this:

```ts
expect(tryGetPreferRestEnvironmentVariable()).to.be.true;
```

I actually find that to be the most readable because it "reads like a story" from left to right.

In general, non-standard abbreviations detract from readability. Avoid them.

Applies here and elsewhere in this file.

dev/test/util.ts
dconeybe: It is perfect. Thank you.

dconeybe (changes requested):

dev/src/util.ts
MarkDuckworth: Good way to look at it.

dev/src/util.ts
```diff
@@ -222,10 +222,10 @@ export function wrapError(err: Error, stack: string): Error {
  * Parses the value of the environment variable FIRESTORE_PREFER_REST, and returns
  * a value indicating if the environment variable enables or disables preferRest.
  *
- * This method will warn to the console if the environment variable is set to an unsupported value.
+ * This function will warn to the console if the environment variable is set to an unsupported value.
```
dconeybe, line 225: nit: wrap at 80 characters

dconeybe (changes requested):

dev/src/util.ts
MarkDuckworth: FWIW, (gts)[https://github.com/google/gts] enforces Google's typescript style guide and does not enforce that column limit on comments and some other line types. It's unclear if that's because there's an issue or the 80-column limit does not actually apply.

dconeybe (approved):

PR #1853, opened 2023-05-30
fix: Updated logging in the client pool and client factory to log information about the required transport and actual transport used.

MarkDuckworth:


dev/src/index.ts
```diff
@@ -617,7 +617,12 @@ export class Firestore implements firestore.Firestore {
           );
         }
 
-        logger('Firestore', null, 'Initialized Firestore GAPIC Client');
+        logger(
+          'clientFactory',
+          null,
+          'Initialized Firestore GAPIC Client (useFallback: %s)',
+          useFallback
+        );
```
ehsannas, line 625: not part of this PR, but the 575 above

```js
const useFallback = !this._settings.preferRest || requiresGrpc ? false : 'rest';
```
is... not great. Took me a second to wrap my head around it, partly because the logic is negated IMO. I think writing it this way makes more sense:

```js
const useFallback = this._settings.preferRest && !requiresGrpc ? 'rest' : false;
```

ehsannas (approved):

ehsannas:
Also, conventionalcommits bot is failing. Need to update the PR title.

PR #1869, opened 2023-07-29
Unhide APIs and enable tests

ehsannas:
All tests pass.

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -2421,8 +2421,7 @@ describe('count queries using aggregate api', () => {
   }
 });
 
-// TODO (sum/avg) enable these tests when sum/avg is supported by the backend
-describe.skip('Aggregation queries', () => {
+describe('Aggregation queries', () => {
   let firestore: Firestore;
   let col: CollectionReference;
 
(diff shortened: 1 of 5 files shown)


ehsannas:
@MarkDuckworth this is merging into `ehsann/sum-avg` and since it enables tests against production, the CI is expected to fail. My intention is to merge this PR and will later have a PR from `ehsann/sum-avg` to `main` after protos are published.

types/firestore.d.ts
```diff
@@ -2090,7 +2089,6 @@ declare namespace FirebaseFirestore {
    */
   export type AggregateType = 'count' | 'avg' | 'sum';
```
MarkDuckworth, line 2090: I believe we also need to remove `@internal` from this type so that the type is exposed in documentation.

dev/system-test/firestore.ts
```diff
@@ -3844,6 +3625,236 @@ describe.skip('Aggregation queries', () => {
...
+          })
+          .get();
+        expect(snapshot.data().totalRating).to.equal(0);
+        expect(snapshot.data().averageRating).to.be.null;
+        expect(snapshot.data().totalPages).to.equal(200);
+        expect(snapshot.data().averagePages).to.equal(100);
+        expect(snapshot.data().countOfDocs).to.equal(2);
+      });
+
+      it('performs aggregations on nested map values', async () => {
```
MarkDuckworth, line 3825: This test can be modified to run against production by dropping the aggregations over the second nested field `'metadata.rating.critic'`. For your consideration.

dev/system-test/firestore.ts
```diff
@@ -3844,6 +3625,236 @@ describe.skip('Aggregation queries', () => {
...
+            totalPages: AggregateField.sum('pages'),
+            averageYear: AggregateField.average('year'),
+          })
+          .get();
+        expect(snapshot.data().totalRating).to.be.NaN;
+        expect(snapshot.data().totalPages).to.equal(300);
+        expect(snapshot.data().averageYear).to.equal(2000);
+      });
+
+      it('performs aggregates when using `in` operator', async () => {
```
MarkDuckworth, line 3727: This test can be modified to run against production by dropping the aggregations over fields other than `'rating'`. For your consideration.

MarkDuckworth (approved):
Looks like one additional member that needs to be made public, but otherwise it's good to go.

dev/system-test/firestore.ts
ehsannas: Done

dev/system-test/firestore.ts
ehsannas: Done

ehsannas:
The CI is expected to fail.

PR #1878, opened 2023-08-08
fix: Add tests for multiple inequality support

milaGGL:
Port integration tests from web SDK to make sure client and server sdks behaviours are consistent.
Update the `createImplicitOrderBy` function for queries with cursor. 

Updated the isInequality check to be consistent with mobile&web sdks:
https://github.com/firebase/firebase-js-sdk/blob/320e58cc57d0fc485548611c5975dd3df69f8d18/packages/firestore/src/core/filter.ts#L177
https://github.com/firebase/firebase-ios-sdk/blob/2eccf796524777ce8fae5b82c3e4b0d9875bb9af/Firestore/core/src/core/field_filter.cc#L146
https://github.com/firebase/firebase-android-sdk/blob/b7a34fb108f1d619f5cc98677582e585c239048a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/FieldFilter.java#L145

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -1804,6 +1804,11 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
     );
   }
 
+  /**
+   * Returns the sorted array of unique inequality filter fields used in this query.
+   *
+   * @return An array of inequality filter fields sorted lexicographically by FieldPath.
+   */
   private getInequalityFilterFields(): FieldPath[] {
     let FieldPathSet = new Set<FieldPath>();
 
@@ -1815,7 +1820,6 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
       }
     }
 
-    // Sort the inequality fields lexicographically.
     return [...FieldPathSet].sort((a, b) => a.compareTo(b));
   }
 
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -710,9 +710,6 @@ abstract class FilterInternal {
   /** Returns a list of all filters that are contained within this filter */
   abstract getFilters(): FilterInternal[];
 
-  /** Returns the field of the first filter that's an inequality, or null if none. */
-  abstract getFirstInequalityField(): FieldPath | null;
-
   /** Returns the proto representation of this filter */
   abstract toProto(): Filter;
 
@@ -733,13 +730,6 @@ class CompositeFilterInternal extends FilterInternal {
 
   public getFilters(): FilterInternal[] {
     return this.filters;
-  }
-
-  public getFirstInequalityField(): FieldPath | null {
-    return (
-      this.getFlattenedFilters().find(filter => filter.isInequalityFilter())
-        ?.field ?? null
-    );
   }
 
   public isConjunction(): boolean {
@@ -805,14 +795,6 @@ class FieldFilterInternal extends FilterInternal {
 
   public getFilters(): FilterInternal[] {
     return [this];
-  }
-
-  getFirstInequalityField(): FieldPath | null {
-    if (this.isInequalityFilter()) {
-      return this.field;
-    } else {
-      return null;
-    }
   }
 
   /**
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -1805,22 +1805,22 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
   }
 
   /**
-   * Returns the sorted array of unique inequality filter fields used in this query.
+   * Returns the sorted array of inequality filter fields used in this query.
    *
    * @return An array of inequality filter fields sorted lexicographically by FieldPath.
    */
   private getInequalityFilterFields(): FieldPath[] {
-    let FieldPathSet = new Set<FieldPath>();
+    const inequalityFields: FieldPath[] = [];
 
     for (const filter of this._queryOptions.filters) {
       for (const subFilter of filter.getFlattenedFilters()) {
         if (subFilter.isInequalityFilter()) {
-          FieldPathSet = FieldPathSet.add(subFilter.field);
+          inequalityFields.push(subFilter.field);
         }
       }
     }
 
-    return [...FieldPathSet].sort((a, b) => a.compareTo(b));
+    return inequalityFields.sort((a, b) => a.compareTo(b));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


dev/src/reference.ts
```diff
@@ -843,6 +843,7 @@ class FieldFilterInternal extends FilterInternal {
       case 'GREATER_THAN_OR_EQUAL':
       case 'LESS_THAN':
       case 'LESS_THAN_OR_EQUAL':
+      case 'NOT_EQUAL':
```
milaGGL, line 846: wonder why we didn't include "!=" in inequality before.

dev/src/reference.ts
```diff
@@ -1846,29 +1862,34 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
     }
```
milaGGL, line 1862: `createImplicitOrderBy` function accepts `DocumentReference` as value for cursor as well, but the server returned result for it is not correct. The cursor parameter type is documented here, it might need further digging.  https://docs.google.com/document/d/1VOJfT4QBOmS8AW8XeWJcpaUdh0dW_RCgs-n9HYTNxTw/edit?resourcekey=0-l27mCN4xSkZLIZx9olzbfw#heading=h.wiam1kg6dphx

dev/src/reference.ts
```diff
@@ -1821,6 +1822,21 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
     );
   }
 
+  private getInequalityFilterFields(): FieldPath[] {
+    let FieldPathSet = new Set<FieldPath>();
```
milaGGL, line 1826: no SortedSet. so save fields into set, then sort them into an array.

dev/system-test/firestore.ts
```diff
@@ -2544,6 +2544,403 @@ describe('Query class', () => {
...
+
+      results = await collection
+        .where('field', '>=', 'field 100')
+        .where(new FieldPath('field.dot'), '!=', 300)
+        .where('field\\slash', '<', 400)
+        .orderBy('name', 'desc')
+        .get();
+      expectDocs(results, 'doc2', 'doc3');
+
+      // Cursor will add implicit order by fields in the same order as server does.
```
milaGGL, line 2707: extra cursor test

dev/src/reference.ts
milaGGL: in the main branch, `col.where(a != 5).startAt(7);` will be validated with error message: `Error: Too many cursor values specified. The specified values must match the orderBy() constraints of the query.`

if we do it with document reference, which bypasses the validation, BE will return error: `order by clause cannot contain more fields after the key`. While if it is not "!=" but other inequality operators, BE will return empty result.

dev/src/reference.ts
```diff
@@ -1821,6 +1804,25 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
     );
   }
 
+  /**
+   * Returns the sorted array of inequality filter fields used in this query.
```
ehsannas, line 1808: We should use consistent implementation when porting between platforms, unless there's a compelling reason not to. IIUC other platforms return a Set with unique inequality filter fields, but this implementation returns an array with potentially duplicate fields. Is there a reason?

We should either use set in all platforms, or an array with potential duplicates on all platforms.

dev/system-test/firestore.ts
```diff
@@ -2544,6 +2544,444 @@ describe('Query class', () => {
...
+      });
+
+      // ordered by: name asc, metadata.createdAt asc, __name__  asc
+      let query = collection
+        .where('metadata.createdAt', '<=', 500)
+        .where('metadata.createdAt', '>', 100)
+        .where('name', '!=', 'room 200')
+        .orderBy('name');
+      let docSnap = await collection.doc('doc4').get();
+      let query_sdk_normalized = query.startAt(docSnap);
```
ehsannas, line 2700: Oops. Shouldn't mix the use of snake case (e.g. `query_sdk_normalized`) and camel case (e.g. `docSnap`).

dev/src/reference.ts
ehsannas: If I understand correctly, prior to adding `case 'NOT_EQUAL'`, the user would _always_ get an error when using cursors (either from sdk or server). This fixes the issue.

dev/system-test/firestore.ts
ehsannas: also, we can probably use a better name than "sdk normalized" . such as: queryWithCursor.

The comment above (line 2669) makes it clear that we're adding cursors to trigger the implicit orderBy logic.

dev/src/reference.ts
milaGGL: In web SDK, we have custom class `SortedSet` which allows us to use custom comparator, ie, the `FieldPath.compareTo()`. But in node sdk, we do not have that `SortedSet`. And `Set` in javascript replies on its own build in equality comparison, which is based on object identity. 

In another word, is we use Set<FieldPath>, `getInequalityFilterFields` on this query will end up with a set of size 2.
`collection
        .where('a', '<', 5)
        .where('a', '>', 1)
`

Should I introduce SortedSet to node SDK? 🤔  It is useful in some cases and not that big: https://github.com/firebase/firebase-js-sdk/blob/78d2738c246555556cba8dcfe2932639f80523ea/packages/firestore/src/util/sorted_set.ts#L27

dev/src/reference.ts
milaGGL: yes, unless the user was SUPER cautious to always add `orderBy`, which makes the query no longer relies on `createImplicitOrderBy` and `isInequalityFilter()` .  Like this:
```
col.where(a != 5).orderby('a').startAt(7);
col.where(a != 5).orderby('a').startAt(docSnap);
col.where(a != 5).orderby('a').startAt(docRef); / /works but wrong value returned

```

dev/src/reference.ts
ehsannas: I didn't realize this SDK doesn't have `SortedSet`. I think it's probably overkill to copy that class into this SDK just for this purpose given that this is the only usage and the caller here has to use a Set to avoid duplicates anyway.

No action needed.

ehsannas (changes requested):
Please port [this](https://www.google.com/url?sa=j&url=https%3A%2F%2Fgithub.com%2Ffirebase%2Ffirebase-android-sdk%2Fpull%2F5194%2Ffiles%23diff-12c7208f39ff0c6321070d9bc38fd0b2b98b65e00f25045b10bc965ef12b9261R619&uct=1677010923&usg=KCWk4nXYTiZODIhTuADlnqBGdUk.&opi=82159412&source=chat) unit test to this PR.

ehsannas (approved):

PR #1887, opened 2023-08-23
feat!: Fix the UpdateData incorrect parameter type issue

tom-andersen:
Port https://github.com/firebase/firebase-js-sdk/pull/7310 to NodeJS

### DIFF
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -576,9 +576,9 @@ export class BulkWriter {
    * });
    * ```
    */
-  create<T>(
-    documentRef: firestore.DocumentReference<T>,
-    data: firestore.WithFieldValue<T>
+  create<AppModelType, DbModelType extends firestore.DocumentData>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
+    data: firestore.WithFieldValue<AppModelType>
   ): Promise<WriteResult> {
     this._verifyNotClosed();
     return this._enqueue(documentRef, 'create', bulkCommitBatch =>
@@ -616,8 +616,8 @@ export class BulkWriter {
    * });
    * ```
    */
-  delete<T>(
-    documentRef: firestore.DocumentReference<T>,
+  delete<AppModelType, DbModelType extends firestore.DocumentData>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
     precondition?: firestore.Precondition
   ): Promise<WriteResult> {
     this._verifyNotClosed();
@@ -626,14 +626,14 @@ export class BulkWriter {
     );
   }
 
-  set<T>(
-    documentRef: firestore.DocumentReference<T>,
-    data: Partial<T>,
+  set<AppModelType, DbModelType extends firestore.DocumentData>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
+    data: Partial<AppModelType>,
     options: firestore.SetOptions
   ): Promise<WriteResult>;
-  set<T>(
-    documentRef: firestore.DocumentReference<T>,
-    data: T
+  set<AppModelType, DbModelType extends firestore.DocumentData>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
+    data: AppModelType
   ): Promise<WriteResult>;
   /**
    * Write to the document referred to by the provided
@@ -675,9 +675,9 @@ export class BulkWriter {
    * });
    * ```
    */
-  set<T>(
-    documentRef: firestore.DocumentReference<T>,
-    data: firestore.PartialWithFieldValue<T>,
+  set<AppModelType, DbModelType extends firestore.DocumentData>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
+    data: firestore.PartialWithFieldValue<AppModelType>,
     options?: firestore.SetOptions
   ): Promise<WriteResult> {
     this._verifyNotClosed();
@@ -687,7 +687,7 @@ export class BulkWriter {
       } else {
         return bulkCommitBatch.set(
           documentRef,
-          data as firestore.WithFieldValue<T>
+          data as firestore.WithFieldValue<AppModelType>
         );
       }
     });
@@ -737,7 +737,7 @@ export class BulkWriter {
    * });
    * ```
    */
-  update<AppModelType, DbModelType>(
+  update<AppModelType, DbModelType extends firestore.DocumentData>(
     documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
     dataOrField: firestore.UpdateData<DbModelType> | string | FieldPath,
     ...preconditionOrValues: Array<
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -336,7 +336,7 @@ export class BulkWriterError extends Error {
     readonly message: string,
 
     /** The document reference the operation was performed on. */
-    readonly documentRef: firestore.DocumentReference<unknown>,
+    readonly documentRef: firestore.DocumentReference<any, any>,
 
     /** The type of operation performed. */
     readonly operationType: 'create' | 'set' | 'update' | 'delete',
@@ -782,7 +782,7 @@ export class BulkWriter {
    */
   onWriteResult(
     successCallback: (
-      documentRef: firestore.DocumentReference<unknown>,
+      documentRef: firestore.DocumentReference<any, any>,
       result: WriteResult
     ) => void
   ): void {
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -737,9 +737,9 @@ export class BulkWriter {
    * });
    * ```
    */
-  update<T>(
-    documentRef: firestore.DocumentReference<T>,
-    dataOrField: firestore.UpdateData<T> | string | FieldPath,
+  update<AppModelType, DbModelType>(
+    documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
+    dataOrField: firestore.UpdateData<DbModelType> | string | FieldPath,
     ...preconditionOrValues: Array<
       {lastUpdateTime?: Timestamp} | unknown | string | FieldPath
     >
--- a/dev/src/bulk-writer.ts
+++ b/dev/src/bulk-writer.ts
@@ -28,7 +28,6 @@ import {
   MAX_RETRY_ATTEMPTS,
 } from './backoff';
 import {RateLimiter} from './rate-limiter';
-import {DocumentReference} from './reference';
 import {Timestamp} from './timestamp';
 import {
   Deferred,
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -1286,25 +1286,20 @@ export class Firestore implements firestore.Firestore {
    * });
    * ```
    */
-  getAll<
-    AppModelType,
-    DbModelType extends firestore.DocumentData = firestore.DocumentData
-  >(
+  getAll(
     ...documentRefsOrReadOptions: Array<
-      | firestore.DocumentReference<AppModelType, DbModelType>
-      | firestore.ReadOptions
+      firestore.DocumentReference<any, any> | firestore.ReadOptions
     >
-  ): Promise<Array<DocumentSnapshot<AppModelType, DbModelType>>> {
+  ): Promise<Array<DocumentSnapshot<any, any>>> {
     validateMinNumberOfArguments(
       'Firestore.getAll',
       documentRefsOrReadOptions,
       1
     );
 
-    const {documents, fieldMask} = parseGetAllArguments<
-      AppModelType,
-      DbModelType
-    >(documentRefsOrReadOptions);
+    const {documents, fieldMask} = parseGetAllArguments(
+      documentRefsOrReadOptions
+    );
     const tag = requestTag();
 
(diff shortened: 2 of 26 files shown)


MarkDuckworth:
I updated the name to `feat!:...` to indicate that this is a breaking change.

dconeybe (changes requested):
There are some pervasive updates required for this PR. Any function that takes a DocumentReference, CollectionReference, etc. _must_ have its template parameter updated to take `<AppModelType, DbModelType extends DocumentData>`. For example, `BulkWriter.set()` and `BulkWriter.delete()`. This _includes_ function signatures that use `unknown` as the type for the first type parameter.

Right now this isn't showing up as a compilation error because nodejs-firestore depends on TypeScript 4.7.4; however, if you switch it to 5.1.3 in package.json you get 536 build errors when running `npm install`. This is because the newer version of TypeScript does more rigorous type checking, complaining about incompatible type usages that the older version did not detect.

These build errors should be fixed so that we don't break customers who are using the newer versions of TypeScript. Note that this had to be done in the web sdk too.

If you're using IntelliJ you can select the newer version of TypeScript and watch all of the red squigglies show up :)

dev/src/watch.ts
```diff
@@ -53,7 +54,7 @@ export const WATCH_IDLE_TIMEOUT_MS = 120 * 1000;
 /*!
  * Sentinel value for a document remove.
  */
-const REMOVED = {} as DocumentSnapshotBuilder<unknown>;
+const REMOVED = {} as DocumentSnapshotBuilder<any, any>;
```
dconeybe, line 57: Yuck. I guess this is fine since it's an internal implementation detail. A `Symbol` would be more appropriate IMO (but that 's a change for another PR).

dev/src/write-batch.ts
```diff
@@ -89,7 +89,7 @@ export class WriteResult implements firestore.WriteResult {
    * @param {*} other The value to compare against.
    * @return true if this `WriteResult` is equal to the provided value.
    */
-  isEqual(other: firestore.WriteResult): boolean {
+  isEqual(other: unknown): boolean {
```
dconeybe, line 92: nit: This isEqual() too

dev/test/document.ts
```diff
@@ -2073,9 +2073,13 @@ describe('update document', () => {
     }).to.throw(INVALID_ARGUMENTS_TO_UPDATE);
 
     expect(() => {
-      firestore
-        .doc('collectionId/documentId')
-        .update({foo: 'bar'}, {exists: true}, 'foo');
+      firestore.doc('collectionId/documentId').update(
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
dconeybe, line 2078: Would `// @ts-expect-error` work here, instead of `@ts-ignore`, and the corresponding lint check disabling?

dev/test/query.ts
```diff
@@ -898,7 +898,7 @@ describe('query interface', () => {
 
     return createInstance(overrides).then(async firestoreInstance => {
       firestore = firestoreInstance;
-      const coll = await firestore
+      const coll = firestore
```
dconeybe, line 901: How did this code _ever_ work lol?

types/firestore.d.ts
```diff
@@ -561,7 +578,7 @@ declare namespace FirebaseFirestore {
      * @param other The `GeoPoint` to compare against.
      * @return true if this `GeoPoint` is equal to the provided one.
      */
-    isEqual(other: GeoPoint): boolean;
+    isEqual(other: unknown): boolean;
```
dconeybe, line 581: nit: Fix this `isEqual()`, and others in this file, to be parameterized.

types/firestore.d.ts
```diff
@@ -1519,8 +1562,8 @@ declare namespace FirebaseFirestore {
     where(
       fieldPath: string | FieldPath,
       opStr: WhereFilterOp,
-      value: any
-    ): Query<T>;
+      value: unknown
```
dconeybe, line 1565: Was it necessary to change `any` to `unknown`? If no, please revert it back to `any` in order to keep this PR focused.

types/firestore.d.ts
```diff
@@ -1616,7 +1659,7 @@ declare namespace FirebaseFirestore {
      * @param snapshot The snapshot of the document to start after.
      * @return The created Query.
      */
-    startAt(snapshot: DocumentSnapshot<any>): Query<T>;
+    startAt(snapshot: DocumentSnapshot): Query<AppModelType, DbModelType>;
```
dconeybe, line 1662: startAt(), startAfter(), endAt(), and endBefore() should be parameterized rather than hardcoding the default type parameters for the given DocumentSnapshot.

types/firestore.d.ts
```diff
@@ -1627,7 +1670,7 @@ declare namespace FirebaseFirestore {
      * of the query's order by.
      * @return The created Query.
      */
-    startAt(...fieldValues: any[]): Query<T>;
+    startAt(...fieldValues: unknown[]): Query<AppModelType, DbModelType>;
```
dconeybe, line 1673: Please undo the change from `any` to `unknown` unless it is strictly required for this PR.

types/firestore.d.ts
```diff
@@ -2377,7 +2463,7 @@ declare namespace FirebaseFirestore {
      * @param documentSnapshot A `DocumentSnapshot` to add.
      * @returns This instance.
      */
-    add<T>(documentSnapshot: DocumentSnapshot<T>): BundleBuilder;
+    add(documentSnapshot: DocumentSnapshot): BundleBuilder;
```
dconeybe, line 2466: add() should be parameterized with `AppDataType` and `DbDataType`. Also on line 2476.

dconeybe (changes requested):
Thanks for updating the PR for the new version of TypeScript. I've done a thorough review and left a lot of nit-picky comments. It appears that _some_ of the changes in this PR are "cleanup" that aren't strictly related to the purpose of this PR. If that is the case, please move those cleanups to a separate PR (which I will be happy to review) in order to keep this PR as focused as possible. This PR is very delicate due to the subtle public API changes and any random, unnecessary changes immediately raise red flags to me because I'm not sure whether they were done as general cleanup or to get this PR to compile, and if the latter, then I want to dig in to make sure I understand. Thanks!

dev/src/reference.ts
```diff
@@ -249,8 +256,8 @@ export class DocumentReference<T = firestore.DocumentData>
    * }):
    * ```
    */
-  get parent(): CollectionReference<T> {
-    return new CollectionReference<T>(
+  get parent(): firestore.CollectionReference<AppModelType, DbModelType> {
```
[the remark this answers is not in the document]
tom-andersen: I was able to fond root cause and reversed this.

dev/src/reference.ts
```diff
@@ -1210,11 +1250,17 @@ export class QueryOptions<T> {
    * @private
    * @internal
    */
-  static forCollectionGroupQuery<T = firestore.DocumentData>(
+  static forCollectionGroupQuery<
+    AppModelType,
```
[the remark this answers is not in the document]
tom-andersen: The default is not needed in this context. Types are picked up from `converter` parameter or they need to be explicit. Nothing in between.

types/firestore.d.ts
```diff
@@ -2474,7 +2583,7 @@ declare namespace FirebaseFirestore {
     static where(
       fieldPath: string | FieldPath,
       opStr: WhereFilterOp,
-      value: unknown
+      value: any
```
dconeybe, line 2586: nit: Change `any` back to `unknown` unless this change is strictly required for the main changes in this PR.

types/firestore.d.ts
dconeybe, line 1: The following methods also need to be parameterized with `<AppModelType, DbModelType extends DocumentData>`:

* Firestore.recursiveDelete (line 451)
* Transaction.update() (line 741)
* Transaction.delete() (line 755)
* BulkWriter.update() (line 898)
* BulkWriter.onWriteResult() (line 912)
* BulkWriterError.documentRef (line 1007)
* WriteBatch.update() (line 1115)
* WriteBatch.delete() (line 1129)

dev/src/bulk-writer.ts
dconeybe, line 1: I think `BulkWriterOperation` needs to be parameterized with `<AppModelType, DbModelType extends DocumentData>` because `DocumentReference<unknown>` is no longer valid because it pins the 2nd type parameter to `DocumentData`.

dconeybe (commented):
Please take a close look at all occurrences of DocumentReference<unknown> and DocumentReference<any> and CollectionReference<unknown> and CollectionReference<any>, etc. because each such occurrence likely needs to replaced by the `AppModelType, DbModelType`. This is because specifying the first parameter only pins the 2nd type parameter as `DocumentData`, which isn't what we want. Basically, audit all occurrences of `any` and `unknown` to make sure we're not unintentionally pinning the 2nd type parameter to DocumentData.

dev/src/reference.ts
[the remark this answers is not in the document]
tom-andersen: This signature is never exposed to the customer. I think we need to constrain ourselves to `types/firestore.d.ts` when we discuss backwards compatibility of signatures.

The consistent approach to method parameter types is to omit defaults. Only on classes do we provide them. On methods, the types can be inferred from parameters. So the consistent approach is what I have implemented.

types/firestore.d.ts
tom-andersen: Please see my other comment.

Instead of introducing type parameters where none existed before, I simply add types `<any, any>`. This makes most sense in cases where type is irrelevant, and follows existing pattern for `BulkWriter` and `WriteBatch`.

tom-andersen (commented):
There might need to be some more discussion on the right approach. I have a feeling we might disagree on some of the changes. There are cases where we want to use `any` and `unknown`, and those discussions might be better handled over GVC.

dconeybe (changes requested):

dev/src/watch.ts
```diff
@@ -214,10 +233,11 @@ abstract class Watch<T = firestore.DocumentData> {
    * @internal
    *
    * @param firestore The Firestore Database client.
+   * @param _converter
```
dconeybe, line 236: nit: Remove this `@param _converter` if you can, since it's unrelated to this PR.

dconeybe (changes requested):
Soooooo close! Just a few very minor nits.

types/firestore.d.ts
```diff
@@ -1396,7 +1394,10 @@ declare namespace FirebaseFirestore {
      * `null` removes the current converter.
      * @return A DocumentReference<U> that uses the provided converter.
      */
-    withConverter<NewAppModelType, NewDbModelType extends DocumentData>(
+    withConverter<
```
dconeybe, line 1397: Please update the definitions of `withConverter()` functions in the implementation files collection-group.ts and reference.ts as well, so that they are consistent with the public signature.

dev/src/index.ts
```diff
@@ -1285,11 +1286,11 @@ export class Firestore implements firestore.Firestore {
    * });
    * ```
    */
-  getAll<T>(
+  getAll(
```
dconeybe, line 1289: After seeing that awful `this as unknown as DocumentReference` kludge in `DocumentReference.getAll()` I realized that this internal implementation signature was different from the one in `firestore.d.ts`.

So I think we should get rid of the kludge and change this to

```ts
getAll<AppModelType, DbModelType extends DocumentData>(
  ...documentRefsOrReadOptions: Array<
    firestore.DocumentReference<AppModelType, DbModelType> | firestore.ReadOptions
  >
): Promise<Array<DocumentSnapshot<AppModelType, DbModelType>>> {
```

Since the definition in `firestore.d.ts` pins the type parameters to `DocumentData` and the internal call to `getAll()` pins the type parameters to its types, there is no need for `getAll()` to support heterogenous type parameters. So we can be consistent with the "old" signature and take both type parameters.

dconeybe (changes requested):
Two more tiny comments.

types/firestore.d.ts
```diff
@@ -390,11 +409,11 @@ declare namespace FirebaseFirestore {
      * @return A Promise that resolves with an array of resulting document
      * snapshots.
      */
-    getAll(
+    getAll<AppModelType, DbModelType extends DocumentData>(
```
dconeybe, line 412: I had originally meant for the declaration of `getAll()` here to remain pinned to DocumentData (despite the implementation in index.ts being generalized), just like it was before the PR. But now that I see that the implementation indeed preserves the type converter, it seems fine to generalize it here. That being said, we should add some test coverage for calling getAll() with a type converter; otherwise, there is no direct test coverage for calling getAll() with type converters. Maybe consider generalizing getAll() in a separate PR so that the corresponding test coverage can be added in a smaller PR and we can get this one merged.

dconeybe (changes requested):
Please add test coverage for calling FirebaseFirestore.getAll() with document references that have type converters, or pin the type parameters back to DocumentData, as it was before this PR.

dconeybe:
> Please add test coverage for calling FirebaseFirestore.getAll() with document references that have type converters, or pin the type parameters back to DocumentData, as it was before this PR.

If you choose to pin the parameters, ONLY pin the types in firestore.d.ts; in index.ts, leave the signature as-is. It is true that they are somewhat out of sync, but this out-of-syncness reduces the diff of this PR and keeps it focussed.

dconeybe (approved):

sofisl (approved):

PR #1906, opened 2023-09-25
chore: Make compile with NodeJS v20.5.0

tom-andersen:
An external change to CI appears to run code on NodeJS v20.5.0.

Unfortunately, this breaks our hacky override of Timeout. So as a quick fix, I made it even hackier. Since we only use this for tests, this should be fine.

### DIFF
--- a/dev/src/backoff.ts
+++ b/dev/src/backoff.ts
@@ -97,7 +97,7 @@ export function setTimeoutHandler(
     // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66176/files#diff-e838d0ace9cd5f6516bacfbd3ad00d02cd37bd60f9993ce6223f52d889a1fdbaR122-R126
     //
     // Adding `[Symbol.dispose](): void;` cannot be done on older versions of
-    // NodeJS. So we simply cast to `NodeJS.Timout`.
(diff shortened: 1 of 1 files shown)


dev/src/backoff.ts
```diff
@@ -92,7 +92,12 @@ export function setTimeoutHandler(
       [Symbol.toPrimitive]: () => {
         throw new Error('For tests only. Not Implemented');
       },
-    };
+    } as unknown as NodeJS.Timeout;
+    // `NodeJS.Timeout` type signature change:
+    // https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66176/files#diff-e838d0ace9cd5f6516bacfbd3ad00d02cd37bd60f9993ce6223f52d889a1fdbaR122-R126
+    //
+    // Adding `[Symbol.dispose](): void;` cannot be done on older versions of
+    // NodeJS. So we simply cast to `NodeJS.Timout`.
```
dconeybe, line 100: nit: Spelling error: `Timout`

dconeybe (approved):
Just one minor typo

dev/src/backoff.ts
tom-andersen: Fixed

PR #1934, opened 2023-10-26
feat: Support for Query Profiling

ehsannas:


ehsannas:
@MarkDuckworth  Please take a look.
Notes:
1- The protos haven't landed in `main`, so the "base" branch for this PR is `query_profiling_base` in which I have manually added the protos. And the branch for this pr (`query_profiling_impl`) builds on top of `query_profiling_base`.
2- The tests are expected to fail as the requests get rejected by the backend.

types/firestore.d.ts
```diff
@@ -1533,6 +1533,18 @@ declare namespace FirebaseFirestore {
     | 'not-in'
     | 'array-contains-any';
 
+  /**
+   * InformationalQueryPlan contains information about the query plan.
+   * Contents are subject to change.
```
MarkDuckworth, line 1538: Wondering if all readers will get the point of this comment, maybe we need to go further and explain what that means for someone coding against this type.

MarkDuckworth (commented):
Overall looks good. There are some thoughts for your consideration.

types/firestore.d.ts
```diff
@@ -1533,6 +1533,18 @@ declare namespace FirebaseFirestore {
...
+   */
+  export type InformationalQueryPlan = Record<string, unknown>;
+
+  /**
+   * InformationalQueryExecutionStats contains statistics about the
+   * execution of the query. Contents are subject to change.
```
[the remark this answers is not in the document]
ehsannas: I'm open to adding `toString`, but not sure if it's necessary.

The intent is certainly for logging and/or displaying in a dashboard/console.

The "informational" objects are just key-value pairs where keys are strings and values are stringify-able (primitive) types: numbers, strings, arrays, and other key-value pairs.

So, when I run

```js
let foo : string = JSON.stringify(profile.stats);
console.log(foo);
```
or simply:

```js
console.log(profile.stats);
```

I get the proper/intended [output](https://paste.googleplex.com/4757517406568448). WDYT?

types/firestore.d.ts
```diff
@@ -2739,6 +2809,29 @@ declare namespace FirebaseFirestore {
...
     static and(...filters: Filter[]): Filter;
   }
+
+  /**
+   * A QueryProfileInfo contains information about planning, execution, and results of a query.
+   * All informational content are subject to change.
+   */
+  export interface QueryProfileInfo<T> {
+    /**
+     * A Map that contains information about the query plan.
```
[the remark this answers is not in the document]
ehsannas: Fixed.

types/firestore.d.ts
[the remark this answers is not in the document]
MarkDuckworth: Agreed, thanks for the detailed reply. I guess my concern is covered by this and the documentation that goes with this feature.

dev/src/convert.ts
```diff
@@ -161,6 +162,45 @@ export function detectValueType(proto: ProtobufJsValue): string {
...
 
+/**
+ * Detects the value kind from a Proto3 JSON `google.protobuf.Value` proto.
+ *
+ * @private
+ * @internal
+ * @param proto The `firestore.v1.Value` proto.
+ * @return The string value for 'valueType'.
+ */
+export function detectGoogleProtobufValueKind(proto: IValue): string {
```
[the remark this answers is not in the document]
MarkDuckworth: This sounds good to me.

types/firestore.d.ts
MarkDuckworth: > How would you recommend expanding the above comment?

Maybe something like this:

```
  /**
   * InformationalQueryPlan contains information about the query plan. This should be inspected or logged,
   * because the contents are intended to be human readable. Contents are subject to change and it is advised
   * to not program against this object.
```

types/firestore.d.ts
```diff
@@ -1756,6 +1768,34 @@ declare namespace FirebaseFirestore {
...
+     */
+    explain(): Promise<InformationalQueryPlan>;
+
+    /**
+     * Plans and executes this query. Returns a Promise that will be resolved
+     * with the planning information, statistics from the query execution, and
+     * the query results.
+     *
+     * Note: the information included in the output of this function is subject
+     * to change.
```
MarkDuckworth, line 1790: Consider adding additional clarifying information.

```
 * Note: `explainAnalyze()` will get query results like the `get()` function, however `explainAnalyze()` uses
 * a different retry strategy than `get()`, so it is recommended to only use `get()` in production code.
```

dev/src/serializer.ts
```diff
@@ -286,6 +289,75 @@ export class Serializer {
...
+        return array;
+      }
+      case 'structValue': {
+        const obj: DocumentData = {};
+        const fields = proto.structValue?.fields;
+        if (fields) {
+          for (const prop of Object.keys(fields)) {
+            obj[prop] = this.decodeGoogleProtobufValue(fields[prop]);
+          }
+        }
```
MarkDuckworth, line 331: nit: there's an opportunity to re-use code in `decodeGoogleProtobufStruct`.

dev/src/reference.ts
```diff
@@ -2699,6 +2743,115 @@ export class Query<
...
+                }
+              );
+              resolve(
+                new QueryProfileInfo<QuerySnapshot<AppModelType, DbModelType>>(
+                  planInfo,
+                  executionStats,
+                  querySnapshot
+                )
+              );
+            });
```
MarkDuckworth, line 2848: I agree that this rewrite of `_getQueryProfileInfo(...)` is simpler than using `_stream(...)` and all of that baggage. Nice change here.

nit: It's unfortunate that there are two implementations of creating a QuerySnapshot between `get()` and `explainAnalyze()`. That brings some risk of deviation in these two implementations in the future.

MarkDuckworth (approved):
I have a few comments for your consideration, but otherwise this looks good.

dev/src/reference.ts
ehsannas: Agreed. IMO to address the risk of deviation, we first need to do some cleanups in `_stream(...)`, and then refactor the two functions into one. I'll file a ticket for this work.

types/firestore.d.ts
ehsannas: Done

types/firestore.d.ts
ehsannas: Done

dev/src/serializer.ts
ehsannas: Done

MarkDuckworth (approved):
New changes look great.

Issue #1929, opened 2023-10-24
ClientPool.clientFactory crashing in serverless environments


vdemcak:
I'm trying to execute Firestore queries inside of Google Cloud Functions utilizing Qwik together with `routeLoader$` in order to fetch and pre-render data server-side. Everything works fine in development and when using the preview command, but upon deploying the app to Firebase Hosting and Cloud Functions ([Using the Firebase integration](https://qwik.builder.io/docs/deployments/firebase/)), the app returns Error 500 - Internal Server Error.

Upon opening the Cloud Function logs, I can see, that it fails with the following error message which leads me to believe, that there is something wrong with the `@google-cloud/firestore` and `firebase-admin`(Which utilizes this package under the hood as far as I know):

```
TypeError: Cannot set properties of undefined (setting 'fromParams')
    at ClientPool.clientFactory (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:167322:43)
    at ClientPool.acquire (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:153428:31)
    at ClientPool.run (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:153535:28)
    at file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:168168:28
    at Firestore._retry (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:168038:26)
    at async file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:145922:33
Caused by: Error
    at AggregateQuery._get (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:145879:21)
    at AggregateQuery.get (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:145869:19)
    at Object.s_eNFdi8qBn48 (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:168225:68)
    at Object.invoke (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:1461:42)
    at file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:4100:131
    at then (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:137:81)
    at file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:4091:14
    at Object.qrl (file:///workspace/server/assets/@qwik-city-plan-c264a646.mjs:4071:18)
[...]

PR #1935, opened 2023-10-27
fix: Remove temporary header encoding workaround

tom-andersen:
Remove temporary workaround to header encoding issue b/291780066.

The workaround appears to cause trouble for some customers.
https://github.com/googleapis/nodejs-firestore/issues/1929


MarkDuckworth (approved):
LGTM. The lint error should be fixed though.

PR #1937, opened 2023-10-28
feat: Expose databaseId getter on Firestore class

### LINKED ISSUE #1936: Expose databaseId and projectId on Firestore class
Given the introducing of multiple databases it becomes more and more often that you want to know from where a Reference or snapshot was created for.

Right now there is no way to go from a `DocumentReference` or a `DocumentSnapshot` to the database or the project.

It should simply be possible to do the following:

```ts
reference.firestore.databaseId
reference.firestore.projectId
```

Those getters already exist but they are marked a private/internal making accessing them in Typescript unavailable.

Right now you need to do some fishy stuff accessing internal variables, silencing Typescript errors (using `as any`) or other unofficial, undocumented things.

Given those things are not officially supported and APIs may change without proper breaking changes this makes depending on those workarounds, not a solution.

IchordeDionysos:

Fixes #1936

### DIFF
--- a/dev/src/index.ts
+++ b/dev/src/index.ts
@@ -758,6 +758,9 @@ export class Firestore implements firestore.Firestore {
   /**
    * Returns the Project ID for this Firestore instance. Validates that
    * `initializeIfNeeded()` was called before.
+   *
+   * @private
+   * @internal
    */
   get projectId(): string {
     if (this._projectId === undefined) {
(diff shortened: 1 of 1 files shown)


tom-andersen:
This seems like a reasonable and simple change. However, since this does affect the API surface, I will need to have this approved internally.

Hang on while I handle our internal process for these types of changes.

tom-andersen:
Hey. The internal API review is ok with exposing databaseId, but projectId was asked to take a different approach.

If you restrict change to just exposing databaseId, I will merge right away.

IchordeDionysos:
@tom-andersen changed

tom-andersen (approved):

PR #1946, opened 2023-11-14
fix: update retry policy to not retry streams that have not made progress receiving documents

MarkDuckworth:
BEGIN_COMMIT_OVERRIDE
fix: update retry policy to not retry streams that have not made progress receiving documents
END_COMMIT_OVERRIDE

### DIFF
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2571,7 +2571,12 @@ export class Query<
   }
 
   _hasRetryTimedOut(methodName: string, startTime: number): boolean {
-    return Date.now() - startTime >= getTotalTimeout(methodName);
+    const totalTimeout = getTotalTimeout(methodName);
+    if (totalTimeout === 0) {
+      return false;
+    }
+
+    return Date.now() - startTime >= totalTimeout;
   }
 
   /**
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2637,6 +2637,9 @@ export class Query<
 
         let streamActive: Deferred<boolean>;
         do {
+          // Set lastReceivedDocument to null before each retry attempt to ensure the retry makes progress
+          lastReceivedDocument = null;
+
           streamActive = new Deferred<boolean>();
           backendStream = await this._firestore.requestStream(
             'runQuery',
--- a/dev/src/reference.ts
+++ b/dev/src/reference.ts
@@ -2650,18 +2650,30 @@ export class Query<
             // If a non-transactional query failed, attempt to restart.
             // Transactional queries are retried via the transaction runner.
             if (!transactionId && !this._isPermanentRpcError(err, 'runQuery')) {
-              logger(
-                'Query._stream',
-                tag,
-                'Query failed with retryable stream error:',
-                err
-              );
-              // Enqueue a "no-op" write into the stream and resume the query
-              // once it is processed. This allows any enqueued results to be
-              // consumed before resuming the query so that the query resumption
-              // can start at the correct document.
-              stream.write(NOOP_MESSAGE, () => {
-                if (lastReceivedDocument) {
+              // Only streams that have received documents should be
+              // restarted to avoid rerunning long-running or high-cost
+              // queries that are not progressing to completion.
+              if (lastReceivedDocument) {
+                logger(
+                  'Query._stream',
+                  tag,
+                  'Query failed with retryable stream error:',
+                  err
+                );
+                // Enqueue a "no-op" write into the stream and resume the query
+                // once it is processed. This allows any enqueued results to be
+                // consumed before resuming the query so that the query resumption
+                // can start at the correct document.
+                stream.write(NOOP_MESSAGE, () => {
+                  // `backendStream` was unpiped at the beginning of the handler
+                  // for `backendStream.on('error', ...)`, so we don't expect
+                  // `lastReceivedDocument` to be updated during the noop write,
+                  // but a check here is included for this unexpected state.
+                  assert(
+                    lastReceivedDocument !== null,
+                    'lastReceivedDocument is unset'
+                  );
+
                   // Restart the query but use the last document we received as
                   // the query cursor. Note that we do not use backoff here. The
                   // call to `requestStream()` will backoff should the restart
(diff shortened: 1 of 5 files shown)


dev/src/reference.ts
```diff
@@ -2650,18 +2650,30 @@ export class Query<
...
-              // Enqueue a "no-op" write into the stream and resume the query
-              // once it is processed. This allows any enqueued results to be
-              // consumed before resuming the query so that the query resumption
-              // can start at the correct document.
-              stream.write(NOOP_MESSAGE, () => {
-                if (lastReceivedDocument) {
+              // Only streams that have received documents should be
+              // restarted to avoid rerunning long-running or high-cost
+              // queries that are not progressing to completion.
+              if (lastReceivedDocument) {
```
tom-andersen, line 2656: I am very suspicious of this code, because I do not see `lastReceivedDocument` ever being reset to null during a retry.

I am also slightly concerned with accessing `lastReceivedDocument` outside of write callback, since we are dependent on stream pushing items through before this logic is evaluated. According to your tests this works, but it still smells like a race condition.

dev/test/query.ts
```diff
@@ -773,6 +788,135 @@ describe('query interface', () => {
...
+    // should result in the stream retrying indefinitely.
+    const retryableErrorCodes = [
+      Status.DEADLINE_EXCEEDED,
+      Status.UNAVAILABLE,
+      Status.INTERNAL,
+      Status.UNAVAILABLE,
+    ];
+
+    const overrides: ApiOverride = {
+      runQuery: () => {
```
tom-andersen, line 843: Is it possible to verify the request being sent to `runQuery`. We want to see that `startAfter` is being set correctly. I think this can be derived from `attempts`.

dev/src/reference.ts
MarkDuckworth: > I am very suspicious of this code, because I do not see lastReceivedDocument ever being reset to null during a retry.

This was a bug that my tests did not catch. Good catch. Fixed and an additional test added.

> I am also slightly concerned with accessing lastReceivedDocument outside of write callback, since we are dependent on stream pushing items through before this logic is evaluated. According to your tests this works, but it still smells like a race condition.

I also was initially. Here's why I decided on this. The noop write only makes sense if we are resuming the query, so we do need to check `lastReceivedDocument` outside of the write callback. I added an assertion to validate that `lastReceivedDocument` did not change and I added a comment regarding why it should never change.

```
// `backendStream` was unpiped at the beginning of the handler
// for `backendStream.on('error', ...)`, so we don't expect
// `lastReceivedDocument` to be updated during the noop write,
// but a check here is included for this unexpected state.
assert(
  lastReceivedDocument !== null,
  'lastReceivedDocument is unset'
);
```

dev/test/query.ts
MarkDuckworth: Good call, this was something I also wanted to do but forgot.

MarkDuckworth:
Aggregate queries should not be retried.

dev/test/query.ts
```diff
@@ -773,6 +788,163 @@ describe('query interface', () => {
...
+        });
+    });
+  }
+
+  it('handles retryable exception until progress stops with heartbeat', () => {
+    handlesRetryableExceptionUntilProgressStops(true);
+  });
+
+  it('handles retryable exception until progress stops without heartbeat', () => {
+    handlesRetryableExceptionUntilProgressStops(false);
```
tom-andersen, line 910:
```suggestion
    return handlesRetryableExceptionUntilProgressStops(false);
```

dev/test/query.ts
```diff
@@ -773,6 +788,163 @@ describe('query interface', () => {
...
+          const initilizationRetries = withHeartbeat ? 1 : 5;
+          expect(attempts).to.equal(
+            initializationsWithProgress + initilizationRetries
+          );
+        });
+    });
+  }
+
+  it('handles retryable exception until progress stops with heartbeat', () => {
+    handlesRetryableExceptionUntilProgressStops(true);
```
tom-andersen, line 906: You need to `return` since this is async.
```suggestion
    return handlesRetryableExceptionUntilProgressStops(true);
```

dev/src/reference.ts
tom-andersen: This still makes me uncomfortable because of buffering:
https://nodejs.org/api/stream.html#buffering

If you add `execSync('sleep 1');` to Transform above (in an attempt to cause buffering) does it still work?

tom-andersen (changes requested):

dev/test/query.ts
```diff
@@ -773,6 +788,209 @@ describe('query interface', () => {
...
+          expect(err.message).to.equal('DEADLINE_EXCEEDED error message');
+
+          // The heartbeat initialized the stream before there was a stream
+          // exception, so we only expect a single attempt at streaming.
+          expect(attempts).to.equal(1);
+        });
+    });
+  });
+
+  function handlesRetryableExceptionUntilProgressStops(withHeartbeat: boolean) {
```
tom-andersen, line 826: nit: Can we apply async/await consistently over returning promises. It will also help readability if the function has a return type.

dev/test/query.ts
```diff
@@ -773,6 +788,209 @@ describe('query interface', () => {
...
+      query._hasRetryTimedOut = () => false;
+      return query
+        .get()
+        .then(() => {
+          throw new Error('Unexpected success in Promise');
+        })
+        .catch(err => {
+          expect(err.message).to.equal('test error message');
+
+          // 20 stream initialization with data that progresses the steam
```
tom-andersen, line 897: I don't think 20 is correct anymore.

dev/test/query.ts
```diff
@@ -773,6 +788,209 @@ describe('query interface', () => {
...
+          // followed by an error. The final X+1 initialization does not get data
+          // before the error, so we may also get retries during stream initialization.
+          const initilizationRetries = withHeartbeat ? 1 : 5;
```
tom-andersen, line 900: Why does the heartbeat make a difference?

dev/src/reference.ts
```diff
@@ -2569,6 +2570,10 @@ export class Query<
     return isPermanentRpcError(err, methodName);
   }
 
+  _hasRetryTimedOut(methodName: string, startTime: number): boolean {
+    return Date.now() - startTime >= getTotalTimeout(methodName);
```
tom-andersen, line 2574: You should return false if `getTotalTimeout(methodName)` returns 0.

tom-andersen (changes requested):

dev/test/query.ts
MarkDuckworth: I added a return type. I also agree that using async/await is more readable than handling promise results with `then`, `catch`, and `finally`. However, in this case, the test implementation is consistent with other similar tests in this file. I don't think be benefit is worth refactoring all the similar tests.

dev/test/query.ts
tom-andersen: Thanks! That makes perfect sense now! :)

tom-andersen (approved):
LGTM

PR #1958, opened 2023-12-12
docs: Re-write the sample code for FirestoreDataConverter.

dconeybe:
The old sample had blatant syntax errors and was lacking in its demonstration of the power of type converters.

These changes are ported from https://github.com/firebase/firebase-js-sdk/pull/7673


MarkDuckworth (approved):
Looks like a clean port.

PR #1967, opened 2024-01-03
fix: Make transaction rollback best effort.

tom-andersen:
Internal tracking: b/316023452

A rollback is meant to be best effort. If the transaction has already expired, it is possible for the rollback to fail due to transaction no longer existing in Firestore. The retry logic will use attempts to rollback, and in the case where transaction no longer exists, all attempts to be exhausted attempted to rollback transaction.

This PR makes the rollback best effort, simply logging any rollback error and continuing.

### DIFF
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -481,7 +481,20 @@ export class Transaction implements firestore.Transaction {
       transaction: this._transactionId,
     };
 
-    return this._firestore.request('rollback', request, this._requestTag);
+    const promise: Promise<void> = this._firestore.request(
+      'rollback',
+      request,
+      this._requestTag
+    );
+
+    return promise.catch(reason => {
+      logger(
+        'Firestore.runTransaction',
+        this._requestTag,
+        'Best effort to rollback failed with error:',
+        reason
+      );
+    });
   }
 
   /**
--- a/dev/test/transaction.ts
+++ b/dev/test/transaction.ts
@@ -574,7 +574,6 @@ describe('failed transactions', () => {
           begin({transactionId: 'foo1'}),
           commit('foo1', /* writes=*/ undefined, serverError),
           rollback('foo1', serverError),
-          rollback('foo1'),
           backoff(),
           begin({
             transactionId: 'foo2',


dev/src/transaction.ts
```diff
@@ -481,7 +481,20 @@ export class Transaction implements firestore.Transaction {
...
+    );
+
+    return promise.catch(reason => {
+      logger(
+        'Firestore.runTransaction',
+        this._requestTag,
+        'Best effort to rollback failed with error:',
+        reason
+      );
+    });
```
ehsannas, line 497: I guess the only question is whether there are rollback failures that should be retried?

maybe `MISSING_TRANSACTION_HANDLE` should be the only one that's best-effort?

tom-andersen: From the documentation "Note that a rollback itself might fail, so the rollback should be a best-effort attempt only."

https://cloud.google.com/datastore/docs/best-practices#api_calls

However, I see your point.

This could be improved such that retryable errors from rollback cause rollback to be retried, except INTERNAL which happens when MISSING_TRANSACTION_HANDLE.

BUT, even if rollback experiences a non-retryable error, the transaction should still retry. That was not the case when I looked at Java, which seems completely wrong even if it was baked into a test case.

The phrasing "Best effort" leaves some nuance to be interpreted.

Ideally MISSING_TRANSACTION_HANDLE should just be a no-op on Firestore backend, and NOT throw an INTERNAL error. This code might simply be working around what should be a change to the backend.

ehsannas: > Ideally MISSING_TRANSACTION_HANDLE should just be a no-op on Firestore backend, and NOT throw an INTERNAL error. This code might simply be working around what should be a change to the backend.

+1

ehsannas (approved):

dev/src/transaction.ts
ehsannas: > This could be improved such that retryable errors from rollback cause rollback to be retried, except INTERNAL

Let's do this then

ehsannas (changes requested):

dev/src/transaction.ts
tom-andersen: I looked into code, and the underlying GAPIC client does do retries according to retry settings. Effectively, that means currently there is appropriate backoff, and 6 attempts to rollback before failure.

This second level of retry on a higher level was a complete duplication and would simply multiply the number of attempts with additional backoff.

In addition, after some investigation, the GAPIC level retry respects the error codes from service config. So it actually does a better job of determining when to retry than the higher transaction level retry.

In other words, we should be fine.

ehsannas: SGTM. Thanks

ehsannas (approved):

PR #2002, opened 2024-02-23
fix: Optimize Transaction PITR

tom-andersen:
For transactions with read time, the begin transaction and commit can be omitted, thereby decreasing latency.

b/324245706

### DIFF
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -79,6 +79,7 @@ export class Transaction implements firestore.Transaction {
    * @param firestore The Firestore Database client.
    * @param requestTag A unique client-assigned identifier for the scope of
    * this transaction.
+   * @param transactionOptions The user-defined options for this transaction.
    */
   constructor(
     firestore: Firestore,
@@ -536,9 +537,6 @@ export class Transaction implements firestore.Transaction {
    * @internal
    * @param updateFunction The user function to execute within the transaction
    * context.
-   * @param requestTag A unique client-assigned identifier for the scope of
-   * this transaction.
-   * @param options The user-defined options for this transaction.
    */
   async runTransaction<T>(
     updateFunction: (transaction: Transaction) => Promise<T>
@@ -583,6 +581,15 @@ export class Transaction implements firestore.Transaction {
     return Promise.reject(lastError);
   }
 
+  /**
+   * Make single attempt to execute `updateFunction()` and commit the
+   * transaction. Will rollback upon error.
+   *
+   * @private
+   * @internal
+   * @param updateFunction The user function to execute within the transaction
+   * context.
+   */
   async runTransactionOnce<T>(
     updateFunction: (transaction: Transaction) => Promise<T>
   ): Promise<T> {
--- a/dev/src/transaction.ts
+++ b/dev/src/transaction.ts
@@ -21,7 +21,7 @@ import * as proto from '../protos/firestore_v1_proto_api';
 
 import {ExponentialBackoff} from './backoff';
 import {DocumentSnapshot} from './document';
-import {Firestore, WriteBatch} from './index';
+import {DEFAULT_MAX_TRANSACTION_ATTEMPTS, Firestore, WriteBatch} from './index';
 import {Timestamp} from './timestamp';
 import {logger} from './logger';
 import {FieldPath, validateFieldPath} from './path';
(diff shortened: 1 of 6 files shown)


dev/src/transaction.ts
```diff
@@ -460,13 +492,14 @@ export class Transaction implements firestore.Transaction {
...
    * @internal
    */
-  commit(): Promise<void> {
-    return this._writeBatch
-      ._commit({
-        transactionId: this._transactionId,
-        requestTag: this._requestTag,
-      })
-      .then(() => {});
+  async commit(): Promise<void> {
```
tom-andersen, line 495: This is purely a refactor to use `async` syntax.

dev/src/transaction.ts
```diff
@@ -443,15 +479,11 @@ export class Transaction implements firestore.Transaction {
...
-    return this._firestore
-      .request<api.IBeginTransactionRequest, api.IBeginTransactionResponse>(
-        'beginTransaction',
-        request,
-        this._requestTag
-      )
-      .then(resp => {
-        this._transactionId = resp.transaction!;
-      });
+    const resp = await this._firestore.request<
```
tom-andersen, line 482: This is purely a refactor to use `async` syntax.

dev/system-test/firestore.ts
```diff
@@ -5319,30 +5319,6 @@ describe('Transaction class', () => {
     expect(snapshot.exists).to.be.true;
     expect(snapshot.get('foo')).to.equal(1);
   });
-
```
tom-andersen, line 5322: I added client side exceptions to attempting write on readonly transaction, so this integration test is no longer valid.

dev/src/transaction.ts
```diff
@@ -568,10 +578,49 @@ export class Transaction implements firestore.Transaction {
...
+
+    try {
+      const promise = updateFunction(this);
+      if (!(promise instanceof Promise)) {
+        throw new Error(
+          'You must return a Promise in your transaction()-callback.'
+        );
+      }
+      const result = await promise;
+      if (!this._readTime) {
```
MarkDuckworth, line 608: same

MarkDuckworth (approved):
Direction looks good. There are a few comments for your consideration.

dev/src/transaction.ts
```diff
@@ -265,6 +290,9 @@ export class Transaction implements firestore.Transaction {
     documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
     data: firestore.WithFieldValue<AppModelType>
   ): Transaction {
+    if (this._readOnly) {
+      throw new Error(READ_ONLY_WRITE_ERROR_MSG);
+    }
```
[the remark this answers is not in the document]
tom-andersen: For ReadTime transactions, we don't call commit which would normally be when backend could respond. So this is less preference, and more technical requirement.

dev/src/transaction.ts
```diff
@@ -460,13 +492,14 @@ export class Transaction implements firestore.Transaction {
...
    */
-  commit(): Promise<void> {
-    return this._writeBatch
-      ._commit({
-        transactionId: this._transactionId,
-        requestTag: this._requestTag,
-      })
-      .then(() => {});
+  async commit(): Promise<void> {
+    if (this._readTime) {
```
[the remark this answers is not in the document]
tom-andersen: ReadOnly but without ReadTime are still transactions that have `BeginTransaction` and `Commit` request. That is why I have to use `_readTime`, since only in the subset of cases where the transaction if ReadOnly AND with ReadTime can we do the optimization.

However, this is also a little weird of a check that I added, since this will only occur if there is a bug in SDK. The customer never has an opportunity to call `commit()` on transactions. Calling `commit()` is done implicitly when update lambda completes. So this should never happen so long as our SDK logic is correct.

dev/src/transaction.ts
```diff
@@ -568,10 +578,49 @@ export class Transaction implements firestore.Transaction {
...
+   *
+   * @private
+   * @internal
+   * @param updateFunction The user function to execute within the transaction
+   * context.
+   */
+  async runTransactionOnce<T>(
+    updateFunction: (transaction: Transaction) => Promise<T>
+  ): Promise<T> {
+    if (!this._readTime) {
```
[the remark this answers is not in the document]
tom-andersen: Again this optimization only works for the subset of ReadOnly transaction that have ReadTime.

dev/src/transaction.ts
tom-andersen: dido

PR #2017, opened 2024-03-15
feat: Lazy-started transactions

### LINKED ISSUE #2015: Improve network latency of `runTransaction()` routine
If I correctly understand the routine for `runTransaction()`, as referenced below, then when running a *read-write* transaction, a [`documents:beginTransaction`](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.documents/beginTransaction) request is *always* issued before running the `updateFn`.

https://github.com/googleapis/nodejs-firestore/blob/e598b9daf628cbc54dc10dab80bb0f46e2a3e2a2/dev/src/transaction.ts#L593-L598

This means we wait for an entire network round trip before beginning the transaction, every attempt.

I also noticed that the [`documents:runQuery`](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.documents/runQuery) and [`documents:batchGet`](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.documents/batchGet) APIs support a `newTransaction` parameter, which would allow the SDK to lazily start a transaction on the first read request, without the additional sequential network round trip.

Lazily starting the transaction on the first read would *always* eliminate one entire sequential network round trip for every read-write transaction. I feel that this is quite a significant impact for high performance applications.

Some implications
- The transaction would be started server-side on first read (eliminate `documents:beginTransaction` and use `newTransaction`). 
- All read operations would need to use `documents:batchGet`, `documents:runQuery`, `documents: runAggregationQuery` (i.e. not `documents:get`) which is fine.
- Subsequent reads would be queued until the first read completes, so as to use that same transaction ID returned. In many cases, transaction reads would be sequential anyways, and if not, the user can improve performance of parallel initial reads using `getAll()`. The only exception is if there is a mix of document IDs and queries (cannot be batched together in one request). Regardless, currently *all* reads are queued after the `documents:beginTransaction` call anyways so there is no downside.
- If there are no reads before the transaction is committed, then [`documents:commit`](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.documents/commit) is called without a transaction.

Thoughts?

brettwillis:
Eliminates at least one round-trip network request for every transaction by the following
- Transactions are started lazily upon the first read request (or not at all if no reads) instead of by the initial separate `beginTransaction` request. All subsequent reads (and commit/rollback) are queued upon the first read operation of each transaction attempt.
- Empty transactions (no reads or writes) become a network no-op (zero requests, no begin and no commit)
- ~Eliminate superfluous rollback request for some types of commit errors (where the transaction is already aborted by the server)~
- Added an internal `getResponse` API to the `DocumentReader`, `Query` and `AggregateQuery` classes to expose the transaction ID in the response.
- The internal `_stream()` APIs now emit a "transaction response" as the first element when appropriate. 

Other opportunistic improvements
- Avoid initialising useless write batch and backoff instances on read-only transactions

Fixes #2015


dev/src/transaction.ts
```diff
@@ -449,57 +472,46 @@ export class Transaction implements firestore.Transaction {
...
+      });
+      this._transactionIdPromise = undefined;
+      this._prevTransactionId = transactionId;
+    } catch (err) {
+      // A failed commit also rolls back the transaction so if we know that the
+      // server responded with an error then we clear the transaction ID so that
+      // we don't continue on to perform a superfluous rollback
+      this._transactionIdPromise = undefined;
+      this._prevTransactionId = transactionId;
+      throw err;
```
brettwillis, line 513: I feel that this statement is true just from my experimentation with the API, but it is not documented... is it correct? I need to determine if the error was an "error response" from the server and not some other comms error for example, what is the best way to determine this?

dev/src/reference.ts
```diff
@@ -2343,28 +2348,49 @@ export class Query<
...
    * @internal
-   * @param transactionIdOrReadTime A transaction ID or the read time at which
-   * to execute the query.
+   * @param transactionOrReadTime A transaction ID, options to start a new
+   *  transaction, or timestamp to use as read time.
    */
-  _get(
-    transactionIdOrReadTime?: Uint8Array | Timestamp
+  async _get(
+    transactionOrReadTime?: Uint8Array | Timestamp | api.ITransactionOptions
```
tom-andersen, line 2355: Since the return type doesn't afford getting the new transactionId, I don't think we should allow passing in transaction options.

```suggestion
    transactionOrReadTime?: Uint8Array | Timestamp
```

tom-andersen (changes requested):
Thank you for the quite large contribution. I prefer if the rollback optimization based on error code was moved to a separate PR. I am not convinced that this change is correct. Verifying will hold up this PR. The problem is with reads having created locks, that should be released. If we wait for timeout of transaction, that can create problems for retry and other clients working on same documents.

I am giving you some preliminary feedback, but will return to do a more thorough review.

I sent you a PR to change the tests. For `readTime`, we do not need a transaction at all. This will eliminate the need to wait for first request to complete with transactionId, and this will also overcome potential transaction timeout.

dev/test/transaction.ts
```diff
@@ -132,62 +132,69 @@ function rollback(
...
+              options.newTransaction.readWrite.prevTransactionId
+            ),
+          }
+        : {},
+    };
+  } else if (options?.newTransaction?.readOnly) {
+    request.newTransaction = {
+      readOnly: options.newTransaction.readOnly.readTime
+        ? {
+            readTime:
```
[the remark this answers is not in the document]
brettwillis: Ok yes this makes sense, subsequent reads will not need to be queued. Will do.

dev/test/transaction.ts
```diff
@@ -377,75 +423,72 @@ function runTransaction<T>(
...
+    [Status.UNAVAILABLE]: {retry: true, rollback: true},
+    [Status.UNAUTHENTICATED]: {retry: true, rollback: true},
+
+    [Status.INVALID_ARGUMENT]: {retry: false, rollback: false},
+    [Status.NOT_FOUND]: {retry: false, rollback: true},
+    [Status.ALREADY_EXISTS]: {retry: false, rollback: false},
+    [Status.FAILED_PRECONDITION]: {retry: false, rollback: false},
+    [Status.OUT_OF_RANGE]: {retry: false, rollback: true},
+    [Status.UNIMPLEMENTED]: {retry: false, rollback: true},
+    [Status.DATA_LOSS]: {retry: false, rollback: true},
```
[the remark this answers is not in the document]
brettwillis: Yes I think some thorough discussion is warranted on this. I can reverse it so that it can be implemented separately.

So this behaviour is not documented anywhere, but I discovered it via experimentation with the REST API, I'll show an example below.

But it kind of makes sense, here's an obvious example: if the commit fails saying "transaction has expired", then obviously there is no need to attempt rolling back an expired transaction.

However I've found is that the behaviour extends to several other error codes too.

Here's an example you can try yourself:

(1) Hold a lock on a document (non-existent document)

```
POST /v1/projects/test-project/databases/(default)/documents:batchGet HTTP/1.1
Host: firestore.googleapis.com
Authorization: Bearer •••••
Content-Type: application/json

{
    "newTransaction": {
        "readWrite": {}
    },
    "documents": [
        "projects/test-project/databases/(default)/documents/collection_id/non_existent_doc"
    ]
}
```

(2) Attempt to write that document outside the transaction. You will find that this request is blocked until the transaction from (1) completes for fails or times out.

```
POST /v1/projects/test-project/databases/(default)/documents:batchWrite HTTP/1.1
Host: firestore.googleapis.com
Authorization: Bearer •••••
Content-Type: application/json

{
    "writes": [
        {
            "update": {
                "name": "projects/test-project/databases/(default)/documents/collection_id/non_existent_doc",
                "fields": {}
            }
        }
    ]
}
```

(3) Commit the transaction from (1) with a precondition that will fail (i.e. precondition "exists" when the document doesn't exists causes the commit to fail).

```
POST /v1/projects/test-project/databases/(default)/documents:commit HTTP/1.1
Host: firestore.googleapis.com
Authorization: Bearer •••••
Content-Type: application/json

{
    "transaction": "<the transaction ID from (1)>"
    "writes": [
        {
            "update": {
                "name": "projects/test-project/databases/(default)/documents/collection_id/non_existent_doc",
                "fields": {}
            },
            "currentDocument": {
                "exists": true
            }
        }
    ]
}
```

The commit will fail with "NOT_FOUND" (no document to update). And in that moment, you will see that the request from (2) completes successfully meaning the failed commit released the lock on that document.

You'll need to complete this sequence within 20s or so otherwise (2) may time out anyway.

We'd definitely need input from the appropriate team on which error codes mean that the transaction has been aborted, and whether or not the behaviour is reliable. That is, it needs to be documented!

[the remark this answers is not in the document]
tom-andersen: I received from clarification from Firestore. The rollback can be called in a non-blocking fashion. We do not need to wait for result of rollback before continuing.

I think making the rollback non-blocking will achieve the same latency as omitting the callback.

Feel free to implement this.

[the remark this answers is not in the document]
brettwillis: Ok I get that. I'll remove those changes.

> The rollback can be called in a non-blocking fashion. We do not need to wait for result of rollback before continuing.

Hmm, but without knowing the specifics of the RPC or REST transport, what's to say that which of the rollback or the next transaction attempt requests arrive at the backend first? I.e. they could be on different streams or connections? Isn't that asking for a race condition?

Or is it that the next attempt would be blocked at the backend until the rollback is received anyway, so it doesn't matter which arrives first?

What will the backend do if/when it receives a `readWrite: { retryTransaction: "..." }` options to retry a transaction that hasn't been committed or rolled back yet? Does it automatically roll back the provided transaction?

Finally, regarding your point about errors from reads. If the initial read operation rejects, then as it is implemented currently we will obviously have no transaction ID to roll back. I was pondering whether or not it may be possible for:
1. The backend to lazily start the transaction and establish the lock, only for something in the SDK (e.g. validation) to reject, preventing us from receiving the transaction ID, and preventing us from rolling back the transaction that was established. I figured this was both highly unlikely and inescapable. 
2. The backend to send the "transaction ID response" in the response stream and then fail/error before sending all of "read responses" into the stream. In this case it is conceivable we could still have received a transaction ID even though the read operation finally rejects. With a bit of reworking we could capture the transaction ID and roll back the transaction even when the read rejects.

What are your thoughts on that?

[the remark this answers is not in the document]
tom-andersen: The `retryTransaction` is more of a hint to backend, that can help process the transaction more efficiently. It won't create adverse behaviour if previous transaction is still active.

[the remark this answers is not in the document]
brettwillis: Ok I've reverted the conditional rollback completely and made rollback complete asynchronously.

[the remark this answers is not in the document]
brettwillis: Ok I've reverted the conditional rollback completely and made rollback complete asynchronously.

dev/src/reference.ts
brettwillis: Done, btw.

brettwillis: Done, btw.

brettwillis: Done, btw.

dev/src/transaction.ts
```diff
@@ -622,17 +619,124 @@ export class Transaction implements firestore.Transaction {
...
+    }: {
+      documents: Array<DocumentReference<AppModelType, DbModelType>>;
+      fieldMask?: FieldPath[] | null;
+    },
+    opts: Uint8Array | api.ITransactionOptions | Timestamp
+  ): Promise<{
+    transaction?: Uint8Array;
+    result: DocumentSnapshot<AppModelType, DbModelType>[];
+  }> {
+    const documentReader = new DocumentReader(this._firestore, documents);
```
tom-andersen, line 719: I think the DocumentReader constructor should take `opts` as a third parameter. If you want, you can do this refactor.

dev/src/transaction.ts
```diff
@@ -771,3 +875,24 @@ function isRetryableTransactionError(error: GoogleError): boolean {
...
+  backoff: ExponentialBackoff,
+  error?: GoogleError
+): Promise<void> {
+  if ((error?.code as number | undefined) === StatusCode.RESOURCE_EXHAUSTED) {
+    backoff.resetToMax();
+  }
+  await backoff.backoffAndWait();
+}
+
+function resolveResult<TResult>({result}: {result: TResult}): TResult {
```
tom-andersen, line 896: This is a very small code snippet. I feel the indirection of calling this method obfuscates the code.

I think inlining this method with `v => v.result` is clearer and more concise.

dev/src/transaction.ts
```diff
@@ -622,17 +619,124 @@ export class Transaction implements firestore.Transaction {
...
+        // This is the first read of the transaction so we create the appropriate
+        // options for lazily starting the transaction inside this first read op
+        let opts: api.ITransactionOptions;
+        if (this._writeBatch) {
+          opts = this._prevTransactionId
+            ? {readWrite: {retryTransaction: this._prevTransactionId}}
+            : {readWrite: {}};
+        } else {
+          opts = {readOnly: {}};
+        }
```
tom-andersen, line 659: Not sure how much better this is, but I found this logic could maybe be written clearer. Your call. This is not a deal breaker.

```suggestion
        const opts: api.ITransactionOptions = {};
        if (this._writeBatch) {
          opts.readWrite = this._prevTransactionId
            ? {retryTransaction: this._prevTransactionId}
            : {};
        } else {
          opts.readOnly = {};
        }
```

tom-andersen (changes requested):
Overall, I really like your implementation. There are a few things I would like to see changed, but then I see no reason why this shouldn't be approved and merged.

Note, there will be a delay before we can merge since this has to fit into our upcoming release schedule. We should have this out by April.

dev/src/transaction.ts
brettwillis: Hmm then wouldn't this also be the case for `readTime` and `fieldMask`? I do kind of think that all of these should actually be immutable options provided at construction rather than mutable fields that could be changed at any point before/during/after the request... Do you want me to change all of them to be provided in a single `opts` constructor parameter?

```
opts?: {
      transactionOrReadTime?: Uint8Array | api.ITransactionOptions | Timestamp;
      fieldMask?: FieldPath[];
    }
```

dev/src/transaction.ts
brettwillis: Sure

dev/src/transaction.ts
brettwillis: True

dev/src/transaction.ts
```diff
@@ -557,15 +554,19 @@ export class Transaction implements firestore.Transaction {
...
 
         return await this.runTransactionOnce(updateFunction);
       } catch (err) {
         lastError = err;
 
-        if (!this._transactionId || !isRetryableTransactionError(err)) {
+        // We may or may not have a previous transaction ID
+        // For example, the initial read operation could have failed with a
+        // retry-able error meaning we did not obtain the ID of the lazy-started
+        // transaction so we will retry without the ID of the previous attempt
```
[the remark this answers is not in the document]
brettwillis: Ok sure will remove

dev/src/transaction.ts
```diff
@@ -509,14 +491,28 @@ export class Transaction implements firestore.Transaction {
...
     const request = {
       database: this._firestore.formattedName,
-      transaction: this._transactionId,
+      transaction: transactionId,
     };
+    this._transactionIdPromise = undefined;
+    this._prevTransactionId = transactionId;
 
     try {
       await this._firestore.request('rollback', request, this._requestTag);
```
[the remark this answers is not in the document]
brettwillis: Yeah I have done so but this PR seems a bit broken at the moment GitHub is stuck processing my last commit for about a day now.

dev/src/document-reader.ts
```diff
@@ -163,7 +197,8 @@ export class DocumentReader<AppModelType, DbModelType extends DocumentData> {
     } catch (error) {
       const shouldRetry =
         // Transactional reads are retried via the transaction runner.
-        !this.transactionId &&
+        !this.transactionIdOrNewTransaction &&
+        !this.retrievedTransactionId &&
```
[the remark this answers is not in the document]
brettwillis: True.

brettwillis:
There we go. With this last commit GitHub seems to have gotten itself unstuck. I've implemented all of those suggestions. You haven't seen the second-to-last commit `Rollback is completed asynchronously`, that was the one that was stuck processing.

brettwillis:
@tom-andersen I ran this branch against a real Firestore backend and oops there were a few gotchas
- Protobuf includes zero/empty "transaction ID" buffers by default
- Sometimes the backend returns the "transaction" in the same response stream element as the "document response" and sometimes it doesn't which surprised me based on the docs. So I made it more resilient to which response comes and when.

Warrants another review sorry.

brettwillis:
> Sometimes the backend returns the "transaction" in the same response stream element as the "document response" and sometimes it doesn't which surprised me based on the docs.

Ah sorry I think I was wrong about this... it was just the fact that protobuf always decodes an empty transaction ID buffer by default.

No harm in what I did to make it more resilient? Or do you want me to revert that?

By the way, will this branch be be run through a real-life test suite before releasing? In both REST and gRPC modes in case that affects the ordering/structuring of the responses?

dev/src/document-reader.ts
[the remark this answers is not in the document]
tom-andersen: Nice. I see you did one better in implementation.

tom-andersen:
> By the way, will this branch be be run through a real-life test suite before releasing? In both REST and gRPC modes in case that affects the ordering/structuring of the responses?

Yes. We have an integration tests that must pass before we merge. I hadn't run them yet, since there were still changes being done.

dev/src/reference.ts
```diff
@@ -3342,53 +3393,93 @@ export class AggregateQuery<
...
     const stack = Error().stack!;
 
     return new Promise((resolve, reject) => {
-      const stream = this._stream(transactionIdOrReadTime);
+      let responseTransaction: Uint8Array | undefined;
+      const stream = this._stream(transactionOrReadTime);
       stream.on('error', err => {
         reject(wrapError(err, stack));
       });
       stream.once('data', result => {
```
tom-andersen, line 3433: Integration tests are fixed by changing this:

```suggestion
      stream.on('data', result => {
```

tom-andersen (changes requested):

dev/src/transaction.ts
tom-andersen: I completely agree that with making this immutable passed through constructor. Whether this should be an `opts` object or have `transactionOrReadTime` and `fieldMask` be parameters, I am inclined to go the parameter route, since that is the pattern used most other places in code. Regardless, what you implemented is a great improvement over previous code IMHO.

tom-andersen:
> Ah sorry I think I was wrong about this... it was just the fact that protobuf always decodes an empty transaction ID buffer by default.
> 
> No harm in what I did to make it more resilient? Or do you want me to revert that?

I am fine with this.

tom-andersen (approved):
LGTM

We are going to hold back on merging until it fits into our release cycle.

Fingers crossed that nothing causes a merge conflict.

brettwillis:
Thanks for all your time and effort reviewing this one @tom-andersen 🙏🏼

brettwillis:
@tom-andersen ok so those recent updates had quite sizeable conflict in `reference.ts`. I've resolved those, and using the `_getResonse()` pattern I introduced afforded quite a sizeable reduction in duplicated code related to "explain" functionality.

Now as a result I discovered something else. The Query response was silently accepting an undefined `readTime`. This was due to a shortfall in type checking. Due to one of my changes it surfaced in typings that `readTime` could be undefined, so I added a guard.

I turns out many tests were incorrectly simulating the "empty query" scenario by not emitting a read time in the response stream. So I made all "empty query" tests always emit a read time in the stream, which properly emulates how the backend should behave.

TLDR; I introduced a behaviour change wherein, if for whatever reason the backend did not emit a read time for query responses, it would previously succeed but the `QuerySnapshot.readTime` would be unexpectedly `undefined` in application code. Now it throws an exception in such case (which should not ever happen because the backend always emits a read time as far as I understand).

Thoughts?

dev/src/reference.ts
```diff
@@ -1725,22 +1775,28 @@ class QueryUtil<
...
-                  DocumentChange<AppModelType, DbModelType>
-                > = [];
-                for (let i = 0; i < docs.length; ++i) {
-                  changes.push(new DocumentChange('added', docs[i], -1, i));
-                }
-                return changes;
+          if (!output.readTime) {
+            reject(wrapError(Error('No read time'), stack));
+            return;
+          }
```
brettwillis, line 1781: @tom-andersen  see here

tom-andersen:
@brettwillis We are now within a window of time where I can merge this.

It is possible to return a query plan without any results. Consequently, there is no read time. This does not align with the types in the type system.

I would like @ehsannas who implemented this feature to take a look.

For now, since this is somewhat unrelated to your changes, I suggest we comment out the check, with an explanation that @ehsannas can follow up upon.

There are integration tests that I run as part of this PR that fail because the query plan does not return a read time.

> @tom-andersen ok so those recent updates had quite sizeable conflict in `reference.ts`. I've resolved those, and using the `_getResonse()` pattern I introduced afforded quite a sizeable reduction in duplicated code related to "explain" functionality.
> 
> Now as a result I discovered something else. The Query response was silently accepting an undefined `readTime`. This was due to a shortfall in type checking. Due to one of my changes it surfaced in typings that `readTime` could be undefined, so I added a guard.
> 
> I turns out many tests were incorrectly simulating the "empty query" scenario by not emitting a read time in the response stream. So I made all "empty query" tests always emit a read time in the stream, which properly emulates how the backend should behave.
> 
> TLDR; I introduced a behaviour change wherein, if for whatever reason the backend did not emit a read time for query responses, it would previously succeed but the `QuerySnapshot.readTime` would be unexpectedly `undefined` in application code. Now it throws an exception in such case (which should not ever happen because the backend always emits a read time as far as I understand).
> 
> Thoughts?

dev/src/reference.ts
tom-andersen: Comment this out for now, with explanation that 'ExplainOptions' with 'analyze = false` will not have read time.

This is because no results are results are returned in this situation.

brettwillis:
Ok I reverted back to null assertion instead of throwing an error. This is back to the behaviour it has always been. And I added to-dos for @ehsannas to follow up in a separate PR.

dev/src/reference.ts
```diff
@@ -2982,99 +3048,62 @@ export class Query<
...
-            if (docs === null) {
-              docs = [];
-            }
-            docs.push(data.document);
-          }
-          if (data.explainMetrics) {
-            metrics = data.explainMetrics;
-
-            if (docs === null && metrics?.executionStats !== null) {
-              // This indicates that the query was executed, but no documents
```
tom-andersen, line 3019: This logic wasn't moved over. 

The effect is that returned snapshot will be null if there are no docs and `explainMetrics !== null` and `explainMetrics.executionStats === null`.

When `options.analyze !== true`, then only the `planSummary` is returned, but no execution actually happens. That means there will be no results, and no `executionStats`.

This is why the type `ExplainResults` has `readonly snapshot: T | null`

dev/src/reference.ts
```diff
@@ -1725,22 +1775,33 @@ class QueryUtil<
...
+          // TODO(ehsannas): Typings of QuerySnapshot suggest that readTime is always available
+          // however there are in the scenario of query plan with no results there will be no read time sent
+          // so the typings should be updated to make readTime optional?
+          // For now we keep the null assertion behaviour which has been the case historically
+          // if (!output.readTime) {
+          //   reject(wrapError(Error('No read time'), stack));
+          //   return;
+          // }
+
+          const result = query._createSnapshot(
```
tom-andersen, line 1787: `result` needs to be `null` when `options.analyze !== true`

The backend should only return `explainMetrics` in this case.

However, if `options.analyze === true`, then returning no docs is an empty result, and we should return a snapshot with no docs.

dev/src/reference.ts
```diff
@@ -1725,22 +1775,33 @@ class QueryUtil<
...
-              () => docs,
-              () => {
-                const changes: Array<
-                  DocumentChange<AppModelType, DbModelType>
-                > = [];
-                for (let i = 0; i < docs.length; ++i) {
-                  changes.push(new DocumentChange('added', docs[i], -1, i));
-                }
-                return changes;
+          // TODO(ehsannas): Typings of QuerySnapshot suggest that readTime is always available
```
tom-andersen, line 1778: I talked to @ehsannas and read through some documentation.

```
// If the query returns no results, a response with `read_time` and no
// `document` will be sent, and this represents the time at which the query
// was run.
```
 
 Note, `readTime` represents the time at which query was run. So, when calling `explain(options)` with `options.analyze !== true`, indicating NO execution is to be done, ONLY a plan is to be returned, then `readTime` will be absent, but we also omit the snapshot from result.

dev/src/reference.ts
```diff
@@ -1458,6 +1458,33 @@ export class VectorQuerySnapshot<
   }
 }
 
+interface QueryStreamResponse<
+  AppModelType = firestore.DocumentData,
+  DbModelType extends firestore.DocumentData = firestore.DocumentData,
+> {
+  transaction?: Uint8Array;
+  readTime?: Timestamp;
```
tom-andersen, line 1466: I don't think `readTime` ever exists outside the context of a snapshot. The snapshot should have a `readTime` property. So, we can remove from this object IMHO.

dev/src/reference.ts
```diff
@@ -1458,6 +1458,33 @@ export class VectorQuerySnapshot<
...
+  explainMetrics?: ExplainMetrics;
+  document?: QueryDocumentSnapshot<AppModelType, DbModelType>;
+}
+
+interface QueryResponse<TSnapshot> {
+  transaction?: Uint8Array;
+  readTime?: Timestamp;
```
tom-andersen, line 1473: Likewise, I think we can remove `readTime`, and only have it exist within the snapshot.

dev/src/reference.ts
```diff
@@ -1458,6 +1458,33 @@ export class VectorQuerySnapshot<
...
+  explainMetrics?: ExplainMetrics;
+  result: TSnapshot;
```
tom-andersen, line 1475: The `result` snapshot is optional. If running `explain(..)` without execution, it should be null. However, it is also possible to run `explain(..)` with execution results.

dev/src/reference.ts
```diff
@@ -2982,99 +3048,62 @@ export class Query<
...
+    const {result, explainMetrics} = await this._getResponse(
+      undefined,
+      options
+    );
+    if (!explainMetrics) {
+      throw new Error('No explain results');
+    }
+    return new ExplainResults(
+      explainMetrics,
+      result as QuerySnapshot<AppModelType, DbModelType>
```
tom-andersen, line 3066: `result` is `null` if `analyze !== true`

tom-andersen (changes requested):
I spoke to @ehsannas.

We should fix this up before merging and remove comment.

I have added comments to help guide change. Basically, if there is a `snapshot`, there will also be a `readTime`. However, we do not always want to produce a `snapshot` if `explain(..)` only returns a plan, without doing any execution.

Absence of documents is NOT the same as an empty result. It can also mean that NO execution was done as part of `explain`. The logic needs to respect this distinction and only return a snapshot for normal queries without explain, or explain query with `option.analyze === true`.

This does present a problem for the type system, since we are plucking values from a stream. So some coercion will be required.

I really appreciate your effort on this PR, and I genuinely feel bad that your work conflicted with two other features we were releasing.

dev/src/reference.ts
brettwillis: This `QueryStreamResponse` represents the shape of any single "element" emitted from the stream. It is possible to the stream to emit only a `readTime` and no documents (empty result). So I think keeping `readTime` in this particular place is valid.

dev/src/reference.ts
brettwillis: Ok I agree, your comment from above does apply in this `QueryResponse` type. Will do.

dev/src/reference.ts
brettwillis: Maybe I was mistaken, but this seemed like a just a long-winded way of achieving what I had instead which was just to initialise `docs` with an empty array to begin with.

I'm about to push a commit which changes the logic to only create a query snapshot if the stream received a `readTime` (and `docs` is still initialised to an empty array). So if the backend sends a readTime, when we get out a query snapshot with the received docs (or empty docs). If no read time received then no snapshot is created, and that should only happen with explain && analyze !== true (but is the prerogative of the backend rather than validating it in the lower levels of the SDK).

I then delegated further up the chain in the `_get()` function to verify that there is indeed a query snapshot present.

I also rationalised the `_get()`, `_getResponse()` and `_getResponseOrExplain()` methods:
- the internal `_get()` now returns response with metadata (and verifies that query snapshot exists). Previously this was just a trivial wrapper underneath the public `get()` of seemingly no use
- `_getReponse()` is the same but without the check for existence of a snapshot, and this is reused by `explain()`
- `_getResponseOrExplain()` removed
- Renamed `getResponse()` on `DocumentReader` to `_get()` to align with this convention

dev/src/reference.ts
tom-andersen: 👍 I stand corrected!

dev/src/reference.ts
```diff
@@ -4179,52 +4231,21 @@ export class AggregateQuery<
...
   ): Promise<
     ExplainResults<
       AggregateQuerySnapshot<AggregateSpecType, AppModelType, DbModelType>
     >
   > {
-    if (options === undefined) {
-      options = {};
+    const {result, explainMetrics} = await this._getResponse(
+      undefined,
+      options
```
tom-andersen, line 4243: We need to pass an object to invoke explain.

```suggestion
      options || {}
```

tom-andersen (changes requested):

tom-andersen (approved):
I made one fix to your logic in order to pass System Tests. LGTM.

dev/src/document-reader.ts
```diff
@@ -142,28 +178,31 @@ export class DocumentReader<AppModelType, DbModelType extends DocumentData> {
             response.found,
             response.readTime!
           );
-        } else {
+        } else if (response.missing) {
           logger(
             'DocumentReader.fetchDocuments',
             requestTag,
             'Document missing: %s',
             response.missing!
```
tom-andersen, line 186:
```suggestion
            response.missing
```

dev/src/transaction.ts
```diff
@@ -622,17 +619,123 @@ export class Transaction implements firestore.Transaction {
...
+    ) => Promise<{transaction?: Uint8Array; result: TResult}>
+  ): Promise<TResult> {
+    if (this._transactionIdPromise) {
+      // Simply queue this subsequent read operation after the first read
+      // operation has resolved and we don't expect a transaction ID in the
+      // response because we are not starting a new transaction
+      return this._transactionIdPromise.then(async opts => {
+        const r = await resultFn.call(this, param, opts);
+        return r.result;
+      });
```
ehsannas, line 641: nit: to make it more consistent with the rest of this function

```js
      return this._transactionIdPromise
                 .then(opts => resultFn.call(this, param, opts))
                 .then(r => r.result);
```

dev/src/reference.ts
ehsannas: SGTM

dev/test/transaction.ts
```diff
@@ -500,25 +558,21 @@ describe('failed transactions', () => {
         await runTransaction(
           /* transactionOptions= */ {},
           transactionFunction,
-          begin({transactionId: 'foo1'}),
-          query('foo1', serverError),
-          rollback('foo1'),
+          query({newTransaction: {readWrite: {}}, error: serverError}),
+          // No rollback because the lazy-start operation failed
+          //rollback('foo1'),
```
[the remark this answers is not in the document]
tom-andersen: There is no transaction id yet, since that is part of the result from first query. So we can't rollback, even if we wanted to.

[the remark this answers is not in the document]
ehsannas: yup. I'm just saying we should remove the commented code.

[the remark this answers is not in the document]
tom-andersen: Ahh. Good point! +1

MarkDuckworth (commented):
This is fantastic. There are many nice code improvements, but it's also a significant community contribution. Thank you for your efforts.

I have a few additional comments for your consideration. But otherwise I think it looks good.

The only potential adverse affect would be if a customer is implementing parallel reads in a transaction using `get()`, then this has a possibility to slow the transaction if the first read is slow/large-doc. But in many cases that could be mitigated with a call to `getAll()`.

dev/src/transaction.ts
```diff
@@ -449,85 +447,87 @@ export class Transaction implements firestore.Transaction {
...
+    const request: api.IRollbackRequest = {
+      database: this._firestore.formattedName,
+      transaction: transactionId,
+    };
+    this._transactionIdPromise = undefined;
+    this._prevTransactionId = transactionId;
+
+    // Rollback request is completed asynchronously
+    // We don't need to wait for it to completed before continuing the next attempt
+    this._firestore
```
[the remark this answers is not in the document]
ehsannas: good point. Yes, I think it's safer to change this to the old behavior. In particular, this affects cases where the customer code is running inside a Cloud Function environment. The Cloud Function could terminate and the response for this rpc may arrive after the Cloud Function termination, which results in errors or warnings in the logs.

ehsannas (approved):
Thanks for your contribution @brettwillis . LGTM -- only the remaining minor comments should be addressed.

dev/src/transaction.ts
```diff
@@ -290,7 +288,7 @@ export class Transaction implements firestore.Transaction {
     documentRef: firestore.DocumentReference<AppModelType, DbModelType>,
     data: firestore.WithFieldValue<AppModelType>
   ): Transaction {
-    if (this._readOnly) {
+    if (!this._writeBatch) {
```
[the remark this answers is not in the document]
brettwillis: @MarkDuckworth I have changed the logic a little, the `_readOnly` field no longer exists, with the following motivation
- `_writeBatch` is now initialised only when it is read-write to avoid initialising a write batch for readonly transactions (same for `_backoff`)
- then the `_readOnly` flag would be superfluous as the presence of `_writeBatch` is now the authoritative flag for whether it is readonly or read/write (so I removed `_readOnly`)
- and `_readOnlyReadTime` field is optional, just because a read time is not set doesn't mean it isn't readonly

dev/src/transaction.ts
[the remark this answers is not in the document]
brettwillis: Ok, will revert

dev/src/transaction.ts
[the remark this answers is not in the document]
MarkDuckworth: Thanks for the detailed reply. Sounds good then.

dev/src/transaction.ts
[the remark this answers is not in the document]
ehsannas: 👍  That sounds fine to me.

MarkDuckworth (approved):

PR #2028, opened 2024-03-22
Markduckworth/qp fn

MarkDuckworth:
Merges find nearest and query profiling.
- Refactors Query.explain and Query.explainStream to use QueryUtil._stream
- Refactors QueryUtil._stream to support query profiling messages
- Does not refactor AggregateQuery.explain to use QueryUtil._stream
- Does not implement VectorQuery.explain or VectorQuery.explainStream


ehsannas (commented):
LGTM

ehsannas (approved):

PR #2037, opened 2024-04-29
chore: refactor reference.ts to one class per file

MarkDuckworth:


MarkDuckworth:
Wait to merge until https://github.com/googleapis/nodejs-firestore/pull/2017 merges, so we don't force the community contribution to refactor.

dev/src/reference/types.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @callback documentSnapshotCallback
+ * @param {DocumentSnapshot} snapshot A document snapshot.
+ */
+
+/**
+ * onSnapshot() callback that receives an error.
+ *
+ * @callback errorCallback
+ * @param {Error} err An error from a listen.
+ */
```
ehsannas, line 81: I see these in the original `reference.ts`, but it looks like we should remove them.

ehsannas (approved):

dev/src/reference/types.ts
MarkDuckworth: These are being used by JSDoc to build the documentation at https://googleapis.dev/nodejs/firestore/latest/global.html#querySnapshotCallback. If we migrate the SDK to only use TSDoc, then we can remove these.

PR #2085, opened 2024-07-15
refactor: tracing using OpenTelemetry.

ehsannas:


.idea/runConfigurations/System_Test.xml
jimit-j-shah, line 1: Is this an intended change? Seems like an Intellij artifact.

ehsannas: yes, it's an intended change. It allows us to run integration tests from the intellij UI. This PR defines a brand-new integration test file for tracing, so needed to update this.

dev/src/collection-group.ts
```diff
@@ -81,48 +82,53 @@ export class CollectionGroup<
...
-      'Firestore.getPartitions',
-      tag,
-      'Received %d partitions',
-      partitions.length
     );
 
-    // Sort the partitions as they may not be ordered if responses are paged.
-    partitions.sort((l, r) => compareArrays(l, r));
-
     for (let i = 0; i < partitions.length; ++i) {
```
milaGGL, line 132: why the code from 132-149 is not wrapped into the `startActiveSpan` cb function?

dev/src/index.ts
```diff
@@ -1811,6 +1875,18 @@ export class Firestore implements firestore.Firestore {
                 'Received response: %j',
                 chunk
               );
+              numResponses++;
+              if (numResponses === 1) {
+                this._traceUtil
+                  .currentSpan()
+                  .addEvent(`Firestore.${methodName}: First response received`);
+              } else if (numResponses % NUM_RESPONSES_PER_TRACE_EVENT === 0) {
```
milaGGL, line 1883: How does tracing records snapshot listeners? Here it could capture the "first response" event, but could it get an "end" event? is it going to keep recording till the snapshot listeners called unsubscribe function?

dev/system-test/tracing.ts
```diff
@@ -0,0 +1,656 @@
...
+    // in order to avoid duplicate global tracers.
+    trace.disable();
+    context.disable();
+
+    // Set up a context manager.
+    const contextManager = new AsyncLocalStorageContextManager();
+    contextManager.enable();
+    context.setGlobalContextManager(contextManager);
+
+    console.log('in beforeEachTest');
```
milaGGL, line 150: Maybe remove this log

dev/system-test/tracing.ts
```diff
@@ -0,0 +1,656 @@
...
+  // backend and making sure it has recorded all the expected spans. These
+  // tests have `e2e` set to `true`.
+  e2e: boolean;
+
+  // OpenTelemetry allows registering a specific OpenTelemetry instance as the
+  // "global" instance. We should make sure that the tracing works in two
+  // circumstances: (1) the user provides us with a specific OpenTelemetry
+  // instance. (2) the user does not provide an instance and we find the
+  // global instance to use.
+  globalOpenTelemetry: boolean;
```
milaGGL, line 96: `useGlobalOpenTelemetry` or `hasGlobalOpenTelemetry` since it is boolean

dev/system-test/tracing.ts
```diff
@@ -0,0 +1,656 @@
...
+    if (config.globalOpenTelemetry) {
+      trace.setGlobalTracerProvider(tracerProvider);
+    }
+
+    const settings: Settings = {
+      preferRest: config.preferRest,
+      openTelemetryOptions: getOpenTelemetryOptions(config, tracerProvider),
+    };
+    if (process.env.FIRESTORE_NAMED_DATABASE) {
+      settings.databaseId = process.env.FIRESTORE_NAMED_DATABASE;
```
milaGGL, line 183: could this "settings.databaseId" be overwritten  by the if statement in line 188? 
knit: chain the 2 if statements together for readability.

dev/system-test/tracing.ts
```diff
@@ -0,0 +1,656 @@
...
+    if (!settings.databaseId && process.env.DATABASE_ID) {
+      settings.databaseId = process.env.DATABASE_ID;
+    }
+
+    firestore = new Firestore(settings);
+  }
+
+  // Returns true on success, and false otherwise.
+  async function waitForCompletedInMemorySpans(): Promise<boolean> {
+    await tracerProvider.forceFlush();
```
milaGGL, line 197: does `forceFlush` return false if it fails?

dev/src/collection-group.ts
ehsannas: this function (`getPartitions`) is an `async*` which generates multiple results asynchronously (`yield`s results) (line 133 and 143). This is not supported in `startActiveSpan` right now, but it's not a concern given that this function is simply yielding a new object in each loop iteration without doing any expensive work (for example, no RPCs or i/o or anything). All of the real work (`initializeIfNeeded`, `requestStream`, etc) is already done above inside the span. So the current span captures the time near-perfectly.

In fact I think this function could have just returned a `Promise<QueryPartition<AppModelType, DbModelType>>` rather than `AsyncIterable<QueryPartition<AppModelType, DbModelType>>`

dev/src/index.ts
ehsannas: This is a great question. This function (`requestStream`) is called from 2 kinds of functions:

1- For server requests that require streaming RPCs. For example: when running `query.get()`. In these cases, the SDK creates a new span in the higher-level API function ([example](https://github.com/googleapis/nodejs-firestore/blob/4d9d4cbb3eab461dfc2250197a4ff9c198e6d3cc/dev/src/reference/query.ts#L1081)). So when `requestStream` is called, an active span exists already. Notice that inside `requestStream` we don't create a new span; we just say "get me the current span" (`.currentSpan()`) and add events to the current span. In these cases the SDK initiates the stream, sends the request, and when it receives the `end` event, it shuts down the stream. And then closes the _span_ in the higher level API function.

2- For snapshot listeners ([here](https://github.com/googleapis/nodejs-firestore/blob/f73e28b69be3ff9f870055af80683f12048549e8/dev/src/watch.ts#L484C12-L484C25)). In this case the higher level function that is calling `requestStream` is _not_ creating a span. So when `.currentSpan()` is called on line 1881, it calls [this](https://github.com/googleapis/nodejs-firestore/blob/4d9d4cbb3eab461dfc2250197a4ff9c198e6d3cc/dev/src/telemetry/enabled-trace-util.ts#L101), which calls `trace.getActiveSpan()` which will return `undefined`, so EnabledTraceUtil will create `new Span(undefined);`, which is a [valid](https://github.com/googleapis/nodejs-firestore/blob/4d9d4cbb3eab461dfc2250197a4ff9c198e6d3cc/dev/src/telemetry/span.ts#L21) usecase (note that the constructor parameter is optional). Then [`addEvent`](https://github.com/googleapis/nodejs-firestore/blob/4d9d4cbb3eab461dfc2250197a4ff9c198e6d3cc/dev/src/telemetry/span.ts#L28) will not do anything (it's a no-op) because there's no underlying otel span. So, for snapshot listeners, we do not create any spans and we do not add any events to any spans. This is intentional and consistent with the java implementation. Snapshot listeners are long-lived streams which are not the right fit for tracing; although we may define specific short-lived traces that raise visibility to some aspects of a snapshot listener in the future.

dev/system-test/tracing.ts
ehsannas: Done

dev/system-test/tracing.ts
ehsannas: oops. Done.

dev/system-test/tracing.ts
ehsannas: no, it's a `Promise<void>`. If the flush is not successful, we likely won't get the correct number of spans and the test will fail.

dev/system-test/tracing.ts
ehsannas: no, if line 183 is executed, then `!settings.databaseId` will be false and line 189 will not get executed.

Done.

dev/src/telemetry/trace-util.ts
```diff
@@ -0,0 +1,75 @@
...
+  | number
+  | boolean
+  | Array<string>
+  | Array<number>
+  | Array<boolean>;
+
+/**
+ * Span names for instrumented operations.
+ */
+export const SERVICE = 'google.firestore.v1.Firestore/';
```
milaGGL, line 33: No `SPAN_NAME_RUN_QUERY`, `SPAN_NAME_RUN_AGGREGATION_QUERY`,`SPAN_NAME_TRANSACTION_BEGIN` in node?

dev/src/reference/query-util.ts
```diff
@@ -254,10 +258,18 @@ export class QueryUtil<
...
 
+        let isRetryRequestWithCursor = false;
         let streamActive: Deferred<boolean>;
         do {
           streamActive = new Deferred<boolean>();
           const methodName = 'runQuery';
+
+          this._firestore._traceUtil.currentSpan().addEvent(methodName, {
+            [ATTRIBUTE_KEY_IS_TRANSACTIONAL]:
+              request.transaction !== null && request.transaction !== undefined,
```
milaGGL, line 269: can this be replaced by `!!request.transaction`?

dev/src/reference/document-reference.ts
```diff
@@ -431,14 +464,19 @@ export class DocumentReference<
       unknown | string | firestore.FieldPath | firestore.Precondition
     >
   ): Promise<WriteResult> {
-    // eslint-disable-next-line prefer-rest-params
-    validateMinNumberOfArguments('DocumentReference.update', arguments, 1);
+    return this._firestore._traceUtil.startActiveSpan(
+      SPAN_NAME_DOC_REF_UPDATE,
+      () => {
+        // eslint-disable-next-line prefer-rest-params
+        validateMinNumberOfArguments('DocumentReference.update', arguments, 1);
```
milaGGL, line 471: replace with `SPAN_NAME_DOC_REF_UPDATE`

dev/src/reference/query-util.ts
ehsannas: yes, good one. Done.

dev/src/telemetry/trace-util.ts
ehsannas: The Node SDK has been updated such that it never needs to send a BeginTransaction rpc to the server. (see https://github.com/googleapis/nodejs-firestore/pull/2017).

Regarding using the span name as method name, I'm hesitant to do it. For example we'd have something like:

```js
       const methodName = SPAN_NAME_RUN_AGGREGATION_QUERY;
        const backendStream = await firestore.requestStream(
          methodName,
          /* bidirectional= */ false,
          request,
          tag
        );
```
because the person reading the code may be confused why we're passing a tracing span name to the requestStream function. Or you may wonder if it's really needed if tracing is OFF. This is in fact the rpc method [name](https://github.com/googleapis/nodejs-firestore/blob/f73e28b69be3ff9f870055af80683f12048549e8/dev/src/types.ts#L98). There are minor differences between Java and Node in this aspect. For example, in node both query and aggregate query call the same underlying "requestStream" function, but in Java there are separate classes for handling query vs aggregate query.

milaGGL: I agree. the above code should not use span constant as it is not tracing related. But, we should use it to creating span. like the code change here:
Screenshot 2024-07-24 at 10 06 20 AM

I prefer not to extract the string to `methodName`, and use the `
SPAN_NAME_RUN_AGGREGATION_QUERY` inside the `addEvent`. 
Even tho the string used are the same, but they are serving the different purposes.

ehsannas: Oh I see what you mean.

ehsannas: Done

milaGGL (approved):
LGTM

PR #2090, opened 2024-07-26
feat: Return computed distance and set distance thresholds on VectorQueries

MarkDuckworth:
Return computed distance and set distance thresholds on VectorQueries


MarkDuckworth:
Needs DTS updates

wu-hui (approved):

wu-hui (approved):
LGTM. you have some `it.only` to remove I think.

wu-hui (approved):
Let's go!

PR #2093, opened 2024-07-29
refactor: e2e tests using cloud trace v1 api.

ehsannas:
Add end-to-end tests. Each end-to-end test enables tracing, performs one or more Firestore operations, and then queries the Cloud Trace backend to see if the expected trace and spans have been generated.


package.json
```diff
@@ -34,14 +34,14 @@
...
     "docs": "jsdoc -c .jsdoc.js",
-    "system-test:rest": "FIRESTORE_PREFER_REST=true mocha build/system-test --timeout 600000",
-    "system-test:named-db:rest": "FIRESTORE_NAMED_DATABASE=test-db FIRESTORE_PREFER_REST=true mocha build/system-test --timeout 600000",
-    "system-test:grpc": "mocha build/system-test --timeout 600000",
-    "system-test:named-db:grpc": "FIRESTORE_NAMED_DATABASE=test-db mocha build/system-test --timeout 600000",
-    "system-test:emulator:rest": "FIRESTORE_EMULATOR_HOST=localhost:8080 FIRESTORE_PREFER_REST=true mocha build/system-test --timeout 600000",
-    "system-test:named-db:emulator:rest": "FIRESTORE_NAMED_DATABASE=test-db FIRESTORE_EMULATOR_HOST=localhost:8080 FIRESTORE_PREFER_REST=true mocha build/system-test --timeout 600000",
-    "system-test:emulator:grpc": "FIRESTORE_EMULATOR_HOST=localhost:8080 mocha build/system-test --timeout 600000",
-    "system-test:named-db:emulator:grpc": "FIRESTORE_NAMED_DATABASE=test-db FIRESTORE_EMULATOR_HOST=localhost:8080 mocha build/system-test --timeout 600000",
+    "system-test:rest": "FIRESTORE_PREFER_REST=true mocha build/system-test --timeout 1200000",
```
milaGGL, line 37: why the timeout is extended? would this slow the full cycle down?

dev/system-test/tracing.ts
```diff
@@ -14,15 +14,20 @@
 
 import * as chaiAsPromised from 'chai-as-promised';
```
milaGGL, line 15: please add PR description of what has been done.

dev/system-test/tracing.ts
```diff
@@ -128,14 +198,49 @@ describe('Tracing Tests', () => {
...
+    return hexString;
+  }
+
+  function getNewSpanContext(): SpanContext {
+    const spanContext: SpanContext = {
+      traceId: generateRandomHexString(NUM_TRACE_ID_BYTES),
+      spanId: generateRandomHexString(NUM_SPAN_ID_BYTES),
+      traceFlags: TraceFlags.SAMPLED,
+    };
+    console.log('custom span context=');
```
milaGGL, line 229: Knit: remove the logs.

milaGGL:
it does not allow me to add comment dt non-diff lines, so here is a question: 
Screenshot 2024-08-02 at 3 05 20 PM

dev/system-test/tracing.ts
```diff
@@ -402,218 +611,235 @@ describe('Tracing Tests', () => {
...
-          useGlobalOpenTelemetry: false,
-          preferRest: false,
-        };
-        beforeEach(async () => beforeEachTest(config));
-        runTestCases(config);
+  describe(IN_MEMORY_TEST_SUITE_TITLE, () => {
+    describe(NON_GLOBAL_OTEL_TEST_SUITE_TITLE, () => {
+      describe(GRPC_TEST_SUITE_TITLE, () => {
+        beforeEach(function () {
+          beforeEachTest(this.currentTest!);
```
milaGGL, line 618: nice code reduction!

dev/system-test/tracing.ts
```diff
@@ -402,218 +611,235 @@ describe('Tracing Tests', () => {
...
-      await firestore.collection('foo').doc().create({});
-
-      await waitForCompletedSpans(config, 2);
-      expectSpanHierarchy(SPAN_NAME_DOC_REF_CREATE, SPAN_NAME_BATCH_COMMIT);
+      await runFirestoreOperationInRootSpan(() =>
+        firestore.collection('foo').doc().create({})
+      );
+      await waitForCompletedSpans(3);
+      expectSpanHierarchy(
+        SPAN_NAME_TEST_ROOT,
```
milaGGL, line 703: is the added span specific to testing or the tracing added root layer span generally? What is the reason behind wrapping the operations with `runFirestoreOperationInRootSpan`and adding this extra root span?

dev/system-test/tracing.ts
ehsannas: Oops. Done.

dev/system-test/tracing.ts
ehsannas: Great question. Many tracing tests are run in a short period of time, and each test sends one trace and potentially several spans to the same tracing backend (cloud trace).

We want each test to be able to query Cloud Trace to find out if the expected spans were reported for that test. In order to distinguish between the different tests, we need to do two things
1- create a custom context (which contains a unique Trace ID) for each test (done in `getNewSpanContext`)
2- create a root span _with that context_ so that all Firestore spans for that test fall under this root span.
3- query the cloud trace backend for that unique Trace ID.

doing (1) without (2) doesn't work.

package.json
ehsannas: yeah it'll slow down the start-to-finish time of the "integration tests CI". This PR adds 120 tests and the tests take a while because each test does at the very least one round trip to Firestore server, and at least 2-3 round-trips to cloud trace server (because the traces don't become available in the cloud trace server immediately, so we need to retry).

I am open to suggestions about how to reduce this cost. Maybe we should run the in-memory tests on PR CI and cloud-trace backend tests on a nightly basis.

dev/system-test/tracing.ts
ehsannas: Done.

ehsannas:
Regarding your question:

For the time being, the reason this interface is here and not in [`firestore.d.ts`](https://github.com/googleapis/nodejs-firestore/blob/main/types/firestore.d.ts) is that we don't want to make the API public yet. So there's currently no public API in the `Settings` interface for setting the otel options. However, Settings [can accept _any_ options](https://github.com/googleapis/nodejs-firestore/blob/main/types/firestore.d.ts#L463). So we'll use that for now.

Once we release this feature, we will move the `export interface FirestoreOpenTelemetryOptions` definition out of here and put it in firestore.d.ts; and we will add a `openTelemetryOptions` field to the `Settings` interface.

Java does have [`FirestoreOpenTelemetryOptions`](https://github.com/googleapis/java-firestore/blob/4384970a7005b1504e71eb1dab93eaefa0cc123c/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java#L29-L31) and a way to [set it in the FirestoreOptions](https://github.com/googleapis/java-firestore/blob/4384970a7005b1504e71eb1dab93eaefa0cc123c/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOptions.java#L228).

Note that there is an inherent difference between OTEL in java and node.js. In Java the Firestore SDK accepts an instance `OpenTelemetry`, but in Node.js it's modularized, so we need to accept an instance of a `TracerProvider`.

package.json
milaGGL: if only 10 minutes are added on top to run the whole system test, I think we could live with that comparing to having some part of the code not tested in PRs.

milaGGL (approved):
LGTM

PR #2094, opened 2024-07-30
[feat] ppl

wu-hui:


dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+      results,
+      {title: "The Hitchhiker's Guide to the Galaxy"},
+      {title: 'Pride and Prejudice'}
+    );
+  });
+
+  it('arrayContainsAll works', async () => {
+    const results = await randomCol
+      .pipeline()
+      .where(Field.of('tags').arrayContainsAll('adventure', 'magic'))
```
MarkDuckworth, line 397: I wish this read `.where(field('tags').arrayContains...` or even `.where(Field.named('tags').arrayContains...`. As I read through the code, I continue to read it as "where field of something" but the meaning is "where field something".

ehsannas: I think for Java Andy added an function overloads that covers the "common" usecase -- which is the case that the first parameter is a string specifying the name of the field. Is that possible to do for TS?

dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+      .limit(1)
+      .execute();
+    expectResults(results, {ratingIsNull: false, ratingIsNotNaN: true});
+  });
+
+  it('testMapGet', async () => {
+    const results = await randomCol
+      .pipeline()
+      .select(
+        Field.of('awards').mapGet('hugo').as('hugoAward'),
```
MarkDuckworth, line 661: I'd also consider `Field.of('awards.hugo').as('hugoAward')`, or is there some special behavior to `mapGet`?

Does `mapGet()` support dot notation? 
`Field.of('awards').mapGet('hugo.year').as('hugoAwardYear')`
vs
`Field.of('awards').mapGet('hugo').mapGet('year').as('hugoAwardYear')`

dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+  //   expectResults(results, {collectionId: 'books'});
+  // });
+
+  it('testDistanceFunctions', async () => {
+    const sourceVector = [0.1, 0.1];
+    const targetVector = [0.5, 0.8];
+    const results = await randomCol
+      .pipeline()
+      .select(
+        cosineDistance(Constant.ofVector(sourceVector), targetVector).as(
```
MarkDuckworth, line 701: Consider just `Constant.vector(sourceVector)`

dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+          'cosineDistance'
+        ),
+        dotProductDistance(Constant.ofVector(sourceVector), targetVector).as(
+          'dotProductDistance'
+        ),
+        euclideanDistance(Constant.ofVector(sourceVector), targetVector).as(
+          'euclideanDistance'
+        )
```
MarkDuckworth, line 709: This pattern will support KNN. If eventually there is ANN support we may need a different pattern. Do we need to consider that now?

dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+      cosineDistance: 0.02560880430538015,
+      dotProductDistance: 0.13,
+      euclideanDistance: 0.806225774829855,
+    });
+  });
+
+  it('testNestedFields', async () => {
+    const results = await randomCol
+      .pipeline()
+      .where(eq('awards.hugo', true))
```
MarkDuckworth, line 724: Cool. Does this make `mapGet` redundant?

MarkDuckworth (commented):
Just starting review. Will resume next week, but providing some thoughts now.

dev/src/stage.ts
```diff
@@ -0,0 +1,245 @@
...
+      args: this.orders.map(o => o._toProto(serializer)),
+      options: {
+        density: serializer.encodeValue(this.density)!,
+        truncation: serializer.encodeValue(this.truncation)!,
+      },
+    };
+  }
+}
+
+export class GenerateStage implements Stage {
```
MarkDuckworth, line 236: Should this be `GenericStage` to match the API method `Pipeline.genericStage()`

dev/src/pipeline.ts
```diff
@@ -0,0 +1,612 @@
...
+  /**
+   * @internal
+   * @private
+   */
+  withStartCursor(arg0: QueryCursor): PaginatingPipeline {
+    throw new Error('Method not implemented.');
+  }
+}
+
+export class PipelineResult<
```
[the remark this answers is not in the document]
MarkDuckworth: The RPC API appears to call these `Document`s

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+  gte(other: any): Gte {
+    if (other instanceof Expr) {
+      return new Gte(this, other);
+    }
+    return new Gte(this, Constant.of(other));
+  }
+
+  arrayConcat(...values: Expr[]): ArrayConcat;
+  arrayConcat(...values: any[]): ArrayConcat;
+  arrayConcat(...values: any[]): ArrayConcat {
```
MarkDuckworth, line 132: Consider making values a standard array parameter, not rest parameters.

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+      return new CosineDistance(this, other);
+    } else {
+      return new CosineDistance(this, Constant.ofVector(other));
+    }
+  }
+
+  dotProductDistance(other: Expr): DotProductDistance;
+  dotProductDistance(other: VectorValue): DotProductDistance;
+  dotProductDistance(other: number[]): DotProductDistance;
+  dotProductDistance(other: Expr | VectorValue | number[]): DotProductDistance {
```
MarkDuckworth, line 303: Might be more accurate to just call this this and the class `dotProduct`. Drop "Distance"

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+  static of(value: GeoPoint): Constant;
+  static of(value: Timestamp): Constant;
+  static of(value: Date): Constant;
+  static of(value: Uint8Array): Constant;
+  static of(value: DocumentReference): Constant;
+  static of(value: api.IValue): Constant;
+  static of(value: Array<any>): Constant;
+  static of(value: Map<string, any>): Constant;
+  static of(value: VectorValue): Constant;
+  static of(value: api.IValue): Constant;
```
MarkDuckworth, line 467: If you want to make this more extensible, you could also add an interface

```
interface FirestoreProtoSerializable {
  toProto(): api.IValue;
}
```

and have a `static of(value: FirestoreProtoSerializable): Constant`

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+class CosineDistance extends Function {
+  constructor(
+    private vector1: Expr,
+    private vector2: Expr
+  ) {
+    super('cosine_distance', [vector1, vector2]);
+  }
+}
+
+class DotProductDistance extends Function {
```
MarkDuckworth, line 892: Same. Just "DotProduct" is probably more accurate.

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+
+  private constructor(private fields: Field[]) {
+    super();
+  }
+
+  static of(name: string, ...others: string[]): Fields {
+    return new Fields([Field.of(name), ...others.map(Field.of)]);
+  }
+
+  static ofAll(): Fields {
```
MarkDuckworth, line 429: Is this used?

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+  constructor(
+    private array: Expr,
+    private values: Expr[]
+  ) {
+    super('array_contains_all', [array, new ListOfExprs(values)]);
+  }
+  filterable = true as const;
+}
+
+class ArrayContainsAny extends Function implements FilterCondition {
```
MarkDuckworth, line 633: Is FilterCondition always a function that returns a boolean? If so, would it be useful to make a more generic BooleanFunction interface in case there is a future use case for boolean functions outside of filtering?

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+    super('euclidean_distance', [vector1, vector2]);
+  }
+}
+
+export function add(left: Expr, right: Expr): Add;
+export function add(left: Expr, right: any): Add;
+export function add(left: string, right: Expr): Add;
+export function add(left: string, right: any): Add;
+export function add(left: Expr | string, right: Expr | any): Add {
+  const normalizedLeft = typeof left === 'string' ? Field.of(left) : left;
```
MarkDuckworth, line 915: Great candidate for a helper `Field.fromArgument(left)`. Similar to `FieldPath.fromArgument(stringOrFieldPath)`

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+  const normalizedRight = right instanceof Expr ? right : Constant.of(right);
```
MarkDuckworth, line 916: `Expr.fromExpressionOrConstant(right)`

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+}
+
+export function arrayElement(): ArrayElement {
+  return new ArrayElement();
+}
+
+export function inAny(element: Expr, others: Expr[]): In;
+export function inAny(element: Expr, others: any[]): In;
+export function inAny(element: string, others: Expr[]): In; // Added overload
+export function inAny(element: string, others: any[]): In; // Added overload
```
MarkDuckworth, line 1102: If this behavior is different from the classic `in()` supporting array, should we also change `Expr.in(...)` to name `Expr.inAny(...)`. Or if this is only supporting `in` arrays, would it make more sense to name it `inArray(...)`?

dev/src/expression.ts
```diff
@@ -0,0 +1,1425 @@
...
+  const exprOthers = others.map(other =>
+    other instanceof Expr ? other : Constant.of(other)
+  );
+  return new In(elementExpr, exprOthers);
+}
+
+export function notInAny(element: Expr, others: Expr[]): Not;
+export function notInAny(element: Expr, others: any[]): Not;
+export function notInAny(element: string, others: Expr[]): Not; // Added overload
+export function notInAny(element: string, others: any[]): Not; // Added overload
```
MarkDuckworth, line 1114: Same

MarkDuckworth (commented):
Done with reviews for now

dev/system-test/pipeline.ts
wu-hui: Good question. It looks like mapGet does not support nested fields (always returns null). 

One benefit is maybe mapGet supports field with special chars better (., /, etc).

dev/system-test/pipeline.ts
wu-hui: See above.

wu-hui (commented):
Thanks for the feedback! I addressed some of those, and maybe we should discuss for the rest.

dev/system-test/pipeline.ts
```diff
@@ -0,0 +1,733 @@
...
+      .where(eq('tagsCount', 3))
+      .execute();
+    expect(results.length).to.equal(10);
+  });
+
+  it('arrayConcat works', async () => {
+    const results = await randomCol
+      .pipeline()
+      .select(
+        Field.of('tags').arrayConcat('newTag1', 'newTag2').as('modifiedTags')
```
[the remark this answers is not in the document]
wu-hui: Will do.

dev/src/pipeline.ts
```diff
@@ -0,0 +1,612 @@
...
+    );
+    const structuredPipeline: IStructuredPipeline = {pipeline: {stages}};
+    return {
+      database: this.db.formattedName,
+      structuredPipeline,
+    };
+  }
+}
+
+class PaginatingPipeline {
```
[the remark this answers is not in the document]
wu-hui: Will delete.

dev/src/expression.ts
wu-hui: Will do.

dev/src/expression.ts
wu-hui: Will remove.

dev/src/expression.ts
wu-hui: Will do.

dev/src/expression.ts
wu-hui: Will do.

dev/src/expression.ts
wu-hui: Will rename Expr.in to inAny.

dev/src/expression.ts
```diff
@@ -0,0 +1,6739 @@
...
+   * // Create a Field instance for a nested field 'author.firstName'
+   * const authorFirstNameField = Field.of("author.firstName");
+   * ```
+   *
+   * @param name The path to the field.
+   * @return A new {@code Field} instance representing the specified field.
+   */
+  static of(name: string): Field;
+  static of(path: firestore.FieldPath): Field;
+  static of(nameOrPath: string | firestore.FieldPath): Field;
```
MarkDuckworth, line 1875: This overload is redundant. `static of(nameOrPath: string | firestore.FieldPath): Field;` Either remove this or keep it and remove the two above.

dev/src/expression.ts
```diff
@@ -0,0 +1,6739 @@
...
+   */
+  static of(value: firestore.DocumentReference): Constant;
+
+  /**
+   * Creates a `Constant` instance for a Firestore proto value.
+   *
+   * @param value The Firestore proto value.
+   * @return A new `Constant` instance.
+   */
+  static of(value: api.IValue): Constant;
```
MarkDuckworth, line 2051: This appears to be redeclared below on line 2083

dev/src/expression.ts
```diff
@@ -0,0 +1,6739 @@
...
+      return new Constant(new VectorValue(value as Array<number>));
+    }
+  }
+
+  _toProto(serializer: Serializer): api.IValue {
+    if (isFirestoreValue(this.value)) {
+      return this.value;
+    }
+
+    return serializer.encodeValue(this.value)!;
```
MarkDuckworth, line 2112: `serializer.endcodeValue(undefined)` will return `null`, which is an permitted, but unhandled case. Do we want to treat undefined as null?

PR #2181, opened 2024-09-04
refactor: Add settings attributes to trace spans.

ehsannas:
This PR adds settings attributes to spans. This information could be useful when debugging performance issues.


dev/src/index.ts
```diff
@@ -808,10 +807,6 @@ export class Firestore implements firestore.Firestore {
     }
 
     if (createEnabledInstance) {
-      // Re-use the existing EnabledTraceUtil if one has been created.
-      if (this._traceUtil && this._traceUtil instanceof EnabledTraceUtil) {
-        return this._traceUtil;
-      }
```
ehsannas, line 814: Since we are now capturing settings as part of the trace attributes, if the settings change we need to also change the traceUtil instance. So we should _not_ re-use the existing one.

dev/src/telemetry/enabled-trace-util.ts
```diff
@@ -42,6 +49,85 @@ export class EnabledTraceUtil implements TraceUtil {
...
+
+    if (settings.preferRest !== undefined) {
+      this.settingsAttributes[`${ATTRIBUTE_SETTINGS_PREFIX}.prefer_REST`] =
+        settings.preferRest;
+    }
+
+    this.settingsAttributes[`${ATTRIBUTE_SETTINGS_PREFIX}.max_idle_channels`] =
+      settings.maxIdleChannels ?? DEFAULT_MAX_IDLE_CHANNELS;
+
+    const defaultRetrySettings = serviceConfig.retry_params.default;
```
milaGGL, line 79: why there are retry and timeout configuration? how this is related to otel tracing?

ehsannas: this code is adding trace attributes to the spans. The purpose is that if a user has misconfigured their SDK, their configurations will surface in the trace and may provide insights into the issues they're running into. This is similar to the [Java impl](https://github.com/googleapis/java-firestore/blob/1c7c12f915e76f25652adbad0864c5c7d810e268/google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledTraceUtil.java#L269). Although not all java configurations exist in the node.js SDK.

milaGGL (commented):
LGTM

milaGGL (approved):
LGTM

PR #2219, opened 2024-10-15
feat: Enable tracing via OpenTelemetry

ehsannas:


ehsannas:
Going to decouple the firestore.d.ts revamp from the tracing work.

PR #2236, opened 2024-11-26
feature: New Pipeline Stages

tom-andersen:


dev/src/pipeline.ts
```diff
@@ -484,14 +477,226 @@ export class Pipeline<AppModelType = firestore.DocumentData>
...
+
+  /**
+   * Performs a pseudo-random sampling of the documents from the previous stage.
+   *
+   * <p>This stage will filter documents pseudo-randomly. The parameter specifies how number of
+   * documents to be returned.
+   *
+   * <p>Examples:
+   *
+   * <pre>{@code
```
MarkDuckworth, line 546: I don't think this `@code` tag is needed.

dev/src/pipeline.ts
```diff
@@ -484,14 +477,226 @@ export class Pipeline<AppModelType = firestore.DocumentData>
...
+   * will for each input array element produce an augmented document. The input array element will
+   * augment the previous stage document by replacing the field specified by `fieldName` parameter
+   * with the element value.
+   *
+   * <p>In other words, the field containing the input array will be removed from the augmented
+   * document and replaced by the corresponding array element.
+   *
+   * <p>Example:
+   *
+   * <pre>{@code
```
MarkDuckworth, line 638: Same.

dev/src/pipeline.ts
```diff
@@ -484,14 +477,226 @@ export class Pipeline<AppModelType = firestore.DocumentData>
...
+  /**
+   * Produces a document for each element in array found in previous stage document.
+   *
+   * <p>For each previous stage document, this stage will emit zero or more augmented documents. The
+   * input array found in the previous stage document field specified by the `fieldName` parameter,
+   * will for each input array element produce an augmented document. The input array element will
+   * augment the previous stage document by replacing the field specified by `fieldName` parameter
+   * with the element value.
+   *
+   * <p>In other words, the field containing the input array will be removed from the augmented
```
MarkDuckworth, line 633: Should we document the behavior if the specified field is not an array?

dev/src/pipeline.ts
```diff
@@ -484,14 +477,226 @@ export class Pipeline<AppModelType = firestore.DocumentData>
...
+   *
+   * <p>Example:
+   *
+   * <pre>{@code
+   * // Input:
+   * // { 'title': 'The Hitchhiker's Guide to the Galaxy', 'tags': [ 'comedy', 'space', 'adventure' ], ... }
+   *
+   * // Emit a book document for each tag of the book.
+   * firestore.pipeline().collection('books')
+   *     .unnest({ field: 'tags', UnnestOptions.indexField('tagIndex'));
```
MarkDuckworth, line 679: `UnnestOptions.indexField('tagIndex')` doesn't look like like it's implemented. Is this correct?

dev/src/pipeline.ts
```diff
@@ -484,14 +477,226 @@ export class Pipeline<AppModelType = firestore.DocumentData>
...
+   * // Output:
+   * // { 'title': 'The Hitchhiker's Guide to the Galaxy', 'tagIndex': 0, 'tags': 'comedy', ... }
+   * // { 'title': 'The Hitchhiker's Guide to the Galaxy', 'tagIndex': 1, 'tags': 'space', ... }
+   * // { 'title': 'The Hitchhiker's Guide to the Galaxy', 'tagIndex': 2, 'tags': 'adventure', ... }
+   * }</pre>
+   *
+   * @param fieldName The name of the field containing the array.
+   * @param options The {@code UnnestOptions} options.
+   * @return A new {@code Pipeline} object with this stage appended to the stage list.
+   */
```
MarkDuckworth, line 690: Can we document the use case for `indexField`? Or otherwise why someone would use this overload?

types/firestore.d.ts
```diff
@@ -9242,6 +9354,59 @@ declare namespace FirebaseFirestore {
...
+     * // Output:
+     * // { "title": "The Hitchhiker's Guide to the Galaxy", "tags": "comedy", ... }
+     * // { "title": "The Hitchhiker's Guide to the Galaxy", "tags": "space", ... }
+     * // { "title": "The Hitchhiker's Guide to the Galaxy", "tags": "adventure", ... }
+     * }</pre>
+     *
+     * @param field The name of the field containing the array.
+     * @return A new {@code Pipeline} object with this stage appended to the stage list.
+     */
+    unnest(field: Selectable | string): Pipeline<AppModelType>;
```
MarkDuckworth, line 9408: Do we also need a declaration for the `unnest(options: UnnestOptions)` overload?

dev/src/expression.ts
```diff
@@ -1942,38 +1942,6 @@ export class Field extends Expr implements Selectable {
   }
 }
 
-/**
- * @beta
- */
-export class Fields extends Expr implements Selectable {
```
MarkDuckworth, line 1948: Did we remove Fields intentionally? I may have missed the decision to do that.

dev/system-test/pipeline.ts
```diff
@@ -972,4 +972,39 @@ describe.only('Pipeline class', () => {
...
+      .execute();
+    expect(results.length).to.equal(6);
+  });
+
+  it('run pipeline with union', async () => {
+    const results = await randomCol
+      .pipeline()
+      .union(randomCol.pipeline())
+      .execute();
+    expect(results.length).to.equal(20);
```
MarkDuckworth, line 999: It's worth asserting that we got the expected documents in the result.

dev/system-test/pipeline.ts
```diff
@@ -972,4 +972,39 @@ describe.only('Pipeline class', () => {
...
+  });
+
+  it('run pipeline with unnest', async () => {
+    const results = await randomCol
+      .pipeline()
+      .where(eq('title', "The Hitchhiker's Guide to the Galaxy"))
+      .unnest('tags')
+      .execute();
+    expect(results.length).to.equal(3);
+  });
```
MarkDuckworth, line 1009: It's worth asserting that we got the expected documents with expected tag values in the result set.

dev/src/expression.ts
tom-andersen: I argued the case for doing this with @wu-hui. The rationale is that some new stages can take a single Field as parameter. So, having Selectable include multi-field implementation creates a type safety problem. The other reason is that `Fields` is effectively the same as `List<Field>`. I don't think there is value in creating a type for this.

So this is a worthy discussion, but I would rather frame the discussion as why should `Fields` exist in the first place?

dev/src/pipeline.ts
tom-andersen: Added and fixed up documentations.

dev/system-test/pipeline.ts
tom-andersen: Done.

tom-andersen (commented):
I have updated PR. @MarkDuckworth, please take another look.

dev/src/expression.ts
MarkDuckworth: SGTM

dev/src/pipeline.ts
MarkDuckworth: I agree on documenting the mechanics. My comment was poorly worded, but I think my concern still stands. IMO, we don't clearly document what happens when `indexField` is set. At least it wasn't clear to me until I read the linked doc.

We could add something like this to the docs for this method or for `UnnestOptions`:

When `indexField` is provided, a new field is added to the emitted documents, storing the offset (starting at zero) into the array the expanded element is from.

dev/src/pipeline.ts
MarkDuckworth: Thanks. Was looking in the wrong place.

MarkDuckworth (approved):
LGTM with one remaining comment about documentation for `indexField`.

PR #2257, opened 2024-12-06
Fix: sort document reference by long type id 

milaGGL:
Document ID supports strings and well as long integers in the format of  "__id<Long>__" . When sorting documents by document ID, it should be sorted in the following order:
1. Long (numeric order)
2. String (lexicographic order)

### DIFF
--- a/dev/src/path.ts
+++ b/dev/src/path.ts
@@ -185,16 +185,13 @@ abstract class Path<T> {
       return 1;
     } else if (isLhsNumeric && isRhsNumeric) {
       // both numeric
-      return Math.sign(this.extractNumericId(lhs) - this.extractNumericId(rhs));
+      return this.compareNumbers(
+        this.extractNumericId(lhs),
+        this.extractNumericId(rhs)
+      );
     } else {
       // both non-numeric
-      if (lhs < rhs) {
-        return -1;
-      }
-      if (lhs > rhs) {
-        return 1;
-      }
-      return 0;
+      return this.compareStrings(lhs, rhs);
     }
   }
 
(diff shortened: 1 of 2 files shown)


dev/system-test/firestore.ts
```diff
@@ -3913,6 +3913,87 @@ describe('Query class', () => {
       unsubscribe();
     });
 
+    it('snapshot listener sorts query by DocumentId same way as server', async () => {
```
milaGGL, line 3916: add negative number id

ehsannas (changes requested):

dev/src/path.ts
```diff
@@ -158,20 +162,50 @@ abstract class Path<T> {
...
+      // Only lhs is numeric
       return -1;
-    }
-    if (this.segments.length > other.segments.length) {
+    } else if (!isLhsNumeric && isRhsNumeric) {
+      // Only rhs is numeric
       return 1;
+    } else if (isLhsNumeric && isRhsNumeric) {
+      // both numeric
+      return Math.sign(this.extractNumericId(lhs) - this.extractNumericId(rhs));
```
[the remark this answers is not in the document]
milaGGL: good point.

ehsannas (approved):

dev/src/path.ts
```diff
@@ -158,20 +162,70 @@ abstract class Path<T> {
...
+      );
+      if (comparison !== 0) {
+        return comparison;
       }
     }
-    if (this.segments.length < other.segments.length) {
+    return this.compareNumbers(
+      BigInt(this.segments.length),
+      BigInt(other.segments.length)
+    );
```
ehsannas, line 176: ultra-nit [feel free to ignore]: segment lengths are usually very small positive numbers.
Doing `return Math.sign(this.segments.length - other.segments.length);` is slightly better than converting to BigInt and then comparing.

ehsannas (approved):

PR #2259, opened 2024-12-12
fix: correctly escape field paths with multiple backslashes or backticks

### LINKED ISSUE #2019: FieldPath only escapes the first `\` and '`', resulting in INVALID_ARGUMENT response
#### Environment details

  - OS: MacOS
  - Node.js version: latest
  - npm version: latest
  - `@google-cloud/firestore` version: v7.3.1

#### Steps to reproduce

1. Try to use a FieldPath with multiple backslashes or backtick, for example: 

    ```ts
    query.where(new FieldPath('containing multiple `` or backslashes \\\\'), '==', 0)
    ```

2. Run the query, then you get: `3 INVALID_ARGUMENT: Invalid property path ...`

This is because of the use of `replace` in the following line: https://github.com/googleapis/nodejs-firestore/blob/e598b9daf628cbc54dc10dab80bb0f46e2a3e2a2/dev/src/path.ts#L624
The `replace()` method only replaces the first occurence of the given pattern (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) for more info).

albertnis:
Previously, only the first backslash and backtick was being escaped. This broke `update` and `set (merge)` calls when a field path contained multiple backslashes.

`replaceAll` is an alternative here, but I think a global regex has better compatibility with older versions of Node.

Fixes #2019 🦕

### DIFF
--- a/dev/src/path.ts
+++ b/dev/src/path.ts
@@ -621,7 +621,7 @@ export class FieldPath extends Path<FieldPath> implements firestore.FieldPath {
       .map(str => {
         return UNESCAPED_FIELD_NAME_RE.test(str)
           ? str
-          : '`' + str.replace('\\', '\\\\').replace('`', '\\`') + '`';
+          : '`' + str.replace(/\\/g, '\\\\').replace(/`/g, '\\`') + '`';
       })
       .join('.');
   }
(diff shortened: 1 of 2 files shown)


dconeybe (approved):
Thanks for the fix, and the updated unit tests!

PR #2269, opened 2025-01-04
patch: Delete duplicate GitHub issue template

dandv:
The .yml is much newer and doesn't have broken links.

image


ehsannas (approved):
Thanks @dandv

PR #2275, opened 2025-01-06
FIX: sort strings in UTF-8 encoded byte order

milaGGL:
Strings should be sorted in UTF-8 encoded byte order. Public document: https://cloud.google.com/firestore/docs/concepts/data-types#data_types

SDK sorts strings using built in comparator method, which sorts lexicographically, and leads to mismatch between server and sdk when special characters are present. This PR fixes the string order mismatches on document field, map key, and document key.

### DIFF
--- a/dev/src/order.ts
+++ b/dev/src/order.ts
@@ -214,7 +214,7 @@ function compareObjects(left: ApiMapValue, right: ApiMapValue): number {
   leftKeys.sort();
   rightKeys.sort();
   for (let i = 0; i < leftKeys.length && i < rightKeys.length; i++) {
-    const keyComparison = primitiveComparator(leftKeys[i], rightKeys[i]);
+    const keyComparison = compareUtf8Strings(leftKeys[i], rightKeys[i]);
     if (keyComparison !== 0) {
       return keyComparison;
     }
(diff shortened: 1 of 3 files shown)


dev/src/order.ts
```diff
@@ -248,6 +248,21 @@ function compareVectors(left: ApiMapValue, right: ApiMapValue): number {
   return compareArrays(leftArray, rightArray);
 }
 
+function stringToUTF8Bytes(str: string): Uint8Array {
```
ehsannas, line 251: to be consistent (with `compareUtf8Strings`), use the name `stringToUtf8Bytes`

dev/system-test/firestore.ts
```diff
@@ -4084,6 +4084,156 @@ describe('Query class', () => {
...
+
+        const unsubscribe = query.onSnapshot(snapshot =>
+          currentDeferred.resolve(snapshot)
+        );
+        const watchSnapshot = await waitForSnapshot();
+        snapshotsEqual(watchSnapshot, {
+          docs: getSnapshot.docs,
+          docChanges: getSnapshot.docChanges(),
+        });
+        unsubscribe();
```
ehsannas, line 4114: there are no `expect` statements after line 4104 in this test.  are you missing the `expect` statement for `watchSnapshot`? same for the other tests.

milaGGL: `snapshotsEqual` asserts `watchSnapshot` has the same docs as getSnapshot.

ehsannas: oops. not sure how I missed that :)

ehsannas: LGTM

ehsannas (approved):

Issue #2280, opened 2025-01-08
Race condition in onSnapshot code crashes the process


jerzozwierz:
### Bottom line up front
When the conditions are just right, it's possible for a grpc stream (underlying an onSnapshot query) to emit an `error` event while not having any error handlers, killing the node process. It's rare, but our server's heavy traffic causes it to happen multiple times per day.

1) Is this a client library issue or a product issue?
Client library. A transient error isn't properly handled by the library under certain conditions.

2) Did someone already solve this?
None of my searches have yielded results.

3) Do you have a support contract?
I don't

#### Environment details

  - OS: Debian GNU/Linux 12 (inside Docker)
  - Node.js version: 20.18.0
  - npm version: 10.8.2
  - `@google-cloud/firestore` version: 7.4.0

#### Steps to reproduce

While I can't write code that reliably reproduces the problem (it's a low probability race condition), I can point to a code path that is hit and support my hypothesis with logs emitted by our server.

https://github.com/googleapis/nodejs-firestore/blob/main/dev/src/watch.ts#L483-L513
```
// inside Watch.initStream() method
...
        return this.firestore
          .requestStream( // (1)
            'listen',
            /* bidirectional= */ true,
            request,
            this.requestTag
          )
          .then(backendStream => {
            if (!this.isActive) { // (2)
              logger(
                'Watch.initStream',
                this.requestTag,
                'Closing inactive stream'
              );
              backendStream.emit('end');
              return; // (3)
            }
            logger('Watch.initStream', this.requestTag, 'Opened new stream');
            this.currentStream = backendStream;
            this.resetIdleTimeout();

            this.currentStream!.on('data', (proto: api.IListenResponse) => {
              this.resetIdleTimeout();
              this.onData(proto);
            })
              .on('error', err => { // (4)
                if (this.currentStream === backendStream) {
                  this.currentStream = null;
                  this.maybeReopenStream(err);
                }
              })
```

1. First thing that must happen is a query being closed very soon after opening. This way, before the async call **(1)**, the `.isActive` field is flipped back to false. Subsequently, the code enters the `if` clause **(2)**. Note that, in this path, we exit the function at **(3)** and `backendStream` **doesn't get an error handler like it does in (4)**. 
2. Secondly, the newly opened `backendStream` must emit an error (it can be anything, like a transient networking issue). The fact that it receives an `end` event earlier doesn't preclude it from doing so.
3. Consequently, that emitted error crashes the node process.

[...]

PR #2283, opened 2025-01-13
fix: prevent crashes if an inactive stream receives an error.

ehsannas:


ehsannas:
Fixes https://github.com/googleapis/nodejs-firestore/issues/2280

wu-hui (approved):

PR #2299, opened 2025-02-18
Fix: Use lazy encoding for utf-8 encoded string comparison

milaGGL:
The previous [fix](https://github.com/googleapis/nodejs-firestore/pull/2299 ) created a performance issue due to expensive UTF-8 encoding. Update `compareUtf8Strings` to use lazy encoding instead.

### DIFF
--- a/dev/src/order.ts
+++ b/dev/src/order.ts
@@ -273,7 +273,9 @@ export function compareUtf8Strings(left: string, right: string): number {
           Buffer.from(leftBytes),
           Buffer.from(rightBytes)
         );
-        if (comp !== 0) return comp;
+        if (comp !== 0) {
+          return comp;
+        }
       }
     }
   }
--- a/dev/src/order.ts
+++ b/dev/src/order.ts
@@ -254,8 +254,7 @@ function compareVectors(left: ApiMapValue, right: ApiMapValue): number {
  * @internal
  */
 export function compareUtf8Strings(left: string, right: string): number {
-  let i = 0;
-  while (i < left.length && i < right.length) {
+  for (let i = 0; i < left.length && i < right.length; i++) {
     const leftCodePoint = left.codePointAt(i)!;
     const rightCodePoint = right.codePointAt(i)!;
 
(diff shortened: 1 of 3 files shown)


dev/src/order.ts
```diff
@@ -248,19 +248,38 @@ function compareVectors(left: ApiMapValue, right: ApiMapValue): number {
...
+    if (leftCodePoint !== rightCodePoint) {
+      if (leftCodePoint < 128 && rightCodePoint < 128) {
+        // ASCII comparison
+        return primitiveComparator(leftCodePoint, rightCodePoint);
+      } else {
+        // Lazy instantiate TextEncoder
+        const encoder = new TextEncoder();
+
+        // UTF-8 encoded byte comparison, substring 2 indexes to cover surrogate pairs
+        const leftBytes = encoder.encode(left.substring(i, i + 2));
```
dconeybe, line 271: What happens if the character at `i` is a single utf16 code unit but the character at `i+1` is a leading surrogate? In that case, will the encoding still occur correctly?

milaGGL: we are going index by index, unless it is surrogate pairs. 
if characters at `i` equals, loop goes to `i+1`, where we start comparison at leading surrogate. 
if they are not equal, we would encode characters at `i` and `i+1`, which falsely included the leading surrogate. but, the characters at index `i` already differs, we would have the result even before the byte code for `i+1` being compared.

milaGGL: encoding all remaining substrings are a safe way to go. But also feels like a waste since we are already at the index where it differs. 

We could go for safe solution, but like u mentioned in offline discussion, if the user is storing non-english language into the fields, they would still encounter the performance issue. Because we are basically encoding all strings for them. ie:
```
{text: "这是一个非常长的文本“}
{text: "那是一个非常长的文本“}
... // and other fields that we could have told they are different at the very beginning
```

milaGGL: How about this:
```
const leftBytes = encoder.encode(getUtf8SafeSubstring(left, i));

function getUtf8SafeSubstring(str: string, index: number): string {
  const firstCodePoint = str.codePointAt(index)!;
  if (firstCodePoint > 0xffff) {
    // It's a surrogate pair, return the whole pair
    return str.substring(index, index + 2);
  } else {
    // It's a single code point, return it
    return str.substring(index, index + 1);
  }
}
```

In this way, we do not have the magic number 2, and it solved the broken surrogate pair issue as well.

dconeybe: I'm not suggesting that there is a bug, and my brain can't reason fully about the logic. And without unit tests it's really hard to see if there is a bug or not. If you port the string generator from https://github.com/firebase/firebase-android-sdk/pull/6706 and the tests pass then I'll feel better :)

dev/src/order.ts
```diff
@@ -248,19 +248,38 @@ function compareVectors(left: ApiMapValue, right: ApiMapValue): number {
...
-  const leftBytes = stringToUtf8Bytes(left);
-  const rightBytes = stringToUtf8Bytes(right);
-  return compareBlobs(Buffer.from(leftBytes), Buffer.from(rightBytes));
+  let i = 0;
+  while (i < left.length && i < right.length) {
+    const leftCodePoint = left.codePointAt(i)!;
+    const rightCodePoint = right.codePointAt(i)!;
+
+    if (leftCodePoint !== rightCodePoint) {
+      if (leftCodePoint < 128 && rightCodePoint < 128) {
```
dconeybe, line 263: Consider making the body of this outer "if" statement a single line, and factor out the logic to do the actual comparison to a helper function. IMO this will improve readability because, from a quick code inspection, it's not obvious that the function is guaranteed to return if the "if" block condition is satisfied.

I'm thinking something like this:

```ts
if (leftCodePoint !== rightCodePoint) {
  return isLeftGreaterAt(left, right, i) ? 1 : -1;
}
```

This also makes it clear that zero is _never_ a potential return value once the code enters this "if" block.

dconeybe: (double-check my logic there... I did not test the code!)

milaGGL: this down the block somehow conveys the same message
```      
if (comp !== 0) {
    return comp;
}
```

`isLeftGreaterAt` returning true/false duplicates the logic where we check if the compare result is 1 or -1 inside the function, and we check the `isLeftGreaterAt` result again to decide return 1 or -1.

ehsannas: I understand that the ascii set is 0-127, but the cutoff for one-char vs two-char code points is `0xffff`. Should this condition be changed to

```
if (leftCodePoint < 0xffff && rightCodePoint < 0xffff) {
```
?

dconeybe: Consider omitting this special treatment of code points less than 128 in this PR. This special treatment is effectively an _optimization_, and I would suggest implementing optimizations like this in a follow-up PR, once correctness has been established.

ehsannas (approved):

PR #2331, opened 2025-04-17
fix: Aggregate query readtime bug

### LINKED ISSUE #2330: ‎AggregateQuery looses nanoseconds in `readTime`
During unit tests we create sample documents and then very quickly execute queries using a `readTime` In some tests we simply use `Date.now()` for the `readTime` and found that tests sometimes fail to find the (very) recently created documents when using an `AggregateQuery`

After some debugging we think we found the issue. When executing a normal query you end up at `Query.toProto` and this uses `runQueryRequest.readTime = transactionOrReadTime.toProto().timestampValue` to set the `readTime` on the request:
https://github.com/googleapis/nodejs-firestore/blob/5937b93aa1aacd2f63bf1678ec569c504cf0b186/dev/src/reference/query.ts#L1371

However, when using `readTime` with an `AggregateQuery` you end up in `AggregateQuery.toProto` and that uses `runQueryRequest.readTime = transactionOrReadTime` to set the `readTime` on the request: https://github.com/googleapis/nodejs-firestore/blob/5937b93aa1aacd2f63bf1678ec569c504cf0b186/dev/src/reference/aggregate-query.ts#L338

So, for the `AggregateQuery` the `.toProto().timestampValue` part is missing. This would have converted the `nanoseconds` property to `nanos` as required by protobuf. Without this conversion, `nanos` remains `undefined` and is translated to `0` over the wire.

Then end effect is that using a `readTime` of `{ _seconds: 1744812999, _nanoseconds: 978000000 }` is sent to the backend as `{ _seconds: 1744812999, _nanoseconds: 0 }` meaning it will not include documents in the aggregation query that were created between `{ _seconds: 1744812999, _nanoseconds: 0 }` and `{ _seconds: 1744812999, _nanoseconds: 978000000 }`. This is a small window (of max 1 second), but it is something we hit during our unit tests.

This is a client library issue.

Could not find anything

Yes, support case #58916783 filed.

#### Environment details

  - OS: MacOS 15.3.2
  - Node.js version: 22.14.0
  - npm version: 10.9.2
  - `@google-cloud/firestore` version: 7.11.0

#### Steps to reproduce

Minimal steps to reproduce:

```typescript
import { Firestore } from '@google-cloud/firestore';
import { Timestamp } from '@google-cloud/firestore/build/src/timestamp';
import assert from 'node:assert';
import { randomUUID } from 'node:crypto';

void (async () => {
    const firestore = new Firestore({ projectId: 'my-project' });
    const collection = firestore.collection(randomUUID());
    const doc = collection.doc();
    await doc.create({ some: 'data' });
    const query = collection.count() as import('@google-cloud/firestore/build/src/reference/aggregate-query').AggregateQuery<
        { count: FirebaseFirestore.AggregateField<number> },
        FirebaseFirestore.DocumentData,
        FirebaseFirestore.DocumentData
    >;
    // uncomment the line below to make the test succeed (since we wait for the next second)
    // await new Promise(resolve => setTimeout(resolve, 1000));
    const timestamp = Timestamp.fromMillis(Date.now());
    const count = await query._get(timestamp);
    // this assertion fails as 0 documents were found
    assert.strictEqual(count.result.data().count, 1);
})();
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

tom-andersen:
Fixes #2330

### DIFF
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -5013,11 +5013,21 @@ describe('Aggregation queries', () => {
   it('can run count within a transaction with readtime', async () => {
     const doc = col.doc();
     const writeResult: WriteResult = await doc.create({some: 'data'});
+
     const count = await firestore.runTransaction(t => t.get(col.count()), {
       readOnly: true,
       readTime: writeResult.writeTime,
     });
     expect(count.data().count).to.equal(1);
+
+    const countBefore = await firestore.runTransaction(
+      t => t.get(col.count()),
+      {
+        readOnly: true,
+        readTime: Timestamp.fromMillis(writeResult.writeTime.toMillis() - 1),
+      }
+    );
+    expect(countBefore.data().count).to.equal(0);
   });
 
   it('can run count query using aggregate api', async () => {
--- a/dev/system-test/firestore.ts
+++ b/dev/system-test/firestore.ts
@@ -5010,6 +5010,16 @@ describe('Aggregation queries', () => {
     return Promise.all(sets);
   }
 
+  it('can run count within a transaction with readtime', async () => {
+    const doc = col.doc();
+    const writeResult: WriteResult = await doc.create({some: 'data'});
+    const count = await firestore.runTransaction(t => t.get(col.count()), {
+      readOnly: true,
+      readTime: writeResult.writeTime,
+    });
+    expect(count.data().count).to.equal(1);
+  });
+
   it('can run count query using aggregate api', async () => {
     const testDocs = {
       a: {author: 'authorA', title: 'titleA'},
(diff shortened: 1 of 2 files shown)


dev/system-test/firestore.ts
```diff
@@ -5010,6 +5010,16 @@ describe('Aggregation queries', () => {
...
   }
 
+  it('can run count within a transaction with readtime', async () => {
+    const doc = col.doc();
+    const writeResult: WriteResult = await doc.create({some: 'data'});
+    const count = await firestore.runTransaction(t => t.get(col.count()), {
+      readOnly: true,
+      readTime: writeResult.writeTime,
+    });
+    expect(count.data().count).to.equal(1);
```
MarkDuckworth, line 5020: Would it make sense to run two transactions, one with a readtime before the write and one with a readtime after?

MarkDuckworth (approved):
Looks good other than the thought on test improvement

dev/system-test/firestore.ts
tom-andersen: I expanded the test. PTAL

MarkDuckworth: LGTM. thanks!

PR #2361, opened 2025-06-12
fix: Update `tsconfig.json` with ES2020 as target.

ehsannas:
The SDK has been [using](https://github.com/search?q=repo%3Agoogleapis%2Fnodejs-firestore%20bigint&type=code) `BigInt` for a long time without explicitly having `target: "es2020"` in the `tsconfig.json` file. Based on my research, ES2016 did not include BigInt. I suspect that the omission of `target` in `tsconfig.json` had resulted in es2020 to be used implicitly.

Also note that Firebase Admin Node sdk [uses](https://github.com/firebase/firebase-admin-node/blob/master/tsconfig.json#L4) es2020.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,8 +3,9 @@
   "compilerOptions": {
     "resolveJsonModule": true,
     "outDir": "build",
+    "target": "es2020",
     "lib": [
-      "es2016",
+      "es2020",
       "dom"
     ],
     "useUnknownInCatchVariables": false,


ehsannas:
@MarkDuckworth @sofisl Is this considered a breaking change? If so, are we allowed to make this change now?

@milaGGL FYI

In support of new data types, Firestore will need to use BigInt literals followed by `n` (things like `const largeNumber: bigint = 12345678901234567890n;`). And this will not work without the change to tsconfig.json.

sofisl:
This can be considered a breaking change, yes. I'd recommend moving up to Node 18 as the library is due to move, see b/397525375. You are safe to do it this year as most other libraries moved to node 18 this year as well.

PR #2364, opened 2025-06-16
fix: Firestore Client caching stub in bad state issue

wu-hui:


dev/src/v1/firestore_client.ts
```diff
@@ -305,6 +306,14 @@ export class FirestoreClient {
...
       this._providedCustomServicePath
     ) as Promise<{[method: string]: Function}>;
+    
+    this.firestoreStub
+    .then(() => {
+      this._stubState = 'Success';
+    })
+    .catch(err => {
+      this._stubState = 'Fail';
+    });
```
ehsannas, line 316: IIUC this file is [auto-generated](https://github.com/googleapis/nodejs-firestore/blob/ae7e38d68d79ca8aa6466a3814340446df52c572/dev/src/v1/firestore_client.ts#L15), and these changes will be overwritten.

PR #2365, opened 2025-06-16
fix: Firestore Client caching stub in bad state issue

wu-hui:
fixes b/421230880

### DIFF
--- a/dev/src/v1/firestore_client.ts
+++ b/dev/src/v1/firestore_client.ts
@@ -62,6 +62,7 @@ export class FirestoreClient {
   private _defaults: {[method: string]: gax.CallSettings};
   private _universeDomain: string;
   private _servicePath: string;
+  private _stubState: 'Success' | 'Fail' | undefined;
   auth: gax.GoogleAuth;
   descriptors: Descriptors = {
     page: {},
@@ -289,7 +290,7 @@ export class FirestoreClient {
    */
   initialize() {
     // If the client stub promise is already initialized, return immediately.
-    if (this.firestoreStub) {
+    if (this.firestoreStub && this._stubState === 'Success') {
       return this.firestoreStub;
     }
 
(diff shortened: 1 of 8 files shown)


MarkDuckworth (dismissed):

dev/src/v1/firestore_client.ts
```diff
@@ -347,9 +351,13 @@ export class FirestoreClient {
             }
             const func = stub[methodName];
             return func.apply(stub, args);
-          },
-        (err: Error | null | undefined) => () => {
-          throw err;
+          };
+        },
+        (err: Error | null | undefined) => {
+          this._stubFailed = true;
```
tom-andersen, line 357: Is there a way to make sure that setting "this._stubFailed = true" only happens once?

If for some reason the stub lives longer than expected and continues to experience errors, then the fresh reset stub will be reset by an old stub.

Also, this seems like an unexpected situation, and compensating behavior. Should we maybe add some logging to gain insight when this happens?

tom-andersen (dismissed):
LGTM

dev/src/v1/firestore_client.ts
MarkDuckworth: The conditional block on ln 293, `if (this.firestoreStub ...) return this.firestoreStub` ensures that only one logical task will attempt to create a `firestoreStub`. The `firestoreStub` instance is a Promise, which will only resolve or reject once. Therefore the error handler on ln 357 will be called at max 1 time per `firestoreStub` instance. And we won't have two concurrent attempts to create the stub. So, I think this is safe.

Good idea on the logging. I will add some.

MarkDuckworth: logging added

MarkDuckworth (approved):

ehsannas (approved):

PR #2395, opened 2025-07-22
Markduckworth/better reference value handling

MarkDuckworth:

### DIFF
--- a/dev/src/path.ts
+++ b/dev/src/path.ts
@@ -44,7 +44,7 @@ export const DEFAULT_DATABASE_ID = '(default)';
  */
 const RESOURCE_PATH_RE =
   // Note: [\s\S] matches all characters including newlines.
-  /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents\/)?([\s\S]*)$/;
+  /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents(?:\/([\s\S]+))?)?$/;
 
 /*!
  * A regular expression to verify whether a field name can be passed to the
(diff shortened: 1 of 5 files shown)


dev/src/path.ts
```diff
@@ -44,7 +44,7 @@ export const DEFAULT_DATABASE_ID = '(default)';
  */
 const RESOURCE_PATH_RE =
   // Note: [\s\S] matches all characters including newlines.
-  /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents\/)?([\s\S]*)$/;
+  /^projects\/([^/]*)\/databases\/([^/]*)(?:\/documents(?:\/([\s\S]+))?)?$/;
```
tom-andersen, line 47: This new regex still has problems. It should not match:

projects/PPP/databases/DDD/wrong
projects/PPP/databases/DDD//
projects/PPP/databases/DDD/documentsBAD
projects/PPP/databases/DDD/documents//
projects/PPP/databases/DDD/documents/ok//ok

Following the database capture is empty string.
projects/PPP/databases//DDD/documents

dev/src/path.ts
```diff
@@ -455,7 +455,7 @@ export class QualifiedResourcePath extends ResourcePath {
     if (elements) {
       const project = elements[1];
       const database = elements[2];
-      const path = elements[3];
+      const path = elements[3] ?? '';
```
tom-andersen, line 458: Nit: If we know path is empty, then I think we should avoid calling append (even it it supports it).

tom-andersen (changes requested):

dev/src/path.ts
MarkDuckworth: Most of those did not match. but I did improve the regex to not match paths with empty path segments. Thanks for the examples!

dev/src/path.ts
MarkDuckworth: Yeah, nice improvement. I've updated this.

tom-andersen (approved):
LGTM.

PR #2417, opened 2025-09-12
chore: bump node 14 to node 18 in kokoro

dconeybe:
The warning below highlights that these changes should be done via synthtool. But I don't know how to use that tool and I just need to get the presubmits working. So I'm going to try to merge this PR anyways and we can deal with synthtool some other day.


MarkDuckworth (approved):
Okay to merge if we want to unblock for a quick release. I'm also looking into the fix for our owlbot post processor.

PR #2426, opened 2025-09-25
change: pool.ts: add even more logging

dconeybe:
This is a follow-up to PRs #2416 and #2420 to add even more logging to `ClientPool`.

The new/updated log messages include:
* an "instance ID" for the ClientPool in each log message, to enable distinguishing log messages from distinct ClientPool instances.
* the result of `this.activeClients.delete(client)` in ClientPool.release(), to ensure that it is, indeed, deleting an entry.
* the active request count for each client in LazyLogStringForAllClientIds.
* a message from ClientPool.terminate() that is unconditionally logged.

This example log output shows what the log messages look like:
```
W9mwH [ClientPool[cplcNNlZ].acquire]: Creating a new client [cliGq5Tk] (requiresGrpc: false)
W9mwH [ClientPool[cplcNNlZ].release]: Garbage collected client [cliGq5Tk] activeClientDeleted=true (0 active clients: {}, 0 failed clients: {})
##### [ClientPool[cplUf7qN].terminate]: Closing all active clients (0 active clients: {}, 0 failed clients: {})
```

The motivation for this PR, and the previous #2416 and #2420 PRs, is to help investigate b/437011084, where ClientPool.acquire() apparently is returning previously-garbage-collected clients, resulting in "GoogleError: The client has already been closed".

BEGIN_COMMIT_OVERRIDE
fix: pool.ts: add even more logging
END_COMMIT_OVERRIDE

### DIFF
--- a/dev/src/pool.ts
+++ b/dev/src/pool.ts
@@ -18,7 +18,7 @@ import {GoogleError} from 'google-gax';
 import * as assert from 'assert';
 
 import {logger} from './logger';
-import {Deferred, requestTag as generateClientId} from './util';
+import {Deferred, requestTag as generateTag} from './util';
 
 export const CLIENT_TERMINATED_ERROR_MSG =
   'The client has already been terminated';
@@ -79,6 +79,11 @@ export class ClientPool<T extends object> {
   private readonly terminateDeferred = new Deferred<void>();
 
   /**
+   * A unique identifier for this object, for inclusion in log messages.
+   */
+  private readonly instanceId = 'cpl' + generateTag();
+
+  /**
    * @param concurrentOperationLimit The number of operations that each client
    * can handle.
    * @param maxIdleClients The maximum number of idle clients to keep before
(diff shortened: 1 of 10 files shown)


dev/src/pool.ts
```diff
@@ -202,22 +207,23 @@ export class ClientPool<T extends object> {
...
+    const activeClientDeleted = this.activeClients.delete(client);
     this.failedClients.delete(client);
     await this.clientDestructor(client);
 
     logger(
-      'ClientPool.release',
+      `ClientPool[${this.instanceId}].release`,
       requestTag,
-      'Garbage collected client [%s] (%s)',
+      'Garbage collected client [%s] activeClientDeleted=%s (%s)',
```
MarkDuckworth, line 224: `activeClientDeleted=%s`

Double checking to ensure that %s will accept the boolean value and format it correctly?

MarkDuckworth (approved):
lgtm with question for your consideration

dev/src/pool.ts
dconeybe: Yep, it does. Here is an example log message:

```
W9mwH [ClientPool[cplcNNlZ].release]: Garbage collected client [cliGq5Tk] activeClientDeleted=true (0 active clients: {}, 0 failed clients: {})
```

The message contains `activeClientDeleted=true` as expected.

PR #2474, opened 2026-01-28
feat: add support for `regexFind` and `regexFindAll`

dlarocque:
Adds support for `regexFind` and `regexFindAll`. Most of this is copied from https://github.com/firebase/firebase-js-sdk/pull/9454.


MarkDuckworth (changes requested):
You will also need to update (types/firestore.d.ts)[https://github.com/googleapis/nodejs-firestore/blob/main/types/firestore.d.ts] with the types. It's an unfortunate side effect of the current structure of this repo, but shouldn't be too much work in this case.

MarkDuckworth (approved):

MarkDuckworth (approved):

PR #2484, opened 2026-02-12
chore: add autogenerated api-report files to `.gitignore`

dlarocque:
When `yarn api-report` is ran, these files are automatically generated. Since these are not tracked by git, they should be added to the `.gitignore`.


milaGGL (approved):
LGTM

PR #2485, opened 2026-02-12
feat: add remaining `map` expressions

dlarocque:
Adds support for the remaining `map` pipeline expressions; `mapSet`, `mapValues`, `mapEntries`, `mapKeys`.

Ported from: https://github.com/firebase/firebase-js-sdk/pull/9483


dev/src/pipelines/expression.ts
```diff
@@ -6342,6 +6431,93 @@ export function mapGet(
...
+ * ```
+ *
+ * @param map - The map to set entries in.
+ * @param key - The key to set. Must be a string or a constant string expression.
+ * @param value - The value to set.
+ * @param moreKeyValues - Additional key-value pairs to set.
+ * @returns A new `Expression` representing the map with the entries set.
+ */
+export function mapSet(
+  map: unknown,
```
milaGGL: new map expressions should use overloads to accept different types of map, like string, expression, record(if applicable), like other existing map expressions, ie, mapRemove, mapMerge.

milaGGL (approved):
LGTM

